PR #177, opened 2014-10-22
add unit/integrted tests for dpc

thegreatwall:


skxray/dpc.py
```diff
@@ -1,276 +1,462 @@
...
+
+    yline : ndarray
+        The column vector of the sums of each row.
+
+    """
+    
+    if bad_pixels:
+        im = im.copy()
+    
+    if bad_pixels is not None:
```
tacaswell: can merge these two if statements.

skxray/dpc.py
```diff
@@ -1,276 +1,462 @@
...
+    padding : int, optional
+        Pad a N-by-M array to be a (N*(2*padding+1))-by-(M*(2*padding+1)) array 
+        with the image in the middle with a (N*padding, M*padding) thick edge 
+        of zeros. Default is 0.
+        padding = 0 --> v (the original image, size = (N, M))
+                        0 0 0
+        padding = 1 --> 0 v 0 (the padded image, size = (3 * N, 3 * M))
+                        0 0 0
+
+    weighting : float, valid in [0, 1], optional
```
tacaswell: Don't include the range in the type spec, it will break the docstring parsing to wrap this for vistrails

thegreatwall: OK. Please leave a note when you finish the review, after which I will push a commit. Thanks.

tacaswell:
I am still very un-happy about the way 'padding' works as it does far more violence to the original data in fourier space (which I think is where it really matters how mangled the data has become), but that seems like something that Han-fei needs to be in on and I don't have the bandwidth to deal with that right now.

Other than those two very minor changes I think I am happy to merge this.

thegreatwall:
OK. I think maybe we can make another pull request for the 'padding' if necessary in the future.

PR #216, opened 2014-12-10
Basic img_proc module revision and re-org using API option

giltis:
Reorganized the thresholding, histogram, morphology and filter operations so that all direct imports are removed and now loaded via the api.

Remaining tools to follow.


skxray/img_proc/histops.py
```diff
@@ -0,0 +1,190 @@
...
+    bin_edges : array
+        1xN array containing the edge values for each bin
+        NOTE: the length of this array is 1 larger than the length of the
+        hist array (e.g. len(bin_edges) = len(hist) + 1)
+    """
+    hist, bin_edges = np.histogram(src_data,
+                                   bins=num_bins,
+                                   density=pd_function)
+    bin_avg = np.empty(len(hist))
+    intensity = iter(bin_edges)
```
giltis: Done

skxray/img_proc/threshops.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Returns
+    -------
+    output : ndarray
+        The function returns a binary array where all voxels with values equal 
+        to 1 correspond to voxels within the identified threshold range.
+    """
+
+    if thresh_value > np.amax(src_data):
+        raise ValueError("Selected threshold value is greater than the "
```
giltis: While you're right mechanically, I don't expect that users will want to use that as a "feature" and instead will want to know that they're actually out of range, and that something is "wrong" with the thresholding value that's being used. So, in my mind, the raising of an error in this case is the most appropriate action. For that reason, I added this check so that the applied threshold value has to fall within the max extents of the data set being processed.

skxray/img_proc/threshops.py
```diff
@@ -0,0 +1,324 @@
...
+                         "maximum array value. Current settings will "
+                         "result in an empty array. Current "
+                         "thresh_value = {0}, array maximum = {1}, "
+                         "resulting array fill value = {2}".format(
+                         thresh_value, np.amax(src_data), False))
+    elif thresh_value < np.amin(src_data):
+        raise ValueError("Selected threshold value is less than the "
```
giltis: Same comment as above.

skxray/img_proc/threshops.py
```diff
@@ -0,0 +1,324 @@
...
+    output : ndarray
+        The function returns a binary array where all voxels with values equal
+        to 1 correspond to voxels within the identified threshold range.
+
+    thresh_value : float
+        The threshold value determined by the thresholding function and used
+        to generate the binary volume data set
+    """
+    data_max = np.amax(src_data)
+    data_min = np.amin(src_data)
```
giltis: Cool. I didn't know that.

danielballan:
@giltis A lot of this is a really interesting to me. I'd like to read through it line by line, and while I'm at it I could put a PR on your branch fixing picky PEP8 stuff as I go. Is this OK / helpful?

skxray/io/tiff_io.py
```diff
@@ -0,0 +1,117 @@
...
+                           y,
+                           x), np.float32)
+    print('loading ' + file_name)
+    src_volume.data[:] = open(file_name).read() #sample file is now loaded
+    target_var = src_volume[:,:,:]
+    print 'Volume loaded successfully'
+    return target_var
+
+
+def read_tiff(file_name):
```
giltis: You're right. Originally it did not occur to me that I could accomplish this with a simple import statement.

giltis: Direct assignment, like you're suggested, assigns input ports to all of the parameters associated with both tifffile.imread and tifffile.imsave. Functionally, the additional parameters are optional, however, after wrapping, vistrails requires value assignment for any and all of the parameters. The defs as I've written them only require entry of the file path and file name. Once executed read_tiff functions correctly to load both 2D and "3D" tiff files.

skxray/img_proc/threshops.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Returns
+    -------
+    output : ndarray
+        The function returns a binary array where all voxels with values equal 
+        to 1 correspond to voxels within the identified threshold range.
+    """
+    # Check to make sure that maximum threshold boundary is less than or
+    # equal to the maximum array value, and
+    if thresh_value_max > np.amax(src_data):
```
giltis: Why not? Is there something I'm overlooking that would make this problematic?

tacaswell: Because it is a waste of time and will not change the output value.

skxray/img_proc/threshops.py
```diff
@@ -0,0 +1,281 @@
...
+"""
+This module contains tools for thresholding data sets.
+"""
+
+
+import numpy as np
+import skimage.filter as sk
+
+
+def thresh_globalGT(src_data, thresh_value, md_dict=None):
```
tacaswell: md_dict does not get used anywhere.

skxray/img_proc/transform.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+    This function allows for volume adjustment and rearrangement by providing 
+    the ability to swap volume axes.
+    
+    Parameters
+    ----------
+    src_data : ndarray
+        Identify source data for modification
+    
+    select_axes : string
```
giltis: Changed

skxray/img_proc/threshops.py
giltis: Ah, gotcha.

skxray/img_proc/threshops.py
tacaswell:
```python
thresh_val = XXX
for img in list_of_tenk_images:
    thresh_globalGT(img, thresh_val)
```

Having this blow up half way through because you have one frame where all of the value drop above or below your threshold does not seem like the right thing to do.  There are use-cases where you may want to check that you are not saturating, but that is easy enough to be checked on the user side.

I really don't see the advantage of this over `np.greater` which is already wrapped in now.

Another part of the curation process is making sure we don't include too many very similar functions which will only serve to over-whelm and confuse the users.

If  this behavior stays it needs to have very prominent warnings in the docstring.

skxray/img_proc/threshops.py
giltis: Changed.

skxray/img_proc/transform.py
```diff
@@ -0,0 +1,171 @@
...
+    data set along a given axix by a fixed number of degrees. After executing 
+    the rotation, arrays are expanded to include all of the original voxel data. 
+    As a result, a subsequent cropping step may be required to return a data set 
+    to its original shape and size.
+    
+    Parameters
+    ----------
+    src_data : ndarray
+        Identify source data for modification
+
```
giltis: Done.

skxray/img_proc/transform.py
```diff
@@ -0,0 +1,171 @@
...
+        select_axis = (1,0)
+    output = ndimage.rotate(src_data,
+                            rotate_degrees,
+                            select_axis,
+                            reshape=True,
+                            cval=fill_value)
+    return output
+
+
+def resize_volume():
```
giltis: Done.

skxray/img_proc/mathops.py
```diff
@@ -0,0 +1,465 @@
+# Module for the BNL image processing project
```
giltis: I removed everything except for logical_subtract and logical_nand, which are not included in numpy.

skxray/img_proc/mathops.py
```diff
@@ -0,0 +1,77 @@
...
+    Example
+    -------
+    result = mathops.logical_NAND('img_1', 'img_2')
+    """
+    output = np.logical_not(np.logical_and(src_data1, 
+                                           src_data2))
+    return output
+
+
+def logical_subtract(src_data1,
```
tacaswell: is this equivalent to `A & !B` ? if it is false in A, it is False in the output, if it is True an A and False in B it is True in the output and if it is True in A and True in B it is False in the output

A  | B  | !B |  A & !B 
--- | ---- | --- | ---
T | T | F |F
T | F | T |T
F | F | T |F
F | T |  F |F

skxray/img_proc/imgQuant.py
```diff
@@ -0,0 +1,135 @@
...
+    analysis and comparison of features or  materials can be useful
+    for quantitative comparison of results.
+
+
+    Parameters
+    ----------
+    src_labelField : ndarray
+         The segmented data set converted to a label field
+
+    num_bins : integer
```
giltis: So, do you think that num_bins should be renamed to something like num_phases, or num_materials? I need to rethink a standard definition for label_field, which will directly impact how the volume quantification functions will need to operate. Your suggestion, a long time ago, that materials within a label_field be assigned integer values according to 2^^Material_Number is the best way to address voxel assignment errors. However, using that format, num_bins (as it is currently included) would/could force the merging of bins, as you mention above.

I could simply define num_bins inside the function as:
num_bins = amax(label_field) - amin(label_field) + 1

That way there's no ambiguity, or room for merging of bins since every possible integer value assigned to a material gets its own bin.

skxray/img_proc/imgQuant.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+results_dict = {}
+def create_label_field(binary_vol_list,
+                       material_names_list,
+                       md_dict):
+    #md_dict should be optional input. If present then add to it.
+    return result_lblfield, mislabeled_voxel_vol, md_dict
+
+def Vol_trial_1(src_labelField,
```
giltis: Old hold over. Currently changed to q_vol_hist.

skxray/img_proc/imgQuant.py
```diff
@@ -0,0 +1,135 @@
...
+        passed to `density` kwarg of `np.histogram`
+
+    Returns
+    -------
+    output : array
+        Currently returns an array containing material number
+        (bin_edges) and quantified volume (voxel count).
+    """
+
+    hist, bin_edges = np.histogram(src_labelField,
```
giltis: Cool, Good call.

skxray/img_proc/imgQuant.py
tacaswell: Yes, or use `np.bincount` which does most of this logic for you.

tacaswell: Ah, I see I already suggested that....

skxray/io/avizo_io.py
```diff
@@ -263,7 +263,7 @@ def _create_md_dict(clean_header):
     return md_dict
 
 
-def load_amiramesh(file_path):
+def read_amiraMesh(file_path):
```
ericdill: I believe you're referencing this commit, which does not appear to be using `load_amiramesh`

https://github.com/strawlab/scikit-xray/commit/e5ce3d4de2637b839fb6abbf9cf55f989dd92782

skxray/img_proc/histops.py
```diff
@@ -0,0 +1,138 @@
...
+    min_final : float
+        Specify the new minimum value for the data set. Default 
+        value is 0
+
+    out_dType : np.dtype
+        Specify the desired data type for the output. The default 
+        resulting data type is 'uint8'. If desired resulting data type is
+        something other than 'uint8' then specify the desired data type here. 
+        Recognizable options include:
+            'int8'
```
ericdill: aren't **all** `np.dtype`s valid here?  Why list this many?

skxray/img_proc/histops.py
```diff
@@ -0,0 +1,138 @@
...
+out_dType = ['int8',
+             'int16',
+             'int32',
+             'int64',
+             'uint16',
+             'uint32',
+             'uint64',
+             'float16',
+             'float32',
+             'float64']
```
ericdill: This list is formatted strangely...  In fact, I'm noticing that many of your imports/function args/lists are formatted like this.  Given that it is different from the way I'm used to reading code and different from the rest of our library, I must ask... why are you formatting your code so differently?

skxray/img_proc/threshops.py
giltis: OHHH! That use case totally didn't occur to me. And you're absolutely right. I'll fix it.

danielballan: The rough usage is:

    images = thresh_globalGT(images, number)

A second function could do:

   list_of_nonblack_counts = check_saturation(images)

If everything is blacked out in some frame, one of the elements of the returned list will be 0. E.g.,
    >>> check_saturation(thresholded_images)
    [100, 50, 0, 25, 1]

The thresholding functions like `thresh_globalGT` should issue a warning stating that at least one element of the Z stack is fully black (or, hey, white), and it can direct the user to `check_saturation` to explore further.

skxray/img_proc/mathops.py
tacaswell: @giltis Ping!

PR #230, opened 2015-01-06
Doc update

tacaswell:
Sundry changes to sphinx source


danielballan:
This is still a bit of a mess, but we are about to do a doc overhaul and this is a step in a right direction, so I am merging.

PR #235, opened 2015-01-08
DOC : make sure outputs have names

tacaswell:
Need these to make VTTools happy (other wise the parsing of the output goes badly).


ericdill:
Looks good

PR #236, opened 2015-01-09
radial integration: powder data

sameera2004:


tacaswell:
This looks really good!

I would probably put the image array as the first position argument as that feels more natural to me, but that is not a huge deal.

Can you add a test?   I would just integrate on an image of all 1s so your integration is just area and by picking your input parameters carefully you can know what the values should be.

sameera2004:
Thanks, sure I will put a test function for the radial integration

skxray/core.py
```diff
@@ -729,6 +729,70 @@ def wedge_integration(src_data, center, theta_start,
     raise NotImplementedError()
 
 
+def radial_integration(pixel_size, calibrated_center, image_array,
+                       dist_sample, detector_size, wavelength,
+                       q_or_twotheta="Q"):
```
ericdill: I think it would be more sensible to rename this to something like `x_axis`, or `axis` or some such and then allow valid options of `q`, `two_theta`, `r`, and `d`. 

there should be an optional `bins` parameter that defaults to some sensible number

everything other than `calibrated_center` and `image_array` should be optional. That means that you'll have to ensure that you've got the correct input for each of the conversions

put image array first

remove `detector_size`

skxray/core.py
```diff
@@ -729,6 +730,118 @@ def wedge_integration(src_data, center, theta_start,
...
+    mask = counts > 10
+    # getting bin centers from bin edges
+    bin_centers = bin_edges_to_centers(bin_edges)[mask]
+    # radial integration
+    ring_averages = sums[mask] / counts[mask]
+
+    return bin_centers, ring_averages
+
+
+def _process_x_val(x_axis, val, dist_sample, wavelength):
```
tacaswell: This should probably be promoted to a full public function.  Converting pixel -> * is a useful thing.

skxray/core.py
```diff
@@ -729,6 +730,118 @@ def wedge_integration(src_data, center, theta_start,
...
+    else:
+        x_val = _process_x_val(x_axis, pixel_val, dist_sample, wavelength)
+
+    if num_bins is None:
+        num_bins = np.sum(image_array.shape)
+
+    bin_edges, sums, counts = bin_1D(np.ravel(x_val), np.ravel(image_array),
+                                     nx=num_bins)
+
+    mask = counts > 10
```
tacaswell: We should not be discarding data based on a hard-code threshold.

skxray/core.py
sameera2004: Sure I will do that

skxray/core.py
```diff
@@ -729,6 +730,118 @@ def wedge_integration(src_data, center, theta_start,
     raise NotImplementedError()
 
 
+def radial_integration(image_array, calibrated_center,
```
tacaswell: I am torn on this API.  On one hand, this is convenient, on the other hand I would like to have this function return bin-edges in pixels and be converted later by what ever code calls this function.

skxray/core.py
```diff
@@ -729,6 +730,118 @@ def wedge_integration(src_data, center, theta_start,
...
+    """
+    # convert to pixels to radius
+    pixel_val = pixel_to_radius(np.shape(image_array), calibrated_center)
+
+    if x_axis == 'r':
+        x_val = pixel_val
+    else:
+        x_val = _process_x_val(x_axis, pixel_val, dist_sample, wavelength)
+
+    if num_bins is None:
```
tacaswell: Is this a reasonable heuristic?

skxray/core.py
```diff
@@ -729,6 +730,118 @@ def wedge_integration(src_data, center, theta_start,
...
+        wavelength of the incoming x-rays (Angstroms)
+
+    pixel_size : tuple, optional
+        The size of a pixel in real units.
+        (height, width). (mm)
+
+    dist_sample : float, optional
+        distance from the sample to the detector (mm)
+
+    num_bins : int, optional
```
tacaswell: add a note about how the default value in chosen.

skxray/core.py
```diff
@@ -729,6 +730,118 @@ def wedge_integration(src_data, center, theta_start,
...
+    ring_average : ndarray
+        radial integration of intensity
+
+    """
+    # convert to pixels to radius
+    pixel_val = pixel_to_radius(np.shape(image_array), calibrated_center)
+
+    if x_axis == 'r':
+        x_val = pixel_val
+    else:
```
tacaswell: I would defer this decision until after doing the binning.

```python
bin_edges, sums, counts = bin_1D(np.ravel(pixel_val), np.ravel(image_array), num_bins)
if x_axis != 'r':
    bin_edges = _process_x_val(x_axis, bin_edges, dist_sample, wavelength)
```

skxray/core.py
```diff
@@ -729,6 +730,118 @@ def wedge_integration(src_data, center, theta_start,
...
+
+    if x_axis == "two_theta":
+        x_val = two_theta_val
+    elif x_axis == "q":
+        if wavelength is None:
+                raise ValueError("Provide the wavelength to find"
+                                 " the q space values")
+        # convert to q space values from known two theta values
+        x_val = twotheta_to_q(two_theta_val, wavelength)
+    else:
```
tacaswell: if `x_axis == 'r'` this exception will be raised.

skxray/core.py
```diff
@@ -729,6 +730,118 @@ def wedge_integration(src_data, center, theta_start,
...
+        wavelength of the incoming x-rays (Angstroms)
+
+    Returns
+    -------
+    x_val : ndarray
+        x axis values (two theta or Q space)
+        shape (image array shape)
+
+    """
+    if dist_sample is 'None':
```
tacaswell: I don't think we need this check anymore

skxray/core.py
```diff
@@ -729,6 +730,118 @@ def wedge_integration(src_data, center, theta_start,
...
+                raise ValueError("Provide sample to detector distance"
+                                 " to find the two theta space values")
+
+    # convert to radius to two theta
+    two_theta_val = radius_to_twotheta(dist_sample, val)
+
+    if x_axis == "two_theta":
+        x_val = two_theta_val
+    elif x_axis == "q":
+        if wavelength is None:
```
tacaswell: `wavelength` is currently required, if it should be optional make it a kwarg with default value `None`.

skxray/tests/test_core.py
```diff
@@ -349,16 +347,19 @@ def test_radius_to_twotheta():
...
-                          0.5036263, 0.51142, 0.51914611, 0.52680461, 0.53439548,
-                          0.54191875, 0.54937448, 0.55676277, 0.56408372, 0.57133748,
-                          0.57852421, 0.58564412, 0.5926974, 0.59968432, 0.60660511,
-                          0.61346007, 0.62024949, 0.62697369, 0.63363301, 0.6402278,
-                          0.64675843, 0.65322528, 0.65962874, 0.66596924, 0.67224718,
-                          0.67846301, 0.68461716, 0.6907101, 0.69674228, 0.70271418,
-                          0.70862627, 0.71447905, 0.720273, 0.72600863, 0.73168643,
-                          0.73730693, 0.74287063, 0.74837805, 0.75382971, 0.75922613,
-                          0.76456784, 0.76985537, 0.77508925, 0.78027, 0.78539816])
+    two_theta = np.array([0.46364761, 0.47177751, 0.47984053, 0.48783644,
```
tacaswell: Change this to use generative tests.  For a given input we _know_ what the answer must be.

skxray/tests/test_core.py
```diff
@@ -538,6 +543,34 @@ def run_image_to_relative_xyi_repeatedly():
             print('{0} calls successful'.format(num_calls))
 
 
+def test_radial_integration():
```
tacaswell: This test is mostly a smoke test.  Can you add a test where the expected value is not all ones?

skxray/tests/test_core.py
```diff
@@ -538,6 +543,34 @@ def run_image_to_relative_xyi_repeatedly():
...
+                                                      dist_sample=dist_s,
+                                                      num_bins=40)
+    # when the x axis is in Q space
+    (bin_cen_q, ring_ave_q) = core.radial_integration(image, calib_center,
+                                                      x_axis='q',
+                                                      wavelength=wavelength,
+                                                      dist_sample=dist_s,
+                                                      num_bins=20)
+
+    assert_array_equal(ring_ave_r, np.ones(49))
```
tacaswell: This should return 50, I think there is on off by one bug in the code someplace...

skxray/core.py
sameera2004: dist_sample is optional for the radial_integration module. However, to get two theta values or q values you need dist_sample. Therefore, there should be raise valueError here to check it

skxray/core.py
sameera2004: Sure I will make threshold optional (take out the hard code threshold)

skxray/core.py
sameera2004: Sure I think it is good to take this decision after binning

sameera2004:
Closing, similar function in roi.py circular_average

PR #238, opened 2015-01-09
Collect calibration functions

danielballan:
Apologies that this PR is a grab bag of changes. I can cherry-pick as needed.
- Move "dictionary-like classes," some of which aren't used at all, away from the top of the core module into a new `utils.py`.
- Nuke `image.py`, which only has one member, here moved into `calibration.py` for now. Maybe reinstate it `image.py` later, but avoid excessive module-splitting for now.
- Remove `wedge_integration`, which was `NotImplemented`.
- Make minor edits to clean up docstrings.
- Adding a plotting module with two new functions (many more to come).


ericdill:
Going to close this. No action in a while. Please submit new commits and reopen if this is still of interest

ericdill:
@danielballan  -- Sorry about the premature close on this one. Do you think that plotting.py should be salvaged from this PR?

ericdill:
Oh wow I just realized this was from before the rename. Brings me back 😂

danielballan:
Wow, this PR would have been from my first week of work at BNL. Yeah, it's think it's worth salvaging the plotting code; the rest has probably be superceded by multiple re-organizations. Will come back to this soon.

PR #239, opened 2015-01-09
Find ring center API

danielballan:
Merge this after merging #238. This implements

    find_ring_center(image, method)

where `method` is an optional string like `'autocorrelation'`. In the future, other methods may be added -- particularly ones that can handle images showing only one quadrant of the rings.


danielballan:
This new API is demonstrated [in this notebook](http://nbviewer.ipython.org/gist/danielballan/4ae9a2c325ceb5e182db). The notebook is fully working, but I would like to add more content demonstrating `estimate_d...` and/or interacting with the hexrd library. Once that is finished a future version of the notebook will go into scikit-xray-examples.

danielballan:
With that, I believe this and #238 are ready to merge.

danielballan:
Said notebook is sitting in this work-in-progress (WIP) PR: Nikea/scikit-xray-examples#13

skxray/calibration.py
```diff
@@ -225,3 +225,101 @@ def refine_center(image, calibrated_center, pixel_size, phi_steps, max_peaks,
...
+
+    The **autocorrelation** method uses correlation between the image and
+    it's mirror to find the approximate center of  a single set of concentric
+    rings. It is assumed that there is only one set of rings in the image.
+    For this method to work well the image must have significant mirror-
+    symmetry in both dimensions.
+    """
+    if method is None:
+        method = 'autocorrelation'
+    if method == 'autocorrelation':
```
tacaswell: Should probably do this with a look up dictionary

skxray/calibration.py
```diff
@@ -225,3 +225,101 @@ def refine_center(image, calibrated_center, pixel_size, phi_steps, max_peaks,
...
+        return _find_ring_center_acorr_1D(input_image)
+    else:
+        raise ValueError("Supported methods are: {0}".format(
+            ', '.join(FIND_RING_CENTER_METHODS)))
+
+
+def _find_ring_center_acorr_1D(input_image):
```
tacaswell: I don't think we want to make this private. I agree that the main entrance should be through the convenience method, but we don't want to hide the guts.

skxray/calibration.py
danielballan: Good idea.

skxray/calibration.py
danielballan: Agreed. For documentation purposes, these guts should not be the first thing people see, but they don't have to be private in the code.

PR #241, opened 2015-01-09
API: Clean up img_to_relative_xyi and put input validation in new file.

### LINKED ISSUE #240: Revisit `img_relative_to_xyi` API
Minor API / docstring issues:

  * `pixel_size_x` and `pixel_size_y` are separate kwargs here, but some other functions somewhere wants `pixel_size=(x, y)`. Pick one and be consistent.

  * Remove "**kwargs" from docstring: it is not in the signature.

danielballan:
This closes #240 by fixing minor issues in the API to `img_to_relative_xyi`. More importantly, it introduces a new file, `validations.py` with a function that is here used to check that `pixel_size` make sense. That function, available at `validations.pixel_size`, should be used in many other places as well. And we can add more functions for other common arguments.

There may be a fancier way to do this, but let's keep it simple for now. Reusing input validations will save a lot of code and mental effort.

### DIFF
--- a/skxray/validations.py
+++ b/skxray/validations.py
@@ -1,0 +1,53 @@
+#! encoding: utf-8
+# ######################################################################
+# Copyright (c) 2014, Brookhaven Science Associates, Brookhaven        #
+# National Laboratory. All rights reserved.                            #
+#                                                                      #
+# Redistribution and use in source and binary forms, with or without   #
+# modification, are permitted provided that the following conditions   #
+# are met:                                                             #
+#                                                                      #
+# * Redistributions of source code must retain the above copyright     #
+#   notice, this list of conditions and the following disclaimer.      #
+#                                                                      #
+# * Redistributions in binary form must reproduce the above copyright  #
+#   notice this list of conditions and the following disclaimer in     #
+#   the documentation and/or other materials provided with the         #
+#   distribution.                                                      #
+#                                                                      #
+# * Neither the name of the Brookhaven Science Associates, Brookhaven  #
+#   National Laboratory nor the names of its contributors may be used  #
+#   to endorse or promote products derived from this software without  #
+#   specific prior written permission.                                 #
+#                                                                      #
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS  #
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT    #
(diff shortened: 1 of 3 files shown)


tacaswell:
It may be worth considering if we want to do normalization + default values along with validation.  That might be a clean way to centralize user-configurable defaults.  The pixel size function would look something like

```python
def pixel_size(val):
    if val is None:
        return (1, 1)
    val = tuple(float(v) for v in val)
    if len(val) != 2:
        raise ValueError(...)
    if any(v<=0 for v in val):
        raise ValueError(...)
    return val
```

and would be used like

```python
pixel_size = validate.pixel_size(pixel_size)
```

skxray/validations.py
```diff
@@ -0,0 +1,53 @@
...
+# INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES   #
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR   #
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)   #
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,  #
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OTHERWISE) ARISING   #
+# IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE   #
+# POSSIBILITY OF SUCH DAMAGE.                                          #
+########################################################################
+"""
+This module is for the 'core' data types.
```
ericdill: should probably change this

danielballan:
Yes, I like the pattern

    pixel_size = validations.pixel_size(pixel_size)

and if we are agreed I will implement that throughout the code when I go on my API-consistency crusade next week.

danielballan:
Any thoughts on this? I am inclined against, but thought I'd mention it:

http://www.pythoncentral.io/validate-python-function-parameters-and-return-types-with-decorators/

tacaswell:
That is sort of what I had in mind, only reading from the `__annotations__` (which I need to find/write a detorator to make easy to add)

danielballan:
Don't forget to include a demo! :- D

danielballan:
@tacaswell I just up on annotations. I have a good plan for this I think. I'll take a first pass at it.

danielballan:
*I just READ up.

PR #254, opened 2015-03-15
One time Correlation

sameera2004:


skxray/tests/test_correlation.py
```diff
@@ -0,0 +1,81 @@
...
+    roi_data = np.array(([20, 50, 10, 8 ], [60, 70, 12, 6], [140, 120, 5, 10]),
+                                       dtype=np.int64)
+
+    (q_inds, num_pixels,
+     pixel_list) = diff_roi.roi_rectangles(num_qs, roi_data, img_dim)
+
+    img_stack = []
+    for i in range(500):
+        img_stack.append(np.random.randint(1, 5, size=(img_dim)))
+
```
licode: add: img_stack = np.array(img_stack)

sameera2004: Sure, Thanks I will add

sameera2004:
c4b_t_250k 9950 images

These are the one time correlation plots for C4B_T 250K (9950) images for 8 number of rings. 12 multi tau levels and 8 number of buffers.

skxray/tests/test_correlation.py
```diff
@@ -0,0 +1,95 @@
...
+    g2, lag_steps, elapsed_time = corr.auto_corr(num_levels, num_bufs,
+                                                 num_qs, num_pixels,
+                                                 pixel_list, q_inds,
+                                                 np.asarray(img_stack))
+
+    assert_array_almost_equal(lag_steps, np.array([0, 1, 2, 3, 4, 6, 8, 12, 16, 24]))
+
+    g2_m = np. array([[1.200, 1.200],
```
ericdill: I am amazed that this even works!  Just for sanity sake, remove the space between `np.` and ` array`.  Turns out white space around the operator is ignored??  Because this works ...what? @tacaswell @danielballan Why does this work?

```python
In [7]: np .  random . random()
Out[7]: 0.2532634036376479
```

danielballan: Ha, that's a new one to me. But np.array <-> np. array is no different than
1+2 <-> 1+ 2.

danielballan:
@sameera2004 Here's a useful bit of numpy. All what is done by the `avg_img` function in your notebook can be done like so:

    np.mean(img_array, 0)

Also, notice that the second argument in `avg_img` (`num_images`) could have been computed without user input by calling `len(img_array)`.

skxray/tests/test_correlation.py
sameera2004: I didn't get any error!! yes it is new to me too
anyway I will fix it

sameera2004:
Thanks @danielballan I will use np.mean instead of avg_img

skxray/correlation.py
```diff
@@ -0,0 +1,299 @@
...
+    lag_steps : ndarray
+        delay or lag steps for the multiple tau analysis
+
+    elapsed_times : float
+        elapsed time for auto correlation (in seconds)
+
+    Notes
+    -----
+    In order to calculate correlations for delays, images must be
+    keep for upto the maximum delay. These are stored in the array
```
tacaswell: kept

skxray/correlation.py
```diff
@@ -0,0 +1,299 @@
...
+        "Area detector based photon correlation in the regime of
+        short data batches: Data reduction for dynamic x-ray
+        scattering," Rev. Sci. Instrum., vol 70, p 3274-3289, 2000.
+
+    """
+    # number of pixels in required roi's, dimensions are : [num_qs]X1
+    num_pixels = np.bincount(q_inds, minlength=(num_qs+1))
+    num_pixels = num_pixels[1:]
+
+    for item in num_pixels:
```
tacaswell: `if np.any(num_pixels == 0):`

skxray/correlation.py
```diff
@@ -0,0 +1,299 @@
...
+    num = np.array(np.zeros(num_levels), dtype=np.int64)
+
+    # number of frames(images)
+    num_frames = img_stack.shape[0]
+
+    start_time = time.time()
+    for n in range(0, num_frames):  # changed the number of frames
+        image_array = img_stack[n]
+
+        cur[0] = 1 + cur[0] % num_bufs  # increment buffer
```
tacaswell: Can you put `()` in this just to be sure the order of operations is correct?

skxray/correlation.py
```diff
@@ -0,0 +1,299 @@
...
+
+    # matrix of one-time correlation for required roi's
+    g2 = np.zeros((num_levels, num_qs), dtype=np.float64)
+
+    # correlation for delays, images must be keep for up to maximum
+    # delay in buf
+    buf = np.zeros((num_levels, num_bufs, np.sum(num_pixels)),
+                   dtype=np.float64)
+
+    cts = np.zeros(num_levels)
```
tacaswell: These cryptic variables need some explanatory comments

skxray/correlation.py
```diff
@@ -0,0 +1,299 @@
...
+                                    buf_no=cur[0] - 1)
+        if num_levels > 1:
+            processing = 1
+            level = 1
+        else:
+            processing = 0
+
+        while processing:
+            if cts[level]:
+                prev = 1 + (cur[level - 1] - 2 + num_bufs) % num_bufs
```
tacaswell: This code block needs comments explaining what it is doing.

skxray/correlation.py
```diff
@@ -0,0 +1,299 @@
...
+
+    tot_channels, lag_steps = core.multi_tau_lags(num_levels,
+                                                  num_bufs)
+
+    lag_steps = lag_steps[:g2.shape[0]]
+
+    return g2, lag_steps, elapsed_time
+
+
+def _process(buf, G, IAP, IAF, q_inds, num_bufs,
```
tacaswell: The arrays that are modified in place need to be marked.

skxray/tests/test_correlation.py
```diff
@@ -0,0 +1,95 @@
...
+                                      num_qs, first_r, delta_r)
+    roi_data = np.array(([60, 70, 12, 6], [140, 120, 5, 10]),
+                        dtype=np.int64)
+
+    (q_inds, num_pixels,
+     pixel_list) = diff_roi.roi_rectangles(num_qs, roi_data, img_dim)
+
+    img_stack = []
+    for i in range(500):
+        img_stack.append(np.random.randint(1, 5, size=(img_dim)))
```
tacaswell: You can do this as `img_stack = np.random.randint(1. 5, size=(500, ) + img_dim)`

skxray/tests/test_correlation.py
```diff
@@ -0,0 +1,95 @@
...
+
+    g2, lag_steps, elapsed_time = corr.auto_corr(num_levels, num_bufs,
+                                                 num_qs, pixel_list, q_inds,
+                                                 np.asarray(img_stack))
+
+    assert_array_almost_equal(lag_steps, np.array([0, 1, 2, 3, 4, 6, 8,
+                                                   12, 16, 24]))
+
+    g2_m = np.array([[1.200, 1.200],
```
tacaswell: Where do these numbers come from?

skxray/correlation.py
tacaswell: What does this function get you over embedding this code directly in the loop above?

tacaswell: never mind, I now see it is used twice

skxray/correlation.py
```diff
@@ -0,0 +1,299 @@
...
+
+    pixel_list : ndarray
+        pixel indices for the required roi's
+
+    q_inds : ndarray
+        indices of the required roi's
+
+    img_stack : ndarray
+        intensity array of the images
+        dimensions are: [num_img][num_rows][num_cols]
```
tacaswell: ''shape is (num_imag, num_rows, num_cols)'`

skxray/correlation.py
```diff
@@ -0,0 +1,299 @@
...
+        g_max = IAP.shape[0]
+
+    g2 = (G[: g_max] / (IAP[: g_max] * IAF[: g_max]))
+
+    tot_channels, lag_steps = core.multi_tau_lags(num_levels,
+                                                  num_bufs)
+
+    lag_steps = lag_steps[:g2.shape[0]]
+
+    return g2, lag_steps, elapsed_time
```
tacaswell: The elapsed time should not be a returned value, but go into a logging call.

skxray/correlation.py
tacaswell: And it looks like this should be a `bool` aray.

skxray/correlation.py
```diff
@@ -0,0 +1,299 @@
...
+        G, IAP, IAF, num = _process(buf, G, IAP, IAF, q_inds,
+                                    num_bufs, num_pixels, num, level=0,
+                                    buf_no=cur[0] - 1)
+        if num_levels > 1:
+            processing = 1
+            level = 1
+        else:
+            processing = 0
+
+        while processing:
```
tacaswell: I can't figure out why this look can not be done as 

```python
for level in range(num_levels):
    ...
```
which would get ride of the `cts` variable.

skxray/correlation.py
```diff
@@ -0,0 +1,299 @@
...
+    buf = np.zeros((num_levels, num_bufs, np.sum(num_pixels)),
+                   dtype=np.float64)
+
+    cts = np.zeros(num_levels)
+    cur = np.ones((num_levels), dtype=np.int64)
+
+    num = np.array(np.zeros(num_levels), dtype=np.int64)
+
+    # number of frames(images)
+    num_frames = img_stack.shape[0]
```
tacaswell: do this as `len(img_stack)` so that we don't have to have all of the images in core to run this code.

sameera2004:
Fixed the code according to comments, ready for review

sameera2004:
@tacaswell I fixed the code it is ready for review.

tacaswell:
This needs better tests if they passed when `cts` was removed.

skxray/correlation.py
```diff
@@ -0,0 +1,319 @@
...
+import numpy as np
+import logging
+logger = logging.getLogger(__name__)
+import time
+
+import skxray.core as core
+
+
+def auto_corr(num_levels, num_bufs, num_qs,
+              pixel_list, q_inds, img_stack):
```
tacaswell: I think it would be simpler if this function took as input a labeled image and then computed `num_qs`, `pixel_list`, and `q_inds` internally.  This will require some changes to the ROI functions, but I think a workflow that looks like

```python
roi = get_roi_func(...)
roi_masked = mask_array(roi, dead_pixel_mask)
corr = auto_corr(image_source, num_levels, num_buffers, roi_masked)
```

skxray/correlation.py
```diff
@@ -0,0 +1,319 @@
...
+                   dtype=np.float64)
+
+    # to track processing each level
+    cts = np.zeros(num_levels)
+
+    # to increment buffer
+    cur = np.ones((num_levels), dtype=np.int64)
+
+    # to track how many images processed in each level
+    num = np.array(np.zeros(num_levels), dtype=np.int64)
```
tacaswell: You can pass `dtype` to `np.zeros`

skxray/correlation.py
```diff
@@ -0,0 +1,319 @@
...
+
+    # starting time for the process
+    t1 = time.time()
+
+    for n in range(0, len(img_stack)):  # changed the number of frames
```
tacaswell: Do this as 

```python
for n, img in enumerate(img_stack):
```
Because you don't need to know how many images there are.  This changes means we can pass generators in to this function.

skxray/correlation.py
```diff
@@ -0,0 +1,319 @@
...
+
+        cur[0] = (1 + cur[0]) % num_bufs  # increment buffer
+
+        # add image data to the buf to use for correlation
+        buf[0, cur[0] - 1] = (np.ravel(img_stack[n]))[pixel_list]
```
tacaswell: along with the change above this becomes

```
buf[0, cur[0] - 1] = np.ravel(img)[pixel_list]
```

I can never remember which of `ravel` and `flatten` makes a copy and which can return a view.

skxray/correlation.py
```diff
@@ -0,0 +1,319 @@
...
+
+        # call the _process function for multi-tau level one
+        G, IAP, IAF, num = _process(buf, G, IAP, IAF, q_inds,
+                                    num_bufs, num_pixels, num, level=0,
+                                    buf_no=cur[0] - 1)
+
+        # check whether the number of levels is one, otherwise
+        # continue processing the next level
+        if num_levels > 1:
+            processing = 1
```
tacaswell: Can you use `True` and `False` for `processing` to make clear that this is a boolean flag?

skxray/correlation.py
```diff
@@ -0,0 +1,319 @@
...
+    # matrix of one-time correlation for required roi's
+    g2 = np.zeros((num_levels, num_qs), dtype=np.float64)
+
+    # correlation for delays, images must be keep for up to maximum
+    # delay in buf
+    buf = np.zeros((num_levels, num_bufs, np.sum(num_pixels)),
+                   dtype=np.float64)
+
+    # to track processing each level
+    cts = np.zeros(num_levels)
```
tacaswell: what does `cts` stand for?  can you use a boolean array here?

skxray/correlation.py
```diff
@@ -0,0 +1,319 @@
...
+            processing = 1
+        else:
+            processing = 0
+
+        # the image data will be saved in buf according to each level then call
+        #  _process function to calculate one time correlation functions
+        level = 1
+        while processing:
+            if cts[level]:
+                prev = 1 + (cur[level - 1] - 1 - 1 + num_bufs)%num_bufs
```
tacaswell: the two `-1` here looks odd, should that be `-2` or is that a typo?

tacaswell: and I don't understand why there is the `num_bufs` in the `()` .  If you are taking the mod of it, that is a null operation.

skxray/correlation.py
```diff
@@ -0,0 +1,319 @@
...
+        # check whether the number of levels is one, otherwise
+        # continue processing the next level
+        if num_levels > 1:
+            processing = 1
+        else:
+            processing = 0
+
+        # the image data will be saved in buf according to each level then call
+        #  _process function to calculate one time correlation functions
+        level = 1
```
tacaswell: Would it be simpler to start this at `0` which would reduce the number of `-1` required below?

skxray/correlation.py
```diff
@@ -0,0 +1,319 @@
...
+                cts[level] = 1
+                processing = 0
+
+    # ending time for the process
+    t2 = time.time()
+
+    logger.info("Processing time for {0} images took {1} seconds."
+                "".format(len(img_stack), (t2-t1)))
+
+    if len(np.where(IAP == 0)[0]) != 0:
```
tacaswell: Can you explain what this code is doing?

skxray/correlation.py
```diff
@@ -0,0 +1,319 @@
...
+    :math ::
+        IAF = <I(t + delay)>
+
+    """
+    num[level] += 1
+
+    if level == 0:
+        i_min = 0
+    else:
+        i_min = int(num_bufs/2)
```
tacaswell: You can use `num_bufs // 2` to get this result more clearly.

Do you really only do half of the correlations?

skxray/correlation.py
sameera2004: np.ravel Return a flattened array. np.flattern Return a copy of the array collapsed into one dimension.

I want to flattened the array

skxray/correlation.py
sameera2004: Sure

skxray/correlation.py
sameera2004: Yes after level 0 all other levels, we have to do the half of the correlations

skxray/correlation.py
sameera2004: To work on pixel list , ... and roi information inside the auto_corr function will not be a good idea. Because user may have different options of creating roi information 

eg: roi rectangels, roi rings(different ring options), etc

Therefore better to find the roi information(pixel list, indices list work on mask) before calling this function

skxray/correlation.py
sameera2004: cts to rack processing each level. When the processed is finished cts will be zero,

skxray/correlation.py
sameera2004: Sure I will use True and False

skxray/correlation.py
sameera2004: Yes then we have to include +1 where ever level required

skxray/correlation.py
tacaswell: The point of my suggestion is to make this more flexible.  The input would be an integer array labeling the roi number of that pixel with 0 meaning "don't use".  These label images are useful out this context as well as scikit-image/scipy.ndimage have a number of functions which work on an image and a label array.

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+
+"""
+
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+import six
+import numpy as np
+import numpy.ma as ma
+import logging
+logger = logging.getLogger(__name__)
```
tacaswell: picky thing, move this below all of the imports.

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+import time
+
+import skxray.core as core
+
+
+def auto_corr(num_levels, num_bufs, indices, img_stack, mask=None):
```
tacaswell: This function should have a name that indicates it is using the multi-tau scheme.

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+    Parameters
+    ----------
+    num_levels : int
+        number of levels of multiple-taus
+
+    num_bufs : int, even
+        number of channels or number of buffers in
+        multiple-taus (must be even)
+
+    indices : ndarray
```
tacaswell: This should be re-named to 'labeled_roi' or something similar.

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+        indices of the required region of interest's (roi's)
+        dimensions are: (num_rows, num_cols)
+
+    img_stack : ndarray
```
tacaswell: This should only need to be an iterable

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+        intensity array of the images
+        dimensions are: (num of images, num_rows, num_cols)
+
+    mask : ndarray, optional
```
tacaswell: The mask should be pre-applied to the ROI labels so we don't need it as an input here.

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+    This module is for one time correlation.
+    The multi-tau correlation scheme was used for
+    finding the lag times (delay times).
+
+    Parameters
+    ----------
+    num_levels : int
+        number of levels of multiple-taus
+
+    num_bufs : int, even
```
tacaswell: can we call this 'channels' ?

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+
+def auto_corr(num_levels, num_bufs, indices, img_stack, mask=None):
+    """
+    This module is for one time correlation.
+    The multi-tau correlation scheme was used for
+    finding the lag times (delay times).
+
+    Parameters
+    ----------
+    num_levels : int
```
tacaswell: can we call this just 'levels'?

sameera2004: I think num_levels and num_bufs are better than levels and channels. Other thing is I am using level for the current level.

skxray/correlation.py
sameera2004: Sure I will change indices to 'labeled_roi'

skxray/correlation.py
sameera2004: Ok I will remove the mask in from one time correlation

skxray/correlation.py
sameera2004: Sure I will move 'logger' below all of the imports

skxray/correlation.py
tacaswell: Can you give this a more descriptive name?

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+                         " shape of the indices array")
+
+    if mask is not None:
+        roi_inds = ma(indices)
+    else:
+        roi_inds = indices
+
+    #  to get indices, number of roi's, number of pixels in each roi's and
+    #  pixels indices for the required roi's.
+    q_inds, num_qs, num_pixels, pixel_list = _get_roi_info(roi_inds)
```
tacaswell: Can you rename this to remove 'q'?  Something like `roi_inds, num_rois, num_pixels, pixel_list = _get_roi_info(...)`

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+    """
+    if num_bufs%2 != 0:
+        raise ValueError("number of channels(number of buffers) in"
+                         " multiple-taus (must be even)")
+
+    if indices.shape == img_stack[0].shape[1:]:
+        raise ValueError("Shape of an image should be equal to"
+                         " shape of the indices array")
+
+    if mask is not None:
```
tacaswell: as per above this block of code should go away.

sameera2004: Sure

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+        raise ValueError("Number of pixels of the required roi's"
+                         " cannot be zero, "
+                         "num_pixels = {0}".format(num_pixels))
+
+    # matrix of auto-correlation function without normalizations
+    G = np.zeros(((num_levels + 1)*num_bufs/2, num_qs),
+                 dtype=np.float64)
+
+    # matrix of past intensity normalizations
+    IAP = np.zeros(((num_levels + 1)*num_bufs/2, num_qs),
```
tacaswell: Can you rename this to `past_intensity_norm` and `IAF` -> `future_intensity_norm` or something like that?

skxray/correlation.py
sameera2004: Sure I will remove 'q' and use 'roi'

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+                   dtype=np.float64)
+
+    # matrix of future intensity normalizations
+    IAF = np.zeros(((num_levels + 1)*num_bufs/2, num_qs),
+                   dtype=np.float64)
+
+    # matrix of one-time correlation for required roi's
+    g2 = np.zeros((num_levels, num_qs), dtype=np.float64)
```
tacaswell: How is this different than `G`?

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+                   dtype=np.float64)
+
+    # to track processing each level
+    cts = np.zeros(num_levels)
+
+    # to increment buffer
+    cur = np.ones(num_levels, dtype=np.int64)
+
+    # to track how many images processed in each level
+    num = np.zeros(num_levels, dtype=np.int64)
```
tacaswell: num -> `img_per_level`

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+
+    # starting time for the process
+    t1 = time.time()
```
tacaswell: t1 -> `start_time`

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+                if level<num_levels:
+                    processing = True
+                else:
+                    processing = False
+            else:
+                cts[level] = 1
+                processing = False
+
+    # ending time for the process
+    t2 = time.time()
```
tacaswell: t2 -> `stop_time`

tacaswell: or `finish_time`

skxray/correlation.py
tacaswell: it looks like you never use this except for on line 227 where it is re-assigned.

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+
+    logger.info("Processing time for {0} images took {1} seconds."
+                "".format(len(img_stack), (t2-t1)))
+
+    # to get the final G, IAP and IAF values
+    g_max = IAP.shape[0]
+
+    # calculate the one time correlation
+    g2 = (G[: g_max] / (IAP[: g_max] * IAF[: g_max]))
```
tacaswell: I don't understand the slicing logic here, they should all be the same shape right?

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+    for i in range(i_min, min(num[level], num_bufs)):
+        t_index = level*num_bufs/2 + i
+
+        delay_no = (buf_no - i) % num_bufs
+
+        IP = buf[level, delay_no]
+        IF = buf[level, buf_no]
+
+        G[t_index] += ((np.bincount(q_inds,
+                                    weights=np.ravel(IP*IF))[1:])/num_pixels
```
tacaswell: why do you drop the first element?

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+        i_min = 0
+    else:
+        i_min = num_bufs//2
+
+    for i in range(i_min, min(num[level], num_bufs)):
+        t_index = level*num_bufs/2 + i
+
+        delay_no = (buf_no - i) % num_bufs
+
+        IP = buf[level, delay_no]
```
tacaswell: Can you give these more descriptive names?  I assume this is `past_image` and `future_image`?

skxray/correlation.py
sameera2004: G2 is without normalization, we calculate that for each image comes, and for each level.
 g2 is the required final one time correlation function.
G   = <I(t)I(t + delay)>,IAP = <I(t)>, IAF = <I(t + delay)>
and g2 = G/(IAP*IAF)

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+        IF = buf[level, buf_no]
```
tacaswell: This can be out side of this loop I think (as both level and buf_no are inputs).

skxray/correlation.py
```diff
@@ -0,0 +1,375 @@
...
+
+    # finding the lag times (delay times) for multi-tau levels
+    tot_channels, lag_steps = core.multi_tau_lags(num_levels,
+                                                  num_bufs)
+    lag_steps = lag_steps[:g2.shape[0]]
+
+    return g2, lag_steps
+
+
+def _process(buf, G, IAP, IAF, q_inds, num_bufs,
```
tacaswell: above comments about naming should propagate through here.

tacaswell:
This is looking good as the comments seem to be down to name changes to make the code clearer.

skxray/correlation.py
sameera2004: First element is the zero bin

tacaswell: Ah, I was misreading the `()` and thought you were slicing on `(IP*IF)[1:]`  I would break this up into two steps

``python
tmp_binned = (np.bin_count(...)[1:]) / num_pixels
G[t_index] += tmp_binned
```

File: skxray/correlation.py (hunk 41 above)
Comment (tacaswell) [inline skxray/correlation.py]: and I would do the raveling here instead of everytime you use them below (and I am pretty sure that they are _already_ 1D by the time they get put into the buf array).

File: skxray/correlation.py (hunk 40 above)
Comment (sameera2004) [inline skxray/correlation.py]: Sure

**Comment by sameera2004** [comment]
Fixed the code, ready for review.

**Comment by danielballan** [comment]
Great. I'll take a look at that last function first thing tomorrow.

**Comment by sameera2004** [comment]
sure, Thanks

**Comment by danielballan** [comment]
Actually `_process` looks about as clear as it can be. There's only some much one can do to clarify at the level. I am +1 for merging. Others?

**Comment by danielballan** [comment]
Actually, I noticed some minor issues introduced since the review, so those should be fixed before merging. Mostly spelling stuff -- should be easy to revert. I also noticed that you removed the pandas bit without comment.

Did it break your tests? If you need the two pieces separately, you could use `np.asarray(df)` and `np.asarray(df.index)` or, equivalently, `df.values` and `df.index.values`. Whichever makes more sense to you.

I don't want to steamroll you on this decision, but I think pandas is the way to go here. Just let me know what you are thinking.

**Comment by sameera2004** [comment]
Yes I wanted to tell you @danielballan , about pandas, it broke tests, not the local but Travis

**Comment by sameera2004** [comment]
Other thing I thought the user will need to know about pandas when they want to plot (lag steps vs g2). Therefore, I thought to better to keep 
return g2, lag_steps

**Comment by sameera2004** [comment]
@danielballan If I keep np.asarray(df) and np.asarray(df.index) then user will be able to plot (lag_steps vs g2) ?

**Comment by danielballan** [comment]
> Yes I wanted to tell you @danielballan , about pandas, it broke tests, not the local but Travis

I guess this is our first use of pandas in scikit-xray. I am sure, whatever we decide here, that will use pandas somewhere in the package. Let's add it to Travis by adding `pandas` to the list of dependencies on [this line](https://github.com/Nikea/scikit-xray/blob/master/.travis.yml#L17).

> even locally it took some time to run the tests.

Are you sure that pandas slowed it down? The cost of putting the data into a DataFrame should very small compared to the cost of running the algorithm. I could be wrong, but it would take actual measured times to convince me. :- D

> Other thing I thought the user will need to know about pandas when they want to plot (lag steps vs g2). Therefore, I thought to better to keep `return g2, lag_steps`

That's true. I think it will be easier to decide whether we want to require users to know about pandas or not once we have a couple more algorithms in scikit-xray and can see how much convenience pandas might add. We should not make a *final* decision at this early stage, so whatever we do with this PR is just provisional.

> @danielballan If I keep np.asarray(df) and np.asarray(df.index) then user will be able to plot (lag_steps vs g2) ?

I'm not sure what exactly you mean by "if I keep". If the user get a DataFrame back, they can plot by doing any of the following

````
df.plot()  # so easy, but Tom hates this
plt.plot(df.index, df)  # probably the best way

# or, go back to numpy...
g2, lag_steps = np.asarray(df), np.asarray(df.index)  # numpy arrays
plt.plot(lag_steps, g2)
```

Maybe leave this as g2, lagtimes for now but stay open to the possibility of making the output a DataFrame in the near- or medium-future.

sameera2004:
Yes I think it is better for now to keep g2, lag_steps and later change it to output a DataFrame.

sameera2004:
fixed the spelling mistake and added more clarification to level in  _process function ?
, ready for review

danielballan:
Looks good to me.

sameera2004:
Any more comments?

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+    # buffer. This algorithm only keeps number of buffers and delays but
+    # several levels of delays number of levels are kept in buf. Each
+    # level has twice the delay times of the next lower one. To save
+    # needless copying, of cyclic storage of images in buf is used.
+
+    if num_bufs % 2 != 0:
+        raise ValueError("number of channels(number of buffers) in "
+                         "multiple-taus (must be even)")
+
+    if labels.shape == images[0].shape[1:]:
```
tacaswell: This does not look right to me.  The first element in `images` should be a 2D array, and `img.shape[1:]` should be a length 1 tuple which can not be equal to a length 2 tuple

tacaswell: also, the logic on this is backwards, it should be `!=`?

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+                         "num_pixels = {0}".format(num_pixels))
+
+    # G holds the un normalized auto-correlation result. We
+    # accumulate computations into G as the algorithm proceeds.
+    G = np.zeros(((num_levels + 1)*num_bufs/2, num_rois),
+                 dtype=np.float64)
+
+    # matrix of past intensity normalizations
+    past_intensity_norm = np.zeros(((num_levels + 1)*num_bufs/2, num_rois),
+                   dtype=np.float64)
```
tacaswell: nitpick: indentation

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+
+    # matrix of future intensity normalizations
+    future_intensity_norm = np.zeros(((num_levels + 1)*num_bufs/2, num_rois),
+                   dtype=np.float64)
```
tacaswell: nitpick indentation

skxray/correlation.py
tacaswell: And can you add a test to exercise this failing?

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+    if num_bufs % 2 != 0:
+        raise ValueError("number of channels(number of buffers) in "
+                         "multiple-taus (must be even)")
+
+    if labels.shape == images[0].shape[1:]:
+        raise ValueError("Shape of an image should be equal to "
+                         "shape of the labels array")
+
+    # Which pixels are in each label?
+    labels, pixel_list = extract_label_indices(labels)
```
tacaswell: can you rename the input `labels` to something like `label_mask`?  re-assigning a very different data structure to the same name is a code tick that drives me up the wall.

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+
+    if num_bufs % 2 != 0:
+        raise ValueError("number of channels(number of buffers) in "
+                         "multiple-taus (must be even)")
+
+    if labels.shape == images[0].shape[1:]:
+        raise ValueError("Shape of an image should be equal to "
+                         "shape of the labels array")
+
+    # Which pixels are in each label?
```
tacaswell: Shouldn't be leaving questions in comments.  I hope you know what this is doing!

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+                "".format(n, (end_time - start_time)))
+
+    # the normalization factor
+    if len(np.where(past_intensity_norm == 0)[0]) != 0:
+        g_max = np.where(past_intensity_norm == 0)[0][0]
+    else:
+        g_max = past_intensity_norm.shape[0]
+
+    # g2 is normalized G
+    g2 = (G[: g_max] / (past_intensity_norm[: g_max] *
```
tacaswell: can you remove the space between `: g_max` -> `[:g_max]`

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+                        future_intensity_norm[: g_max]))
+
+    # Convert from num_levels, num_bufs to lag frames.
+    tot_channels, lag_steps = core.multi_tau_lags(num_levels,
+                                                  num_bufs)
```
tacaswell: nitpick: unwrap this line

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+
+def extract_label_indices(labels):
+    """
+    This will find the label's required region of interests (roi's),
+    number of roi's count the number of pixels in each roi's and pixels
+    list for the required roi's.
+
+    Parameters
+    ----------
+    labels : array
```
tacaswell: Please rename the input variable for the same reason as above.  It makes debugging code you are not familiar with super annoying.

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+    for n, img in enumerate(images):
+
+        cur[0] = (1 + cur[0]) % num_bufs  # increment buffer
+
+        # Put the image into the ring buffer.
+        buf[0, cur[0] - 1] = (np.ravel(img))[pixel_list]
+
+        # Compute the correlations between the first level
+        # (undownsampled) frames. This modifies G,
+        # past_ and future_intensity_norm, and img_per_level
```
tacaswell: one too many ands

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+        t_index = level*num_bufs/2 + i
+
+        delay_no = (buf_no - i) % num_bufs
+
+        past_img = buf[level, delay_no]
+        future_img = buf[level, buf_no]
+
+        #  get the matrix of auto-correlation function without normalizations
+        tmp_binned = (np.bincount(labels,
+                                  weights=np.ravel(past_img*future_img))[1:])
```
tacaswell: I don't think you need this ravel as the data is raveled on L175

```python
tmp_binned = np.bincount(labels, weights=past_img*future_img)[1:]
```

should work fine and is more readable.

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+        G[t_index] += ((tmp_binned/num_pixels - G[t_index])/
+                       (img_per_level[level] - i))
+
+        # get the matrix of past intensity normalizations
+        pi_binned = (np.bincount(labels,
+                                 weights=np.ravel(past_img))[1:])
```
tacaswell: same comment about ravel

skxray/correlation.py
```diff
@@ -0,0 +1,371 @@
...
+        past_intensity_norm[t_index] += ((pi_binned/num_pixels
+                                         - past_intensity_norm[t_index])/
+                                         (img_per_level[level] - i))
+
+        # get the matrix of future intensity normalizations
+        fi_binned = (np.bincount(labels,
```
tacaswell: same comment about ravel

skxray/tests/test_correlation.py
```diff
@@ -0,0 +1,96 @@
...
+        coins_stack.append(coins)
+
+    coins_mesh = np.zeros_like(coins)
+    coins_mesh[coins < 30] = 1
+    coins_mesh[coins > 50] = 2
+
+    g2, lag_steps = corr.multi_tau_auto_corr(num_levels, num_bufs,
+                                  coins_mesh, np.asarray(coins_stack))
+
+    assert_almost_equal(True, np.all(g2[:, 0], axis = 0))
```
tacaswell: This is only testing that all the returned values are non-zero

PR #257, opened 2015-03-28
CDI shrinkwrap

licode:
Basic functions for Shrinkwrap algorithm in CDI is included in this PR. The aim of the current CDI code is for both 2D and 3D reconstruction.

As CSX will have data soon for image retrieval, CDI method needs to be ranked in high priority, per @stuwilkins 's request. I break the whole task into smaller PRs, and will send them soon.

### DIFF
--- a/skxray/cdi.py
+++ b/skxray/cdi.py
@@ -66,7 +66,7 @@ def _dist(dims):
     dist_sum = []
     shape = np.ones(len(dims))
     for idx, d in enumerate(dims):
-        vec = (np.arange(d) - (d-1)/2) ** 2
+        vec = (np.arange(d) - d/2) ** 2
         shape[idx] = -1
         vec = vec.reshape(*shape)
         shape[idx] = 1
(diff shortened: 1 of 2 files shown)


licode:
This work is a combination of my previous IDL code and the work of Xiaojing Huang in HXN. And this PR is ready to review.

licode:
https://github.com/Nikea/scikit-xray/issues/258

ericdill:
Looks good!  :+1:

tacaswell:
What are the boundary condition differences? The `mode` switch does not cover the needed use-case?

licode:
No, I tried different mode, and we usually use mode='same' for same size return.

ericdill:
@tacaswell see https://github.com/Nikea/scikit-xray/issues/258

skxray/cdi.py
```diff
@@ -94,3 +94,29 @@ def gauss(dims, sigma):
...
+def convolution(array1, array2):
+    """
+    Calculate convolution of two arrays. Transfer into q space to perform the calculation.
+
+    Parameters
+    ----------
+    array1 : array
+        The size of array1 needs to be normalized.
+    array2 : array
+        The size of array2 keeps the same
```
danielballan: keeps the same what?

PR #261, opened 2015-04-09
Param correction

licode:
Correction on some bugs/API.

### DIFF
--- a/skxray/fitting/base/parameter_data.py
+++ b/skxray/fitting/base/parameter_data.py
@@ -267,8 +267,9 @@ default_param = {'coherent_sct_amplitude': {'bound_type': 'none',
                   'tool_tip': 'width**2 = (b1/2.3548)**2 + 3.85*b2*E',
                   'value': 0.178},
                  'non_fitting_values': {'element_list': 'Ar, Fe, Ce_L, Pt_M',
-                  'energy_bound_high': 12.0,
-                  'energy_bound_low': 2.5}}
+                  'energy_bound_low': {'value': 1.5, 'default_value': 1.5, 'description': 'E low [keV]'},
+                  'energy_bound_high': {'value': 13.5, 'default_value': 13.5, 'description': 'E high [keV]'}}
+}
 
 
 def get_para():
--- a/skxray/fitting/base/parameter_data.py
+++ b/skxray/fitting/base/parameter_data.py
@@ -266,9 +266,12 @@ default_param = {'coherent_sct_amplitude': {'bound_type': 'none',
                   'min': 0.16,
                   'tool_tip': 'width**2 = (b1/2.3548)**2 + 3.85*b2*E',
                   'value': 0.178},
-                 'non_fitting_values': {'element_list': 'Ar, Fe, Ce_L, Pt_M',
-                  'energy_bound_low': {'value': 1.5, 'default_value': 1.5, 'description': 'E low [keV]'},
-                  'energy_bound_high': {'value': 13.5, 'default_value': 13.5, 'description': 'E high [keV]'}}
+                 'non_fitting_values': {
+                     'element_list': 'Ar, Fe, Ce_L, Pt_M',
+                     'energy_bound_low': {'value': 1.5, 'default_value': 1.5, 'description': 'E low [keV]'},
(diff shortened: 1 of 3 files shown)


skxray/fitting/base/parameter_data.py
```diff
@@ -266,9 +266,13 @@
                   'min': 0.16,
                   'tool_tip': 'width**2 = (b1/2.3548)**2 + 3.85*b2*E',
                   'value': 0.178},
-                 'non_fitting_values': {'element_list': 'Ar, Fe, Ce_L, Pt_M',
-                  'energy_bound_high': 12.0,
-                  'energy_bound_low': 2.5}}
+                 'non_fitting_values': {
+                     'element_list': 'Ar, Fe, Ce_L, Pt_M',
+                     'energy_bound_low': {'value': 1.5, 'default_value': 1.5, 'description': 'E low [keV]'},
+                     'energy_bound_high': {'value': 13.5, 'default_value': 13.5, 'description': 'E high [keV]'},
```
ericdill: PEP8 this line and the one above

skxray/fitting/xrf_model.py
```diff
@@ -1059,15 +1076,12 @@ def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
     # ratio to transfer energy value back to channel value
     approx_ratio = 100
```
ericdill: where does this number come from?

skxray/fitting/xrf_model.py
```diff
@@ -1085,21 +1099,31 @@ def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
...
+        return x, result_dict, area_dict
+    else:
+        result_dict = OrderedDict()
+
+        for i in range(len(total_list)):
+            if np.sum(total_y[:, i]) == 0:
+                continue
+            result_dict.update({total_list[i]: total_y[:, i]})
+        result_dict.update(background=bg)
+        return x, result_dict
```
ericdill: This seems very strange to have optional return values. I would suggest using a different pattern

skxray/fitting/base/parameter_data.py
```diff
@@ -266,9 +266,13 @@
...
+                     'electron_hole_energy': 3.51
```
ericdill: maybe change this to `epsilon` and add a comment that it is the electron_hole_energy.

skxray/fitting/xrf_model.py
licode: The default transfer from channel to energy is energy = 0.01*channel number. So to transfer back is just 100.

skxray/fitting/xrf_model.py
licode: So let's do two functions instead?

skxray/fitting/base/parameter_data.py
licode: Good idea. will do.

danielballan:
It looks like the function that returns a two-element tuple or a three-element tuple depending on its inputs was merged. I agree with @ericdill that that is trouble. Can you make refactoring that a priority?

skxray/fitting/xrf_model.py
```diff
@@ -940,14 +951,16 @@ def construct_linear_model(channel_number, params,
...
 
     p = MS.elastic.make_params()
     y_temp = MS.elastic.eval(x=channel_number, params=p)
     matv.append(y_temp)
+    element_area.update({'elastic': p['elastic_coherent_sct_amplitude'].value})
 
     matv = np.array(matv)
     matv = matv.transpose()
-    return selected_elements, matv
+    return selected_elements, matv, element_area
```
tacaswell: did this api change get documented?

licode:
The format of function returns keeps fixed now, and documentation is added.

ericdill:
@danielballan Li and I made some changes to the codebase during code review and added some commits to this PR.  I then merged it locally and pushed to master with the new commits.

See https://github.com/Nikea/scikit-xray/commit/bddfde865687466467c62379d3aa8e1373b3ced3 in particular, which addresses your concern

ericdill:
@tacaswell no, it did not get documented, I will do that now

PR #264, opened 2015-04-12
BUG/TST: linear_spectrum_fitting function in test file

licode:
Fix a bug in test file due to API change on Friday.

### DIFF
--- a/skxray/fitting/xrf_model.py
+++ b/skxray/fitting/xrf_model.py
@@ -1031,8 +1031,9 @@ def weighted_nnls_fit(spectrum, expected_matrix, constant_weight=10):
     return results, residue
 
 
-def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
-                            constant_weight=10, area_option=False):
+def linear_spectrum_fitting(spectrum, params,
+                            elemental_lines=None,
+                            constant_weight=10):
     """
     Fit a spectrum to a linear model.
 
@@ -1053,8 +1054,6 @@ def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
         value used to calculate weight like so:
         weights = constant_weight / (constant_weight + spectrum)
         Default is 10. If None, performed unweighted nnls fit.
-    area_option : Bool
-        return the area of the first gaussian that corresponds to each element.
 
     Returns
     -------
@@ -1063,8 +1062,7 @@ def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
     result_dict : dict
         Fitting results
     area_dict : dict
-        Returns None for the area if area_option=False
-        Returns the area if area_option=True
+        the area of the first main peak, such as Ka1, of a given element
     """
     if elemental_lines is None:
         elemental_lines = K_LINE + L_LINE + M_LINE
@@ -1102,6 +1100,8 @@ def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
         out, res = nnls_fit(y, matv)
 
     total_y = out * matv
+    total_y = np.transpose(total_y)
+
     x = (params['e_offset']['value'] +
          params['e_linear']['value']*x +
          params['e_quadratic']['value'] * x**2)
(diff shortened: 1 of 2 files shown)


tacaswell:
What commit broke this?

licode:
https://github.com/Nikea/scikit-xray/commit/bddfde865687466467c62379d3aa8e1373b3ced3

licode:
Change area to None.

skxray/fitting/xrf_model.py
```diff
@@ -1117,12 +1116,12 @@ def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
         if np.sum(total_y[:, i]) == 0:
             continue
         result_dict.update({total_list[i]: total_y[:, i]})
-        area = 0
+        area = None
         if area_option:
             area = out[i]*element_area[total_list[i]]
```
tacaswell: This computation should be _very_ cheap.

skxray/fitting/xrf_model.py
```diff
@@ -1117,12 +1116,12 @@ def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
         if np.sum(total_y[:, i]) == 0:
```
tacaswell: You should use `for i, func in enumerate(total_list):` which will reduce the amount of slicing you have to do below.  With clever application of zip could probably get rid of the slicing all together.

skxray/fitting/xrf_model.py
```diff
@@ -1117,12 +1116,12 @@ def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
...
+        area = None
         if area_option:
             area = out[i]*element_area[total_list[i]]
         area_dict.update({total_list[i]: area})
     result_dict.update(background=bg)
-    bg = 0
+    bg = None
     if area_option:
         bg = np.sum(bg)
     area_dict.update(background=bg)
```
tacaswell: why are you using `update` instead of `area_dict['background'] = bg`?

skxray/fitting/xrf_model.py
```diff
@@ -1117,12 +1116,12 @@ def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
...
-        area = 0
+        area = None
         if area_option:
             area = out[i]*element_area[total_list[i]]
         area_dict.update({total_list[i]: area})
     result_dict.update(background=bg)
-    bg = 0
+    bg = None
     if area_option:
         bg = np.sum(bg)
```
tacaswell: This is a no-op as you re-assign bg to be `None` just above this.

skxray/fitting/xrf_model.py
licode: This function is also used in single pixel fitting, so I prefer to reduce unnecessary calculation.

skxray/fitting/xrf_model.py
licode: Any major difference?

tacaswell: I suspect that `[]` is marginally faster

skxray/fitting/xrf_model.py
tacaswell: The `if` block also isn't free.

skxray/fitting/xrf_model.py
licode: Ah, good to know. I sometimes feel confused in choosing which one to use.

skxray/fitting/xrf_model.py
licode: OK, I will run some tests, then make the decision.

tacaswell: If I were going to optimize this function, this is not where I would start.

Also, don't start optimizing until you know a) that it matters b) where the bottle neck is.

tacaswell:
If you are always going to reference `total_y` by extracting columns, I strongly suggest transposing it before working on it so that you can pull out rows instead.  The memory access patterns are much better.

licode:
Good to know. Thanks. I will send the correction soon. 

BTW, why accessing row faster? Is it because the underlining pointer structure, such as pointer referencing to row first, then to each item of the row?

tacaswell:
Yes, by default numpy arrays are stored in memory as row-major (but you can ask them to be column major if you want as well).

It does not matter too much for small arrays (when the whole thing fits in cache) but once you get bigger you can start to get awful page hits (as you need to go back to main memory for _every_ element rather than every row).  I also suspect that there is a way to vertorize this loop.

licode:
Done. 

Persuaded by Data structure is more central to programming. :+1:  https://twitter.com/igrigorik/status/587446362880344064?s=15

skxray/fitting/xrf_model.py
```diff
@@ -1102,30 +1100,23 @@ def linear_spectrum_fitting(spectrum, params, elemental_lines=None,
...
-        if np.sum(total_y[:, i]) == 0:
+    for i, func in enumerate(total_list):
+        if np.sum(total_y[i, :]) == 0:
             continue
-        result_dict.update({total_list[i]: total_y[:, i]})
-        area = 0
-        if area_option:
-            area = out[i]*element_area[total_list[i]]
+        result_dict.update({total_list[i]: total_y[i, :]})
+        area = out[i]*element_area[total_list[i]]
```
tacaswell: you don't need to slice into total_list, you have it as `func`

tacaswell:
```python
for name, y, _out in izip(total_list, total_y, out):
    if not np.sum(y):
        continue
    result_dict[name] = y
    aread_dict[name] = _out * element_area[name]
```

licode:
Corrections added.

PR #248, opened 2015-02-27
Divide the circle : find the indices, number of pixels, pixel indices 

sameera2004:

PR #267, opened 2015-04-15
Fix diff roi

sameera2004:
Fixed the functions in the diff_roi_choice.py after changes in the correlation.py

### DIFF
--- a/skxray/diff_roi_choice.py
+++ b/skxray/diff_roi_choice.py
@@ -525,3 +525,45 @@ def extract_label_indices(labels, rotate='N'):
     label_mask = labels[labels > 0]
 
     return label_mask, indices
+
+
+def _get_roi_grid(detector_size, calibrated_center, num_angles):
+    """
+    Helper function to get the grid values and indices values from
+    the angles of the grid
+
+    Parameters
+    ----------
+    detector_size : tuple
+        shape of the image (detector X and Y direction)
+        Order is (num_rows, num_columns)
+
+    calibrated_center : tuple
+        defining the center of the image
+        (column value, row value) (mm)
+
+    Returns
+    -------
+    ind_grid : array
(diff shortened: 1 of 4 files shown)


sameera2004:
@tacaswell I think I have to close the PR #248 Divide the circle : find the indices, number of pixels, pixel indices. I fixed this new PR, I fixed all the functions in this , the test are passing but there is a problem with the coverage,(missed 7 lines)

danielballan:
Suggested name changes:

`detector_size`, `img_shape`, etc. -> `shape`
`roi_data` -> `roi_specs` and remove `num_rois`, which we can infer from the length of `roi_specs`

danielballan:
Scikit-image has fast functions based on published CS literature for drawing rings: http://scikit-image.org/docs/dev/api/skimage.draw.html#skimage.draw.circle

skxray/diff_roi_choice.py
```diff
@@ -116,18 +109,15 @@ def roi_rectangles(num_rois, roi_data, detector_size):
...
-    w = np.where(mesh > 0)
-    grid = np.indices((detector_size[0], detector_size[1]))
-    pixel_list = ((grid[0]*detector_size[1] + grid[1]))[w]
+    roi_labels = mesh[mesh > 0]
 
-    roi_inds = mesh[mesh > 0]
+    roi_labels, indices = extract_label_indices(mesh)
 
-    return mesh.reshape(detector_size), roi_inds, num_pixels, pixel_list
+    return mesh.reshape(detector_size), roi_labels, indices
```
danielballan: Only return the first bit, not the second two.

tacaswell:
What is this PR?

danielballan:
Superceded by Sameera's #269, my #272, and a forthcoming PR that Sameera and I are pair-coding currently....

PR #269, opened 2015-04-23
ENH: changed the functions in this module(diff_roi_choice) to easily use...

sameera2004:
... in one time correlation

so now roi functions only returns the labels array

### DIFF
--- a/skxray/diff_roi_choice.py
+++ b/skxray/diff_roi_choice.py
@@ -52,11 +52,13 @@ import sys
 import numpy as np
 import numpy.ma as ma
 
+import skxray.correlation as corr
+
 import logging
 logger = logging.getLogger(__name__)
 
 
-def roi_rectangles(num_rois, roi_data, detector_size):
+def rectangles(num_rois, roi_data, image_shape):
     """
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


skxray/diff_roi_choice.py
```diff
@@ -68,66 +70,48 @@ def roi_rectangles(num_rois, roi_data, detector_size):
         from those co-ordinates
         shape is [num_rois][4]
 
-    detector_size : tuple
+    image_shape : tuple
         2 element tuple defining the number of pixels in the detector.
         Order is (num_rows, num_columns)
 
     mask : ndarray, optional
```
tacaswell: remove this secino of docs.

skxray/diff_roi_choice.py
```diff
@@ -68,66 +70,48 @@ def roi_rectangles(num_rois, roi_data, detector_size):
...
     Returns
     -------
-    roi_inds : ndarray
-        indices of the required shape
-         after discarding zero values from the shape
-        ([detector_size[0]*detector_size[1]], )
-
-    num_pixels : ndarray
-        number of pixels in certain rectangle shape
+    labels_grid : array
```
tacaswell: I think we are standardizing on 'label_array'

skxray/diff_roi_choice.py
```diff
@@ -157,58 +141,61 @@ def roi_rings(img_dim, calibrated_center, num_rings,
 
     Returns
     -------
-    mesh : nedarry
+    labels_grid : array
         indices of the required rings
-        shape is ([detector_size[0]*detector_size[1]], )
+        shape is ([image_shape[0]*image_shape[1]], )
```
tacaswell: This shape is wrong, should be '`image_shape`'

skxray/diff_roi_choice.py
```diff
@@ -157,58 +141,61 @@ def roi_rings(img_dim, calibrated_center, num_rings,
...
+    labels_grid[labels_grid > num_rings] = 0
 
-    num_pixels : ndarray
-        number of pixels in each ring
+    return labels_grid.reshape(image_shape)
 
-    pixel_list : ndarray
-        pixel indices for the required rings
 
+def rings_edges(num_rings, first_r, delta_r):
```
tacaswell: `rings_from_edges` as a name?  It returns the full rings, not just the edges.

skxray/diff_roi_choice.py
```diff
@@ -157,58 +141,61 @@ def roi_rings(img_dim, calibrated_center, num_rings,
...
+    # edges of all the rings
+    ring_edges = np.linspace(first_r, last_r, num=(num_rings+1))
 
-    return mesh.reshape(img_dim), ring_inds, ring_vals, num_pixels, pixel_list
+    return ring_edges
 
 
-def roi_rings_step(img_dim, calibrated_center, num_rings,
-               first_r, delta_r, *step_r):
+def rings_step(image_shape, calibrated_center, num_rings, first_r, delta_r,
```
tacaswell: drop `num_rings` and this will be equal to the length of `step_r`.

skxray/diff_roi_choice.py
```diff
@@ -157,58 +141,61 @@ def roi_rings(img_dim, calibrated_center, num_rings,
...
+               *step_r):
```
tacaswell: why is `step_r` a variadic?  It should probably be an list/array so that the call signature looks like 

```python
rings = rings_step(shape, center, first_r, delta_r, [1, 1, 1, 2, 1, 1])
```

rather than

```python
rings = rings_step(shape, center, first_r, delta_r, 1, 1, 1, 2, 1, 1)
```

skxray/diff_roi_choice.py
```diff
@@ -245,25 +225,58 @@ def roi_rings_step(img_dim, calibrated_center, num_rings,
...
+    labels_grid[indx] = (labels_grid[indx] + 1) // 2
 
-    num_pixels : ndarray
-        number of pixels in each ring
+    return labels_grid.reshape(image_shape)
 
-    pixel_list : ndarray
-        pixel indices for the required rings
 
+def rings_step_edges(num_rings, first_r, delta_r, *step_r):
```
tacaswell: How is this different than the previous function?

skxray/diff_roi_choice.py
```diff
@@ -289,97 +302,47 @@ def roi_rings_step(img_dim, calibrated_center, num_rings,
...
-    y_ = (yy - calibrated_center[0])
-    grid_values = np.float_(np.hypot(x_, y_))
+    ring_vals = np.asarray(ring_vals).reshape(-1, 2)
 
-    return grid_values
+    return ring_vals
 
 
-def _process_rings(num_rings, img_dim, ring_vals, ring_inds):
+def _grid_values(image_shape, calibrated_center):
```
tacaswell: This is returning an array of radius, I thought we had this function is `skxray.core`?

skxray/diff_roi_choice.py
sameera2004: This will use when there is a step value between rings

skxray/diff_roi_choice.py
sameera2004: No in this we can either give one value between rings or different step value between rings

skxray/diff_roi_choice.py
sameera2004: Sure, took it out

PR #273, opened 2015-04-26
Gendered Language: Feature or Bug in Software Documentation? 

ericdill:
I was inspired by https://modelviewculture.com/pieces/gendered-language-feature-or-bug-in-software-documentation.

This PR adds tests to all docstrings in our library to prevent gendered language from leaking in

### DIFF
--- a/contributing.txt
+++ b/contributing.txt
@@ -44,7 +44,7 @@ Here's the long and short of it:
    * Push your changes back to your fork on GitHub::
 
       git push origin transform-speedups
-      
+
   * **Explain your contribution.** Include a description aimed at someone
      who may not be following the issue closely. If your contribution
      includes new functionality or enhancements, include a link to an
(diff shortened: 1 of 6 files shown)


ericdill:
Also inspired by this https://modelviewculture.com/pieces/the-open-source-identity-crisis

danielballan:
For a practical review of ways to avoid gendered assumptions without resorting to contorted English, see Brian Garder, author of "Modern American Usage" in a series of three short posts. The most relevant is [this one](http://www.lawprose.org/blog/?p=499).  @ericdill maybe add a link to this PR?

danielballan:
Is full-on blacklisting the male pronoun is the right way to ensure openness in the library? Sometimes we might want to refer to the work of a specific researcher who has a gender. Maybe a warning would be better than a test failure.

tacaswell:
:+1: on this.

I am inclined to say we deal with the case where we _must_ use a pronoun to refer to a researcher in a docstring at a when it comes up.  I suspect that if we have enough text in the docstring that using a pronoun is required there is too much text and it should move off to the rst documents anyway.

danielballan:
It looks like there is some accidentally-committed code in here.

ericdill:
It is not accidentally committed.  This new test actually imports every module and apparently some of our relative imports were incorrect

ericdill:
@danielballan Added a link to the writing guidelines. Thanks for that

ericdill:
Also, this should really make use of regex, but I suck at it

PR #275, opened 2015-05-05
Pileup peaks added

licode:
In this PR, I added pileup peaks for fitting. Pileup peak is the combination of two different elemental lines.

### DIFF
--- a/skxray/fitting/base/parameter_data.py
+++ b/skxray/fitting/base/parameter_data.py
@@ -301,7 +301,8 @@ default_param = {
         'min': -1e-06,
         'tool_tip': 'E(channel) = a0 + a1*channel+ a2*channel**2',
         'value': 0.0},
-    'fwhm_fanoprime': {'bound_type': 'fixed',
+    'fwhm_fanoprime': {
+        'bound_type': 'fixed',
         'description': 'fwhm Coef, b2',
         'max': 0.0001,
         'min': 1e-07,
@@ -323,7 +324,8 @@ default_param = {
             'value': 13.5,
             'default_value': 13.5,
             'description': 'E high [keV]'},
-        'epsilon': 3.51 # electron hole energy
+        'epsilon': 3.51,  # electron hole energy
+        "background_width": 0.5
     }
 }
 
--- a/skxray/fitting/base/parameter_data.py
+++ b/skxray/fitting/base/parameter_data.py
@@ -237,7 +237,8 @@ default_param = {
         'bound_type': 'lohi',
         'description': 'Incident E [keV]',
         'max': 13.0,
-        'min': 9.0, 'value': 11.0},
+        'min': 9.0,
+        'value': 10.0},
     'compton_amplitude': {
         'bound_type': 'none',
         'max': 10000000.0,
--- a/skxray/fitting/models.py
+++ b/skxray/fitting/models.py
@@ -128,7 +128,7 @@ class ElasticModel(Model):
 
     def __init__(self, *args, **kwargs):
         super(ElasticModel, self).__init__(elastic, *args, **kwargs)
-        set_default(self, elastic)
+        #set_default(self, elastic)
         self.set_param_hint('epsilon', value=2.96, vary=False)
 
 
@@ -139,7 +139,7 @@ class ComptonModel(Model):
     def __init__(self, *args, **kwargs):
 
         super(ComptonModel, self).__init__(compton, *args, **kwargs)
-        set_default(self, compton)
+        #set_default(self, compton)
         self.set_param_hint('epsilon', value=2.96, vary=False)
 
 
--- a/skxray/fitting/xrf_model.py
+++ b/skxray/fitting/xrf_model.py
@@ -537,34 +537,19 @@ class ModelSpectrum(object):
         """
         setup parameters related to Elastic model
         """
+        param_hints_elastic = ['e_offset', 'e_linear', 'e_quadratic',
+                               'fwhm_offset', 'fwhm_fanoprime', 'coherent_sct_energy']
+
         elastic = ElasticModel(prefix='elastic_')
 
         logger.debug('Started setting up parameters for elastic model')
 
-        # set constraints for the following global parameters
-        elastic.set_param_hint('e_offset',
-                               value=self.compton_param['e_offset'].value,
-                               expr='e_offset')
-        elastic.set_param_hint('e_linear',
-                               value=self.compton_param['e_linear'].value,
-                               expr='e_linear')
-        elastic.set_param_hint('e_quadratic',
-                               value=self.compton_param['e_quadratic'].value,
-                               expr='e_quadratic')
-        elastic.set_param_hint('fwhm_offset',
-                               value=self.compton_param['fwhm_offset'].value,
-                               expr='fwhm_offset')
-        elastic.set_param_hint('fwhm_fanoprime',
-                               value=self.compton_param['fwhm_fanoprime'].value,
-                               expr='fwhm_fanoprime')
-        elastic.set_param_hint('coherent_sct_energy',
-                               value=self.compton_param['coherent_sct_energy'].value,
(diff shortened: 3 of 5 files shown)


skxray/fitting/base/parameter_data.py
```diff
@@ -237,7 +237,8 @@
         'bound_type': 'lohi',
         'description': 'Incident E [keV]',
         'max': 13.0,
-        'min': 9.0, 'value': 11.0},
+        'min': 9.0,
+        'value': 10.0},
```
tacaswell: Why the value change?

skxray/fitting/models.py
```diff
@@ -128,7 +128,7 @@ class ElasticModel(Model):
 
     def __init__(self, *args, **kwargs):
         super(ElasticModel, self).__init__(elastic, *args, **kwargs)
-        set_default(self, elastic)
+        #set_default(self, elastic)
```
tacaswell: Why did these get removed?

tacaswell: Or, if you are going to remove them, just delete them.

skxray/fitting/xrf_model.py
```diff
@@ -807,6 +816,98 @@ def setup_element_model(self, elemental_line, default_area=1e5):
...
+                else:
+                    # only the first line for M
+                    e_cen = e.emission_line['ma1']
+                return e_cen
+
+            e1_cen = get_line_energy(element_line1)
+            e2_cen = get_line_energy(element_line2)
+
+            # no '-' allowed in prefix name in lmfit
+            pre_name = 'pileup_' + elemental_line.replace('-', '_') + '_'
```
tacaswell: You should probably make a helper function that does all of this sanitation as this line of code seems to show up a bunch of times.

skxray/fitting/xrf_model.py
```diff
@@ -807,6 +816,98 @@ def setup_element_model(self, elemental_line, default_area=1e5):
                 else:
                     element_mod = gauss_mod
 
+        else:
+            logger.debug('Started setting up pileup peaks for {}'.format(
+                elemental_line))
+
+            element_line1, element_line2 = elemental_line.split('-')
+
+            def get_line_energy(elemental_line):
```
tacaswell: Why is this function defined here?  It does not look like you are closing over anything.

skxray/fitting/xrf_model.py
```diff
@@ -521,11 +532,9 @@ def setup_elastic_model(self):
                                value=self.compton_param['coherent_sct_energy'].value,
                                expr='coherent_sct_energy')
 
-        elastic.set_param_hint('coherent_sct_energy',
```
tacaswell: was this intended to be deleted?

tacaswell: I think you can replace this whole chunk of code with

```python
for label in list_of_labels:
    elastic.set_param_hint(label,value=self.compton_param[label], expr=label)
```

which is what it looks like you are doing in the next look anyway.

skxray/fitting/xrf_model.py
```diff
@@ -807,6 +816,98 @@ def setup_element_model(self, elemental_line, default_area=1e5):
...
+                    e_cen = e.emission_line['ma1']
+                return e_cen
+
+            e1_cen = get_line_energy(element_line1)
+            e2_cen = get_line_energy(element_line2)
+
+            # no '-' allowed in prefix name in lmfit
+            pre_name = 'pileup_' + elemental_line.replace('-', '_') + '_'
+            gauss_mod = ElementModel(prefix=pre_name)
+            gauss_mod.set_param_hint('e_offset',
```
tacaswell: the above comment about making this more readable.

skxray/fitting/xrf_model.py
```diff
@@ -807,6 +816,98 @@ def setup_element_model(self, elemental_line, default_area=1e5):
...
+            gauss_mod.set_param_hint('e_quadratic',
+                                     value=self.compton_param['e_quadratic'].value,
+                                     expr='e_quadratic')
+            gauss_mod.set_param_hint('fwhm_offset',
+                                     value=self.compton_param['fwhm_offset'].value,
+                                     expr='fwhm_offset')
+            gauss_mod.set_param_hint('fwhm_fanoprime',
+                                     value=self.compton_param['fwhm_fanoprime'].value,
+                                     expr='fwhm_fanoprime')
+            gauss_mod.set_param_hint('epsilon', value=self.epsilon, vary=False)
```
tacaswell: why doesn't this hove an expr?

skxray/fitting/xrf_model.py
```diff
@@ -807,6 +816,98 @@ def setup_element_model(self, elemental_line, default_area=1e5):
...
+
+            area_name = pre_name + 'area'
+            if area_name in self.params:
+                default_area = self.params[area_name]['value']
```
tacaswell: This will silently ignore user input.

skxray/fitting/base/parameter_data.py
licode: Typo, will change it back.

skxray/fitting/xrf_model.py
```diff
@@ -807,6 +816,98 @@ def setup_element_model(self, elemental_line, default_area=1e5):
...
+                    # only the first line for M
+                    e_cen = e.emission_line['ma1']
+                return e_cen
+
+            e1_cen = get_line_energy(element_line1)
+            e2_cen = get_line_energy(element_line2)
+
+            # no '-' allowed in prefix name in lmfit
+            pre_name = 'pileup_' + elemental_line.replace('-', '_') + '_'
+            gauss_mod = ElementModel(prefix=pre_name)
```
tacaswell: can we not call this `gaus_mod` if it is not a Gaussian?

skxray/fitting/models.py
licode: I will delete them. I don't want to set parameters in multiple places. This will cause some bugs.

skxray/fitting/xrf_model.py
licode: Good catch. This pattern does appear several times.

skxray/fitting/xrf_model.py
licode: I think model level function is a better choice for it.

skxray/fitting/xrf_model.py
licode: although it is just a local variable, sure.

skxray/fitting/xrf_model.py
licode: epsilon will always stay as a fixed value, not for fitting, so we don't need to use expr to set constraint.

skxray/fitting/xrf_model.py
tacaswell: Yes, but people will assume that if it has the name 'gaus' it is a Gaussian which makes it that much harder for someone else maintain this code in the future.

skxray/fitting/xrf_model.py
tacaswell: k, I was aggressively pattern matching.

skxray/fitting/xrf_model.py
```diff
@@ -214,9 +241,11 @@ def update_parameter_dict(param, fit_results):
         if k in elastic_list:
             k_temp = 'elastic_' + k
         else:
-            k_temp = k
+            k_temp = k.replace('-', '_')  # pileup peak, i.e., 'Cl_K-Cl_K'
         if k_temp in fit_results.values:
             param[k]['value'] = float(fit_results.values[k_temp])
+        elif k == 'non_fitting_values':
```
tacaswell: I thought that `non_fitting_values` had been removed all together.

licode: The current plan is to save all the values in one config file.

licode:
I think this is ready to go. @tacaswell ?

licode:
Ready to review, @tacaswell ?

PR #276, opened 2015-05-05
Two time correlation

sameera2004:


skxray/correlation.py
```diff
@@ -367,3 +344,337 @@ def extract_label_indices(labels):
     label_mask = labels[labels > 0]
 
     return label_mask, pixel_list
+
+
+def two_time_corr(num_levels, num_bufs, labels, images):
```
ericdill: I'm not sure the `_corr` is needed, as the module where this function resides is called `correlation`.  I think that the import would look nice and be just as descriptive if the `_corr` was dropped.  e.g.,  
```python
from skxray.correlation import two_time
```

Thoughts?

skxray/correlation.py
```diff
@@ -1,13 +1,16 @@
 # ######################################################################
-# Original code(in Yorick):                                            #
+# Copyright (c) 2014, Brookhaven Science Associates, Brookhaven        #
+# National Laboratory. All rights reserved.                            #
+#                                                                      #
+# multi_tau_auto_corr Original code(in Yorick):                        #
 # @author: Mark Sutton                                                 #
 #                                                                      #
+# two_time_corr Original code                                          #
+# @author: Yugang Zhang                                                #
```
ericdill: Do these attributions belong in the docstrings of `two_time_corr` and `multi_tau_auto_corr` instead of in the license header?

skxray/correlation.py
```diff
@@ -219,8 +204,8 @@ def multi_tau_auto_corr(num_levels, num_bufs, labels, images):
     # ending time for the process
     end_time = time.time()
 
-    logger.info("Processing time for {0} images took {1} seconds."
-                "".format(n, (end_time - start_time)))
+    logger.info("One Time Correlation - Processing time for {0} images took"
```
ericdill: I'm fairly certain that using old-style string formatting for log messages is appropriate for performance reasons

skxray/correlation.py
```diff
@@ -367,3 +344,337 @@ def extract_label_indices(labels):
...
+
+    num_bufs : int, must be even
+        maximum lag step to compute in each generation of
+        downsampling
+
+    labels : array
+        labeled array of the same shape as the image stack;
+        each ROI is represented by a distinct label (i.e., integer)
+
+    images : iterable of 2D arrays
```
ericdill: Is our vistrails autowrapper going to choke on this docstring?

skxray/correlation.py
```diff
@@ -367,3 +344,337 @@ def extract_label_indices(labels):
...
+    The longest lag time computed is num_levels * num_bufs.
+    ** see comments on multi_tau_auto_corr
+
+    Parameters
+    ----------
+    num_levels : int
+        how many generations of downsampling to perform, i.e.,
+        the depth of the binomial tree of averaged frames
+
+    num_bufs : int, must be even
```
ericdill: Same: "Is our vistrails autowrapper going to choke on this docstring?"

skxray/correlation.py
sameera2004: Yes, It(images : iterable of 2D arrays) is. I just checked. However, 
num_bufs : int, must be even 
is fine, can be used as this way

is fine

skxray/correlation.py
sameera2004: Sure, yes I think it is better in this way

skxray/correlation.py
tacaswell: yes

ericdill:
@sameera2004 what is the development status of this branch?

sameera2004:
@ericdill Yes this is my next plan, Right now it is broken, I will fix this and discuss with Andrei and Yugang. Looks like CHX will not need it soon, may be CSX will need soon?
But before that I have to take a look at the XSVS fitting, I think the scipy.stats module(you suggested) has all the fitting functions. I am going to use them and see the results and discuss with Andrei.

sameera2004:
Closing this due to merge conflicts, New PR is added for two time correlation

PR #279, opened 2015-05-11
Cdi recon

licode:
The full version of cdi reconstruction is added.

### DIFF
--- a/skxray/cdi.py
+++ b/skxray/cdi.py
@@ -259,148 +259,182 @@ def cal_diff_error(sample_obj, diff_array):
     return cal_relative_error(diff_array, new_diff)
 
 
-class CDI(object):
+def generate_random_phase_field(diff_array):
+    """
+    Initiate random phase.
 
-    def __init__(self, diffamp, **kwargs):
-        """
-        Parameters
-        ----------
-        deffmap : array
-            diffraction pattern from experiments
-        kwargs : dict
-            parameters related to cdi reconstruction
-        """
+    Parameters
+    ----------
+    diff_array : array
+        diffraction pattern
 
-        self.diff_array = np.array(diffamp)     # diffraction data
+    Returns
+    -------
+    sample_obj : array
+        sample information with phase
+    """
+    pha_tmp = np.random.uniform(0, 2*np.pi, diff_array.shape)
+    sample_obj = (_ifft_helper(diff_array * np.exp(1j*pha_tmp))
+                  * np.sqrt(np.size(diff_array)))
+    return sample_obj
 
-        self.ndim = self.diff_array.ndim    # 2D or 3D case
-        if self.ndim == 2:
-            self.nx, self.ny = np.shape(self.diff_array)    # array dimension
-        if self.ndim == 3:
-            self.nx, self.ny, self.nz = np.shape(self.diff_array)    # array dimension
 
-        self.pi_modulus_flag = kwargs['pi_modulus_flag']  # 'Complex'
+def generate_box_support(sup_radius, shape_v):
+    """
+    Generate support area as a box for either 2D or 3D cases.
 
-        self.beta = kwargs['beta']  # feedback parameter for difference map algorithm, around 1.15
-        self.start_ave = kwargs['start_ave']  # 0.8
+    Parameters
+    ----------
+    sup_radius : float
+        radius of support
+    shape_v : list
+        shape of diffraction pattern, which can be either 2D or 3D case.
 
-        # initial support
-        self.init_obj_flag = kwargs['init_obj_flag']
-        self.init_sup_flag = kwargs['init_sup_flag']
-        self.sup_radius = kwargs['support_radius']
-        self.sup_shape = kwargs['support_shape']  # 'Box' or 'Disk'
+    Returns
+    -------
+    sup : array
+        support with a box area
+    """
+    sup = np.zeros(shape_v)
+    if len(shape_v) == 2:
+        nx, ny = shape_v
+        sup[nx/2-sup_radius: nx/2+sup_radius,
+            ny/2-sup_radius: ny/2+sup_radius] = 1
+    elif len(shape_v) == 3:
+        nx, ny, nz = shape_v
+        sup[nx/2-sup_radius: nx/2+sup_radius,
+            ny/2-sup_radius: ny/2+sup_radius,
+            nz/2-sup_radius: nz/2+sup_radius] = 1
+    return sup
 
-        # parameters for shrink wrap
-        self.shrink_wrap_flag = kwargs['shrink_wrap_flag']
-        self.sw_sigma = kwargs['sw_sigma']  # 0.5
-        self.sw_threshold = kwargs['sw_threshold']  # 0.1
-        self.sw_start = kwargs['sw_start']  # 0.2
-        self.sw_end = kwargs['sw_end']  # 0.8
-        self.sw_step = kwargs['sw_step']  # 10
 
-    def init_obj(self):
-        """
-        Initiate phase. Focus on 2D case here, and to be extended.
-        """
-        pha_tmp = np.random.uniform(0, 2*np.pi, (self.nx, self.ny))
-        self.obj = (_ifft_helper(self.diff_array * np.exp(1j*pha_tmp))
-                    * np.sqrt(np.size(self.diff_array)))
+def generate_disk_support(sup_radius, shape_v):
+    """
+    Generate support area as a disk for either 2D or 3D cases.
 
-    def init_sup(self):
-        """ Initiate sample support.
-        """
-        self.sup = np.zeros_like(self.diff_array)
-        if self.ndim == 2:
-            if self.sup_shape == 'Box':
-                self.sup[self.nx/2-self.sup_radius: self.nx/2+self.sup_radius,
-                         self.ny/2-self.sup_radius:self.ny/2+self.sup_radius] = 1
-        if self.ndim == 3:
-            if self.sup_shape == 'Box':
-                self.sup[self.nx/2-self.sup_radius:self.nx/2+self.sup_radius,
-                         self.ny/2-self.sup_radius:self.ny/2+self.sup_radius,
-                         self.nz/2-self.sup_radius:self.nz/2+self.sup_radius] = 1
-        if self.sup_shape == 'Disk':
-            dummy = _dist(np.shape(self.diff_array))
-            self.sup_index = np.where(dummy <= self.sup_radius)
-            self.sup[self.sup_index] = 1
+    Parameters
+    ----------
+    sup_radius : float
+        radius of support
+    shape_v : list
+        shape of diffraction pattern, which can be either 2D or 3D case.
 
-        self.sup_index = np.where(self.sup == 1)
-        self.sup_out_index = np.where(self.sup == 0)
+    Returns
+    -------
+    sup : array
+        support with a disk area
+    """
+    sup = np.zeros(shape_v)
+    dummy = _dist(shape_v)
+    sup_index = np.where(dummy <= sup_radius)
+    sup[sup_index] = 1
+    return sup
 
-    def recon(self, n_iterations=1000):
-        """
-        Run reconstruction with difference map algorithm.
 
-        Parameters
-        ---------
-        n_iterations : int
-            number of reconstructions to run
+def cdi_recon(diff_array, sample_obj, sup, **kwargs):
+    """
+    Run reconstruction with difference map algorithm.
 
-        Returns
-        -------
-        array :
-            reconstructed sample object
-        """
-        gamma_1 = -1/self.beta
-        gamma_2 = 1/self.beta
+    Parameters
+    ---------
+    diff_array : array
+        diffraction pattern from experiments
+    sample_obj : array
+        initial sample with phase
+    sup : array
+        initial support
+    kwargs : dict
+        parameters related to cdi reconstruction
 
-        # initiate shape and phase
-        if(self.init_obj_flag):
-            self.init_obj()
-        if(self.init_sup_flag):
-            self.init_sup()
+    Returns
+    -------
+    obj_ave : array
+        reconstructed sample object
+    error_dict : dict
+        error information for all iterations
+    """
 
-        self.obj_error = np.zeros(n_iterations)
-        self.diff_error = np.zeros(n_iterations)
-        self.sup_error = np.zeros(n_iterations)
+    diff_array = np.array(diff_array)     # diffraction data
 
-        sup_old = np.zeros_like(self.diff_array)
-        obj_ave = np.zeros_like(self.diff_array).astype(complex)
-        ave_i = 0
+    n_iterations = kwargs['n_iterations']
 
-        self.time_start = time.time()
-        for n in range(n_iterations):
-            self.obj_old = np.array(self.obj)
+    pi_modulus_flag = kwargs['pi_modulus_flag']  # 'Complex'
 
-            obj_a = pi_modulus(self.obj, self.diff_array)
-            if self.pi_modulus_flag == 'real':
-                obj_a = np.abs(obj_a)
-            obj_a = (1 + gamma_2) * obj_a - gamma_2 * self.obj
-            obj_a = pi_support(obj_a, self.sup_out_index)
+    beta = kwargs['beta']  # feedback parameter for difference map algorithm, around 1.15
+    start_ave = kwargs['start_ave']  # 0.8
 
-            obj_b = pi_support(self.obj, self.sup_out_index)
-            obj_b = (1 + gamma_1) * obj_b - gamma_1 * self.obj
-            obj_b = pi_modulus(obj_b, self.diff_array)
-            if self.pi_modulus_flag == 'real':
-                obj_b = np.abs(obj_b)
+    # parameters for shrink wrap
+    sw_flag = kwargs['sw_flag']
+    sw_sigma = kwargs['sw_sigma']  # 0.5
+    sw_threshold = kwargs['sw_threshold']  # 0.1
+    sw_start = kwargs['sw_start']  # 0.2
+    sw_end = kwargs['sw_end']  # 0.8
+    sw_step = kwargs['sw_step']  # 10
 
-            self.obj += self.beta * (obj_a - obj_b)
+    gamma_1 = -1/beta
+    gamma_2 = 1/beta
 
-            # calculate errors
-            self.obj_error[n] = cal_relative_error(self.obj_old, self.obj)
-            self.diff_error[n] = cal_diff_error(self.obj, self.diff_array)
+    #sup_index == np.where(sup == 1)
+    sup_out_index = np.where(sup != 1)
 
-            if self.shrink_wrap_flag is True:
-                if((n >= (self.sw_start * n_iterations)) and (n <= (self.sw_end * n_iterations))):
-                    if np.mod(n, self.sw_step) == 0:
-                        logger.info('Refine support with shrinkwrap')
-                        sup, self.sup_index, self.sup_out_index = find_support(self.obj,
-                                                                               self.sw_sigma,
-                                                                               self.sw_threshold)
-                        self.sup_error[n] = np.sum(sup_old) #cal_relative_error(sup_old, sup)
-                        sup_old = np.array(sup)
+    error_dict = {}
+    obj_error = np.zeros(n_iterations)
+    diff_error = np.zeros(n_iterations)
+    sup_error = np.zeros(n_iterations)
 
-            if n > self.start_ave*n_iterations:
-                obj_ave += self.obj
-                ave_i += 1
+    sup_old = np.zeros_like(diff_array)
+    obj_ave = np.zeros_like(diff_array).astype(complex)
+    ave_i = 0
 
-            logger.info('{} object_chi= {}, diff_chi={}'.format(n, self.obj_error[n],
-                                                                self.diff_error[n]))
+    time_start = time.time()
+    for n in range(n_iterations):
+        obj_old = np.array(sample_obj)
 
-        obj_ave = obj_ave / ave_i
-        self.time_end = time.time()
+        obj_a = pi_modulus(sample_obj, diff_array)
+        if pi_modulus_flag == 'real':
+            obj_a = np.abs(obj_a)
 
-        logger.info('object size: {}'.format(np.shape(self.diff_array)))
-        logger.info('{} iterations takes {} sec'.format(n_iterations, self.time_end - self.time_start))
+        obj_a = (1 + gamma_2) * obj_a - gamma_2 * sample_obj
+        obj_a = pi_support(obj_a, sup_out_index)
 
-        return obj_ave
+        obj_b = pi_support(sample_obj, sup_out_index)
+        obj_b = (1 + gamma_1) * obj_b - gamma_1 * sample_obj
+
+        obj_b = pi_modulus(obj_b, diff_array)
+        if pi_modulus_flag == 'real':
+            obj_b = np.abs(obj_b)
+
+        sample_obj += beta * (obj_a - obj_b)
+
+        # calculate errors
+        obj_error[n] = cal_relative_error(obj_old, sample_obj)
+        diff_error[n] = cal_diff_error(sample_obj, diff_array)
+
+        if sw_flag is True:
+            if((n >= (sw_start * n_iterations)) and (n <= (sw_end * n_iterations))):
+                if np.mod(n, sw_step) == 0:
+                    logger.info('Refine support with shrinkwrap')
+                    sup, sup_index, sup_out_index = find_support(sample_obj,
+                                                                 sw_sigma,
+                                                                 sw_threshold)
+                    sup_error[n] = np.sum(sup_old) #cal_relative_error(sup_old, sup)
+                    sup_old = np.array(sup)
+
+        if n > start_ave*n_iterations:
+            obj_ave += sample_obj
+            ave_i += 1
+
+        logger.info('{} object_chi= {}, diff_chi={}'.format(n, obj_error[n],
+                                                            diff_error[n]))
+
+    obj_ave = obj_ave / ave_i
+    time_end = time.time()
+
+    logger.info('object size: {}'.format(np.shape(diff_array)))
+    logger.info('{} iterations takes {} sec'.format(n_iterations, time_end - time_start))
+
+    error_dict['obj_error'] = obj_error
+    error_dict['diff_error'] = diff_error
+    error_dict['sup_error'] = sup_error
+
+    return obj_ave, error_dict
--- a/skxray/cdi.py
+++ b/skxray/cdi.py
@@ -44,6 +44,7 @@ from __future__ import (absolute_import, division, print_function,
 import six
 import numpy as np
 import time
+from scipy import signal
 
 import logging
 logger = logging.getLogger(__name__)
@@ -123,11 +124,12 @@ def convolution(array1, array2):
     fft_norm = lambda x:  np.fft.fftshift(np.fft.fftn(x)) / np.sqrt(np.size(x))
     fft_1 = fft_norm(array1)
     fft_2 = fft_norm(array2)
-    return np.abs(np.fft.ifftshift(np.fft.ifftn(fft_1*fft_2)) *
-                  np.sqrt(np.size(array2)))
+    #return np.abs(np.fft.ifftshift(np.fft.ifftn(fft_1*fft_2)) *
+    #              np.sqrt(np.size(array2)))
+    return signal.fftconvolve(array1, array2, mode='same')
 
 
-def pi_modulus(array_in, diffraction_array, pi_modulus_flag):
+def pi_modulus(array_in, diff_array, pi_modulus_flag):
     """
     Transfer sample from real space to q space.
     Use constraint based on diffraction pattern from experiments.
@@ -136,7 +138,7 @@ def pi_modulus(array_in, diffraction_array, pi_modulus_flag):
     ----------
     array_in : array
         reconstructed pattern in real space
-    diffraction_array : array
+    diff_array : array
         experimental data
     pi_modulus_flag : str
         Complex or Real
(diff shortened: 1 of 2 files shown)


licode:
This should be ready to review.

skxray/cdi.py
```diff
@@ -96,6 +98,27 @@ def gauss(dims, sigma):
     return y / np.sum(y)
 
 
+def _fft_helper(x, option='fft'):
```
tacaswell: I would suggest two functions `_fft_helper` and `_ifft_helper`.  Switching on a string like this is not a great pattern.  If you really want to be able to switch based on a string key, I would implement this as a dictionary look up of lambda functions.  This will fail in really surprising ways if option is not 'fft' or 'ifft'.

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
+    Transfer sample from real space to q space.
+    Use constraint based on diffraction pattern from experiments.
+
+    Parameters
+    ----------
+    array_in : array
+        reconstructed pattern in real space
+    diff_array : array
+        experimental data
+    pi_modulus_flag : str
```
tacaswell: enums should be documented as `pi_modulus_flag : {'Complex', 'Real'}`.  Should also use all lower case or all upper case (and normalize input just in case).

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
+    Returns
+    -------
+    array :
+        updated pattern in q space
+    """
+    diff_tmp = _fft_helper(array_in) / np.sqrt(np.size(array_in))
+    index = np.where(diff_array > 0)
+    diff_tmp[index] = diff_array[index] * diff_tmp[index] / (np.abs(diff_tmp[index]) + thresh_v)
+
+    if pi_modulus_flag == 'Complex':
```
tacaswell: This should be done with a dict look up.

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
-    fft_1 = fft_norm(array1)
-    fft_2 = fft_norm(array2)
-    return np.abs(np.fft.ifftshift(np.fft.ifftn(fft_1*fft_2)) *
-                  np.sqrt(np.size(array2)))
+    fft_1 = _fft_helper(array1) / np.sqrt(np.size(array1))
+    fft_2 = _fft_helper(array2) / np.sqrt(np.size(array2))
+    return np.abs(_fft_helper(fft_1*fft_2, option='ifft')) * np.sqrt(np.size(array2))
+
+
+def pi_modulus(array_in, diff_array,
```
tacaswell: Again, I don't see why this is not being done as two functions.

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
+    Parameters
+    ----------
+    sample_obj : array
+        sample data
+    index_v : array
+        index to define sample area
+
+    Returns
+    -------
+    array :
```
tacaswell: This needs a name as well as a type to make the wrapping to VT happy.

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
+    Returns
+    -------
+    float :
+        relative error in q space
+    """
+    new_diff = np.abs(_fft_helper(sample_obj)) / np.sqrt(np.size(sample_obj))
+    return cal_relative_error(diff_array, new_diff)
+
+
+class CDI(object):
```
tacaswell: Why is this a class?

skxray/cdi.py
licode: That is actually my original thought. However, the two helper functions are so similar, so I summarized them as one. If you strongly suggest that way, sure.

skxray/cdi.py
licode: got it.

skxray/cdi.py
licode: So you suggest to write two functions for real and complex respectively? By doing that, we can know the return type for sure?

skxray/tests/test_cdi.py
```diff
@@ -71,3 +73,106 @@ def test_convolution():
...
+
+def test_pi_support():
+    a1 = np.ones([2, 2])
+    a1[0, 0] = 1
+    index = np.where(a1 == 1)
+    a2 = pi_support(a1, index)
+    assert_equal(np.sum(a2), 0)
+
+
+def cal_diff():
```
tacaswell: Can you call this something like `make_synthetic_data` to make it clear what is going on?  This does something very different than `cal_diff_error`

skxray/cdi.py
tacaswell: As a general rule, drastically changing the code-path of a function based on a string switch is an anti-pattern.

skxray/cdi.py
tacaswell: If these are for internal use only I would just do

```python
_fft_normed = lambda x: np.fft.ifftshift(np.fft.fftn(np.fft.fftshift(x)))
_ifft_normed = lambda x: np.fft.ifftshift(np.fft.ifftn(np.fft.fftshift(x)))
```

Wouldn't it be simpler to just leave the fft in the natural order and only shift it when needed?

tacaswell: also, why is `fftshift` being called on real space data on its way in/out?

tacaswell: I also would not call this 'normed' with the amplitude being scaled by something.

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
+        pha_tmp = np.random.uniform(0, 2*np.pi, (self.nx, self.ny))
+        self.obj = (_fft_helper(self.diff_array * np.exp(1j*pha_tmp), option='ifft')
+                    * np.sqrt(np.size(self.diff_array)))
+
+    def init_sup(self):
+        """ Initiate sample support.
+        """
+        self.sup = np.zeros_like(self.diff_array)
+        if self.ndim == 2:
+            if self.sup_shape == 'Box':
```
tacaswell: These should each be a seperate functions `generate_2DBox_support` `generate_3Dbox_support`... which return numpy arrays which are in turn passed to the `recon` function.

licode: OK, I am thinking to break this class into small pieces. In the original code from beamline scientists, all the functions inside this file are included in a single class, causing me lots of headache...

licode: And I went through a tough process to restructure everything.

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
+                         self.nz/2-self.sup_radius:self.nz/2+self.sup_radius] = 1
+        if self.sup_shape == 'Disk':
+            dummy = _dist(np.shape(self.diff_array))
+            self.sup_index = np.where(dummy <= self.sup_radius)
+            self.sup[self.sup_index] = 1
+
+        self.sup_index = np.where(self.sup == 1)
+        self.sup_out_index = np.where(self.sup == 0)
+
+    def recon(self, n_iterations=1000):
```
tacaswell: This should be the top-level function.

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
+
+            obj_a = pi_modulus(self.obj, self.diff_array, self.pi_modulus_flag)
+            obj_a = (1 + gamma_2) * obj_a - gamma_2 * self.obj
+            obj_a = pi_support(obj_a, self.sup_out_index)
+
+            obj_b = pi_support(self.obj, self.sup_out_index)
+            obj_b = (1 + gamma_1) * obj_b - gamma_1 * self.obj
+            obj_b = pi_modulus(obj_b, self.diff_array, self.pi_modulus_flag)
+
+            self.obj = self.obj + self.beta * (obj_a - obj_b)
```
tacaswell: if you are going to update in place use `arr += beta * (arr_a - arr_b)`

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
+
+        sup_old = np.zeros_like(self.diff_array)
+        obj_ave = np.zeros_like(self.diff_array).astype(complex)
+        ave_i = 0
+
+        self.time_start = time.time()
+        for n in range(n_iterations):
+            self.obj_old = np.array(self.obj)
+
+            obj_a = pi_modulus(self.obj, self.diff_array, self.pi_modulus_flag)
```
tacaswell: `obj` is usally the abbreviation for 'object' when writing generic code that can take in _any_ python object, it is really confusing to use it here.

skxray/cdi.py
licode: sure, like the dpc code.

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
+
+        # parameters for shrink wrap
+        self.shrink_wrap_flag = kwargs['shrink_wrap_flag']
+        self.sw_sigma = kwargs['sw_sigma']  # 0.5
+        self.sw_threshold = kwargs['sw_threshold']  # 0.1
+        self.sw_start = kwargs['sw_start']  # 0.2
+        self.sw_end = kwargs['sw_end']  # 0.8
+        self.sw_step = kwargs['sw_step']  # 10
+
+    def init_obj(self):
```
tacaswell: This should be a top-level function `generate_random_phase_field(diff_array , shape)`

skxray/cdi.py
```diff
@@ -119,9 +142,284 @@ def convolution(array1, array2):
...
+    -------
+    float :
+        relative error in q space
+    """
+    new_diff = np.abs(_fft_helper(sample_obj)) / np.sqrt(np.size(sample_obj))
+    return cal_relative_error(diff_array, new_diff)
+
+
+class CDI(object):
+
```
tacaswell: Doing this as a class makes integrating this with VT very difficult, everything in the library should be functional unless there is a _very_ good reason.

skxray/cdi.py
licode: I would just call it _fft_helper or _ifft_helper

skxray/cdi.py
tacaswell: Yes, exactly like DPC

licode:
Any more comments? I think this is ready to go.

skxray/cdi.py
```diff
@@ -119,9 +125,316 @@ def convolution(array1, array2):
...
+        support with a disk area
+    """
+    sup = np.zeros(shape_v)
+    dummy = _dist(shape_v)
+    sup_index = np.where(dummy <= sup_radius)
+    sup[sup_index] = 1
+    return sup
+
+
+def cdi_recon(diff_array, sample_obj, sup, **kwargs):
```
tacaswell: Put all of the required args in the signature + document them

skxray/cdi.py
```diff
@@ -119,9 +125,316 @@ def convolution(array1, array2):
...
+    sup : array
+        initial support
+    kwargs : dict
+        parameters related to cdi reconstruction
+
+    Returns
+    -------
+    obj_ave : array
+        reconstructed sample object
+    error_dict : dict
```
tacaswell: What keys are in this dictionary and what do they mean?

licode:
Done with more corrections.

licode:
More comments, @tacaswell ?

skxray/cdi.py
```diff
@@ -109,7 +115,7 @@ def convolution(array1, array2):
 
     Returns
     -------
-    array :
+    arr : array
```
tacaswell: This function needs to be renamed to something like 'approximate_convolution' and have a `.. warning` which says something to the effect of "This function is incorrect but runs faster; use at your own risk".

It also needs a warning that it only works in the case where the two inputs are the same length (which is very surprising).

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+               thresh_v=1e-12):
+    """
+    Transfer sample from real space to q space.
+    Use constraint based on diffraction pattern from experiments.
+
+    Parameters
+    ----------
+    array_in : array
+        reconstructed pattern in real space
+    diff_array : array
```
tacaswell: `diff_array` is a very confusing name as it can be read as 'difference array' not 'diffraction array'

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+
+def pi_modulus(array_in, diff_array,
+               thresh_v=1e-12):
+    """
+    Transfer sample from real space to q space.
+    Use constraint based on diffraction pattern from experiments.
+
+    Parameters
+    ----------
+    array_in : array
```
tacaswell: This probably also needs a better name like 'proposed_phase'.  It also make more sense to me for the measured data (`diff_array` or what ever it gets renamed to) to be the first input

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+    thresh_v : float
+        add small value to avoid the case of dividing something by zero
+
+    Returns
+    -------
+    arr : array
+        updated pattern in q space
+    """
+    diff_tmp = _fft_helper(array_in) / np.sqrt(np.size(array_in))
+    index = np.where(diff_array > 0)
```
tacaswell: You can do just `index = diff_array > 0`

Exactly which will be faster will depend on the typical size of these arrays.  from rough testing, dropping the `where` saves 10x, but makes the indexing later about 2x slower.

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+    Transfer sample from real space to q space.
+    Use constraint based on diffraction pattern from experiments.
+
+    Parameters
+    ----------
+    array_in : array
+        reconstructed pattern in real space
+    diff_array : array
+        experimental data
+    thresh_v : float
```
tacaswell: This is not used as a threshold, but as an offset.

tacaswell:
Do we really need `_dist` here or can we re-using something we already have in core?

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+    new_sup : array
+        updated sample support
+    s_index : array
+        index for sample area
+    s_out_index : array
+        index not for sample area
+    """
+
+    sample_obj = np.abs(sample_obj)
+    gauss_fun = gauss(sample_obj.shape, sw_sigma)
```
tacaswell: The support of the gaussian should be set by the sigma.  You only need support out to a couple of standard deviations.  Using a smaller kernel will speed up the computation.

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+
+    conv_max = np.max(conv_fun)
+
+    s_index = np.where(conv_fun >= (sw_threshold*conv_max))
+    s_out_index = np.where(conv_fun < (sw_threshold*conv_max))
+
+    new_sup = np.zeros_like(sample_obj)
+    new_sup[s_index] = 1
+
+    return new_sup, s_index, s_out_index
```
tacaswell: These are all redundent with each other.  The one place this is used is the code the second two return values are discarded.  

I suggest dropping the second and third return values and just doing `return conv_fun >= (sw_threshold * conv_max)`

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+
+
+def pi_support(sample_obj, index_v):
```
tacaswell: Why is this called `pi_support`?

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+        radius of support
+    shape_v : list
+        shape of diffraction pattern, which can be either 2D or 3D case.
+
+    Returns
+    -------
+    sup : array
+        support with a box area
+    """
+    sup = np.zeros(shape_v)
```
tacaswell: This can be done much more simply:

```python
slc_list = [slice(s//2 - sup_radius, s//2 + sup_radius) for s in shape_v]
sup = np.zeros(shapev)
sup[slc_lst] = 1
```

It is also better to be explicit about dealing with the case of one of the dimensions being odd rather than relying on numpy to do the right thing with slicing with floats (it seems to cast the values to ints so this is the same behavior).

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+        shape of diffraction pattern, which can be either 2D or 3D case.
+
+    Returns
+    -------
+    sup : array
+        support with a disk area
+    """
+    sup = np.zeros(shape_v)
+    dummy = _dist(shape_v)
+    sup_index = np.where(dummy <= sup_radius)
```
tacaswell: `sup[dummy < sup_radius] = 1` 

The pattern in python leans towards open rather than closed intervals.

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+    sw_step : float, optional
+        the frequency to perform sw algorithm.
+        defualt is 10
+    n_iterations : int, optional
+        number of iterations to run.
+        default is 1000.
+
+    Returns
+    -------
+    obj_ave : array
```
tacaswell: what does 'ave' stand for?

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+        reconstructed sample object
+    error_dict : dict
+        Error information for all iterations. The dict keys include
```
tacaswell: If these are important enough to return, they probably should be their own return values, not buried in a dictionary.

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+        support with a disk area
+    """
+    sup = np.zeros(shape_v)
+    dummy = _dist(shape_v)
+    sup_index = np.where(dummy <= sup_radius)
+    sup[sup_index] = 1
+    return sup
+
+
+def cdi_recon(diff_array, sample_obj, sup,
```
tacaswell: This all assumes that q=0 is exactly at the center of the input image.

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+    error_dict : dict
+        Error information for all iterations. The dict keys include
+        obj_error, diff_error and sup_error. Obj_error is a list of
+        the relative error of sample object. Diff_error is calculated as
+        the difference between new diffraction pattern and the original
+        diffraction pattern. And sup_error stores the size of the
+        sample support.
+    """
+
+    diff_array = np.array(diff_array)     # diffraction data
```
tacaswell: I think it would make more sense here fftshift the input array and then work with that rather than doing two fftshifts _every_ time

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+    time_start = time.time()
+    for n in range(n_iterations):
+        obj_old = np.array(sample_obj)
+
+        obj_a = pi_modulus(sample_obj, diff_array)
+        if pi_modulus_flag.lower() == 'real':
+            obj_a = np.abs(obj_a)
+
+        obj_a = (1 + gamma_2) * obj_a - gamma_2 * sample_obj
+        obj_a = pi_support(obj_a, sup_out_index)
```
tacaswell: You can do this as `obj_a = obj_a * sup`

skxray/cdi.py
```diff
@@ -119,9 +125,339 @@ def convolution(array1, array2):
...
+        if pi_modulus_flag.lower() == 'real':
+            obj_b = np.abs(obj_b)
+
+        sample_obj += beta * (obj_a - obj_b)
+
+        # calculate errors
+        obj_error[n] = cal_relative_error(obj_old, sample_obj)
+        diff_error[n] = cal_diff_error(sample_obj, diff_array)
+
+        if sw_flag is True:
```
tacaswell: should be just `if sw_flag:`

skxray/cdi.py
danielballan: I am pretty sure we should remove convolution in favor of http://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.ndimage.filters.gaussian_filter.html'

As we discussed offline, the `truncate` argument controls how many sigma the Gaussian kernel is computed out to, tuning performance vs. accuracy. The `mode` argument might also be relevant. If I understand correctly, the standard in your application is to use `reflect`. In my naive opinion, using `constant` with `cval=0` might be more accurate.

licode: Thanks. will check that.

tacaswell: :+1: for using constant with `cval=0`

skxray/cdi.py
licode: I see, probably np.where double calculated it?

skxray/cdi.py
licode: good point.

skxray/cdi.py
licode: Indexing is faster than multiply operation. In cdi case, indexing is around 5 times faster, and even better when it comes to 3D case.

skxray/cdi.py
licode: pi_support is just to redraw the sample shape each time. Both pi_support and pi_modulus are conventional names borrowed from CDI world, and they also represent two basic operations in sample space and reciprocal space respectively.

skxray/cdi.py
tacaswell: fair enough.

skxray/cdi.py
licode: This is pretty neat.

skxray/cdi.py
licode: We always assume the center is q=0. Finding center is another step beyond scope of the current code.

licode:
@tacaswell , any more comments on the update?

tacaswell:
Back on my list of things to re-read.

skxray/cdi.py
licode: I think you might get a bit confused about the algorithm. diffraction array has nothing to do with fft or ifft. 

fft/ifft only works on sample_obj within the iteration part below. I didn't see the reason why it is better to take one fftshift out of the helper function, while keep the second fftshift inside the function. This will make the code difficult to read.

tacaswell: I am guessing that there will be a major performance gain due to reduced memory allocation.  The reason that you need the two fftshifts in the fft/ifft_helpers is because the diffraction image is [-w, 0, w] where as the order that fft yieds/expects is [0, 1, ... w, -w, -(w-1), -1].  If you re-order the diffraction image on the way in all of your internal functions can assume the 'hardware' ordering.

ex

```python
In [30]:  n = np.random.rand(200, 200)

In [32]: %timeit np.fft.fft(n)
1000 loops, best of 3: 493 µs per loop

In [33]: %timeit np.fft.fftshift(np.fft.fft(np.fft.fftshift(n)))
1000 loops, best of 3: 1.11 ms per loop

In [34]: n = np.random.rand(2000, 2000)

In [35]: %timeit np.fft.fft(n)
10 loops, best of 3: 72.1 ms per loop

In [36]: %timeit np.fft.fftshift(np.fft.fft(np.fft.fftshift(n)))
10 loops, best of 3: 109 ms per loop

```

skxray/cdi.py
```diff
@@ -96,32 +99,321 @@ def gauss(dims, sigma):
...
+    Transfer sample from real space to q space.
+    Use constraint based on diffraction pattern from experiments.
+
+    Parameters
+    ----------
+    recon_pattern : array
+        reconstructed pattern in real space
+    diffracted_pattern : array
+        diffraction pattern from experiments
+    offset_v : float
```
tacaswell: missing `, optional`

skxray/cdi.py
```diff
@@ -96,32 +99,321 @@ def gauss(dims, sigma):
...
+    Returns
+    -------
+    arr : array
+        updated pattern in q space
+    """
+    diff_tmp = _fft_helper(recon_pattern) / np.sqrt(np.size(recon_pattern))
+    index = diffracted_pattern > 0
+    diff_tmp[index] = (diffracted_pattern[index] *
+                       diff_tmp[index] / (np.abs(diff_tmp[index]) + offset_v))
+    return _ifft_helper(diff_tmp) * np.sqrt(np.size(diffracted_pattern))
```
tacaswell: I think you can do both of the multiplications here which will give a bit of a speed up (and make sure to use `()` so that the two scalar multiplications happen first).

skxray/cdi.py
```diff
@@ -96,32 +99,321 @@ def gauss(dims, sigma):
...
+    """
+
+    sample_obj = np.abs(sample_obj)
+    conv_fun = gaussian_filter(sample_obj, sw_sigma)
+
+    conv_max = np.max(conv_fun)
+
+    s_index = conv_fun >= (sw_threshold*conv_max)
+
+    new_sup = np.zeros_like(sample_obj)
```
tacaswell: these two lines are not needed, I think `s_index == new_sup`

skxray/cdi.py
```diff
@@ -96,32 +99,321 @@ def gauss(dims, sigma):
...
+    -------
+    obj_avg : array
+        reconstructed sample object
+    error_dict : dict
+        Error information for all iterations. The dict keys include
+        obj_error, diff_error and sup_error. Obj_error is a list of
+        the relative error of sample object. Diff_error is calculated as
+        the difference between new diffraction pattern and the original
+        diffraction pattern. And sup_error stores the size of the
+        sample support.
```
tacaswell: should this have a paper citation?

skxray/cdi.py
```diff
@@ -96,32 +99,321 @@ def gauss(dims, sigma):
...
+    """
+
+    diffracted_pattern = np.array(diffracted_pattern)     # diffraction data
+
+    gamma_1 = -1/beta
+    gamma_2 = 1/beta
+
+    # get support index
+    sup_out_index = sup < 1
```
tacaswell: might be worth checking if binary negation of `sup` works and is faster.  Also, this seems to be very badly named now.

skxray/cdi.py
```diff
@@ -96,32 +99,321 @@ def gauss(dims, sigma):
...
+    sup_old = np.zeros_like(diffracted_pattern)
+    obj_avg = np.zeros_like(diffracted_pattern).astype(complex)
+    avg_i = 0
+
+    time_start = time.time()
+    for n in range(n_iterations):
+        obj_old = np.array(sample_obj)
+
+        obj_a = pi_modulus(sample_obj, diffracted_pattern)
+        if pi_modulus_flag.lower() == 'real':
```
tacaswell: probably worth doing this once and stashing the result above.  String operations can be expensive.

skxray/cdi.py
```diff
@@ -96,32 +99,321 @@ def gauss(dims, sigma):
...
+            obj_a = np.abs(obj_a)
+
+        obj_a = (1 + gamma_2) * obj_a - gamma_2 * sample_obj
+        obj_a = pi_support(obj_a, sup_out_index)
+
+        obj_b = pi_support(sample_obj, sup_out_index)
+        obj_b = (1 + gamma_1) * obj_b - gamma_1 * sample_obj
+
+        obj_b = pi_modulus(obj_b, diffracted_pattern)
+        if pi_modulus_flag.lower() == 'real':
```
tacaswell: ditto on minimizing string operations.

skxray/cdi.py
```diff
@@ -96,32 +99,321 @@ def gauss(dims, sigma):
...
+                    sup = find_support(sample_obj, sw_sigma, sw_threshold)
+                    sup_out_index = sup < 1
+                    sup_error[n] = np.sum(sup_old)
+                    sup_old = np.array(sup)
+
+        if n > start_avg*n_iterations:
+            obj_avg += sample_obj
+            avg_i += 1
+
+        logger.info('{} object_chi= {}, diff_chi={}'.format(n, obj_error[n],
```
tacaswell: use old-style formatting here as it lets the logger be lazy and only do the formatting if strictly needed.  Something like

```python
logger.info("%s my message %s", 'foo', 'bar')
```

Note that the formatting inputs are args passed to the logging function.

skxray/cdi.py
```diff
@@ -96,32 +99,321 @@ def gauss(dims, sigma):
...
+                                                            diff_error[n]))
+
+    obj_avg = obj_avg / avg_i
+    time_end = time.time()
+
+    logger.info('object size: {}'.format(np.shape(diffracted_pattern)))
```
tacaswell: ditto on logging comment.

skxray/tests/test_cdi.py
```diff
@@ -62,12 +66,120 @@ def test_gauss():
...
+    cenv = shape_v[0]/2
+    r = 20
+    a = np.zeros(shape_v)
+    a[cenv-r:cenv+r, cenv-r:cenv+r] = 1.0
+    sw_sigma = 0.5
+    sw_threshold = 0.01
+
+    new_sup = find_support(a, sw_sigma, sw_threshold)
+    # the area of new support becomes larger
+    assert(np.sum(new_sup) > np.sum(a))
```
tacaswell: Can you predict how much bigger?  Should this add a 1 pixel boundary?

licode:
Corrections are made except fft/ifft part. As fft_helper is shown in multiple places, it is better to be a bit cautious when we do the updates.

skxray/cdi.py
licode: not very clear.

tacaswell: `fft(x * a)  == a * fft(x)` so you can factor out some of the multiplication and do `return _ifft_helper(diff_tmp) * (np.sqrt(np.size(diffraction_pattern) / np.sqrt(np.size(recon_pattern)`

tacaswell:
I don't think you addressed why cdi has it's own version of _dist when I think there is a function in `skxray.core` which does the same thing.

skxray/cdi.py
licode: Disagree. I don't think line 127 to 131 is linear transform. Already tested.

licode:
http://nbviewer.ipython.org/gist/licode/5c56c0bedd8fbc7e2ac0

PR #280, opened 2015-05-12
BUG: changed scikit-image to scikit-xray

sameera2004:


sameera2004:
I fixed this in the scikit-xray. documentation introduction.rst file.

PR #282, opened 2015-05-27
FIX: Do not use operands, which is nditer-specific.

danielballan:


skxray/correlation.py
```diff
@@ -119,9 +119,11 @@ def multi_tau_auto_corr(num_levels, num_bufs, labels, images):
...
                          "multiple-taus (must be even)")
 
-    if labels.shape != images.operands[0].shape[1:]:
-        raise ValueError("Shape of the image stack should be equal to"
-                         " shape of the labels array")
+    if hasattr(images, 'frame_shape'):
+        # Give a user-friendly error if we can detect the shape from pims.
+        if labels.shape != images.frame_shape:
+            raise ValueError("Shape of the image stack should be equal to"
+                             " shape of the labels array")
```
tacaswell: <rabbit hole> this should have an `elif` looking for `shape` and then use a `tee` to check the shape of the first image </rabbit hole>

danielballan: Too cute for function-specific code. Maybe a good utility.

tacaswell:
@danielballan @sameera2004 Something is very broken here.

tacaswell:
@sameera2004 can you figure out why this broke the tests?

sameera2004:
@tacaswell sure I will take look

sameera2004:
@tacaswell I fixed the tests and made a PR to @danielballan branch

PR #283, opened 2015-06-01
XSVS1

sameera2004:
This will provide most of the static tests need to do before - X-ray speckle visibility spectroscopy(XSVS)


sameera2004:
http://nbviewer.ipython.org/github/sameera2004/my_ipy_notebook/blob/new_duke/Duke_data_spe_vis.ipynb

ipython-notebook for skxray.speckle_visibility.speckle_visibility module functions

sameera2004:
This PR is ready for review (Failing test due to python 3)

tacaswell:
Can you please fix the python3 issues?  They both are minor syntax/usage fixes.

sameera2004:
Fixed the python 3 issue

skxray/core.py
```diff
@@ -1180,3 +1180,47 @@ def multi_tau_lags(multitau_levels, multitau_channels):
...
+                                  pixel_size)
+
+    bin_edges, sums, counts = bin_1D(np.ravel(radial_val),
+                                          np.ravel(image), nx)
+    th_mask = counts > thershold
+    ring_average = sums[th_mask] / counts[th_mask]
+
+    bin_centers = bin_edges_to_centers(bin_edges)
+
+    return bin_centers, ring_average
```
tacaswell: In general these are not the same length. The length of `bin_centers` will always `nx`, the length of `ring_average` will be `sum(counts > threshold)`.

skxray/core.py
```diff
@@ -1180,3 +1180,47 @@ def multi_tau_lags(multitau_levels, multitau_channels):
...
+
+    Parameters
+    ----------
+    image : array
+        input image
+
+    calibrated_center : tuple
+        The center in pixels-units (row, col)
+
+    thershold : float, optional
```
tacaswell: This is compared against something that is always a integer.  This is setting the minimum number of pixels in the bins that will be included.  This doc string gives a very different impressoin.

skxray/core.py
```diff
@@ -1180,3 +1180,47 @@ def multi_tau_lags(multitau_levels, multitau_channels):
...
+    pixel_size : tuple, optional
+        The size of a pixel in real units. (height, width). (mm)
+
+    Returns
+    -------
+    bin_centers : array
+        bin centers from bin edges
+
+    ring_averages : array
+        circular integration of SAXS intensity
```
tacaswell: Don't need to include that this is SAXS in the doc string.

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+def max_counts(sample_dict, label_array):
+    """
+    This will determine the highest speckle counts occurred in the required
+    ROI's in required images.
+
+    Parameters
+    ----------
+    sample_dict : dict
```
tacaswell: why are images coming in as a dictionary?

sameera2004: In speckle visibility spectroscopy(XSVS) they take image sets.
e.g. : Same sample three different positions (of the capillary) they take image sets
first set - 50 images
second sets - 100 images
third sets - 75 images 

Therefore, I thought better to give as a dictionary

skxray/core.py
sameera2004: I will take it out SAXS in the docstring

skxray/speckle_visibility/speckle_visibility.py
tacaswell: A series of image series would probably be better if they have a natural order.

tacaswell:
Are you aware of http://docs.scipy.org/doc/scipy/reference/ndimage.html#module-scipy.ndimage.measurements ?  I think some of the functions which take in label arrays have already implemented some of the same things.

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,293 @@
...
+        Each ROI is represented by a distinct label (i.e., integer).
+
+    Returns
+    -------
+    max_counts : int
+        maximum speckle counts
+    """
+    max_cts = 0
+    for key, img_sets in iteritems(sample_dict):
+        for n, img in enumerate(img_sets.operands[0]):
```
tacaswell: I think http://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.measurements.maximum.html#scipy.ndimage.measurements.maximum implements this

```python
import scipy.ndimage as ndi
max_cts = 0
for img_set in images_sets:
    for image in img_set:
        frame_max = ndi.measurements.maximum(img, label_array)
        max_cts = max(max_cts, frame_max)
return max_cts
```

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,293 @@
...
+            int_dist = intensity_distribution(img, label_array)
+            for j in range(len(int_dist)):
+                counts = np.max(list(int_dist.values())[j])
+                if max_cts < counts:
+                    max_cts = counts
+    return max_cts
+
+
+def intensity_distribution(image_array, label_array):
```
tacaswell: The name and docstring make me think of http://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.measurements.histogram.html#scipy.ndimage.measurements.histogram but it looks like you are just pulling out the arrays of the pixels in each ROI?

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,293 @@
...
+    for n in label_num:
+        value = (np.ravel(image_array)[indices[labels == n].tolist()])
+        intensity_distribution[n] = value
+
+    return intensity_distribution
+
+
+def time_bining(number=2, number_of_images=50):
+    """
+    This will provide the time binning for the integration.
```
tacaswell: This docstring is too vauge.

tacaswell: This is a geometric series http://mathworld.wolfram.com/GeometricSeries.html so you can directly compute the number of bins you need an the bin sizes are just `r ** np.arange(k)`

For example

```python
import numpy as np

def num_time_bins(n, r):
    """
    Given a base r find the number of steps such that the
    geometric sum to k with base r < n
    """

    k = np.log((n * (r - 1) + 1) / r) / np.log(r)
    return np.floor(k) + 1



def test_time_bins():
    for r in np.random.randint(2, 7, 100):
        for n in np.random.randint(70, 1000, 100):
            k = num_time_bins(n, r)
            try:
                assert np.sum(r ** np.arange(k)) < n
                assert np.sum(r ** np.arange(k + 1)) > n
            except:
                print(r, n, k)
                raise
```

tacaswell: although, it looks like logs are really expensive and the existing version is faster....

sameera2004:
Fixed the issues, added tests, and ready for PR review

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,356 @@
...
+        Each ROI is represented by a distinct label (i.e., integer).
+
+    Returns
+    -------
+    max_counts : int
+        maximum speckle counts
+    """
+    max_cts = 0
+    for img_set in images_sets:
+        for n, img in enumerate(img_set.operands[0]):
```
tacaswell: what is `operands[0]`?  If `image_sets` is just an array, this iterator should be returning ndarrays.

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,356 @@
...
+            frame_max = ndi.measurements.maximum(img, label_array)
+            max_cts = max(max_cts, frame_max)
+    return max_cts
+
+
+def intensity_distribution(image_array, label_array):
```
tacaswell: Do we need this function?  It is not used anywhere else (except it's test)

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,356 @@
...
+    """
+
+    time_bin = [1]
+
+    while time_bin[-1]*number < number_of_images:
+        time_bin.append(time_bin[-1]*number)
+    return time_bin
+
+
+def static_test_sets_one_label(sample_dict, label_array, num=1):
```
tacaswell: Why do these functions have 'test' in the name?

sameera2004: looks like confusing, I will change the names? ( take out 'test' in the names)

skxray/speckle_visibility/speckle_visibility.py
sameera2004: Yes, before doing XSVS, they need to see the radial intensity distribution to make sure that they found the correct calibrated_center. 

Usually if they have 1000 images they take either average image of the images (or sum of those images) and see the intensity distribution. Other thing is this is one of Andrei's requirement he wants to get following plots for XSVS data.

screen shot 2015-06-17 at 1 16 28 pm

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,356 @@
...
+        raise ValueError("Shape of the image data should be equal to"
+                         " shape of the labeled array")
+
+    labels, indices = corr.extract_label_indices(label_array)
+    label_num = np.unique(labels)
+
+    intensity_distribution = {}
+
+    for n in label_num:
+        value = (np.ravel(image_array)[indices[labels == n].tolist()])
```
tacaswell: `value = image_array[labels == n]`

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,356 @@
...
+    -------
+    radial_intensity : dict
+        radial intensity of each ROI's
+    """
+
+    if label_array.shape != image_array.shape:
+        raise ValueError("Shape of the image data should be equal to"
+                         " shape of the labeled array")
+
+    labels, indices = corr.extract_label_indices(label_array)
```
tacaswell: don't need this line

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,356 @@
...
+    label_num = np.unique(labels)
+
+    intensity_distribution = {}
+
+    for n in label_num:
+        value = (np.ravel(image_array)[indices[labels == n].tolist()])
+        intensity_distribution[n] = value
+
+    return intensity_distribution
```
tacaswell: ` return {n: test[label == n] for n in range(np.max(label))}`

skxray/speckle_visibility/speckle_visibility.py
tacaswell: `roi_pixel_values`

skxray/speckle_visibility/speckle_visibility.py
tacaswell:
```python

for img_set in images_sets:
    for img in img_set:
        max_cts = max(max_cts, ndi.measurements.maximum(img, label_array))

return max_cts
```

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,356 @@
...
+
+
+def time_bining(number=2, number_of_images=50):
```
tacaswell: are these bin widths or bin edges?

sameera2004: bin edges

tacaswell: Then they should start at 0?  That should also be documented.

sameera2004:
Ready for PR review

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
...
+        Each ROI is represented by a distinct label (i.e., integer).
+
+    Returns
+    -------
+    mean_int_labels : dict
+        average intensity of each ROI as a dictionary
+        {roi 1: average intensities, roi 2 : average intensities}
+
+    """
+    return {n+1 : mean_intensity(images_set[n],
```
tacaswell: The docstring and the code don't match.  The output will be a dict with `len(images_set)` entries, not `len(np.unique(labels))`

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
...
+        iterable of 2D arrays
+        dimensions are: (rr, cc)
+
+    labels : array
+        labeled array; 0 is background.
+        Each ROI is represented by a distinct label (i.e., integer).
+
+    Returns
+    -------
+    mean_int_labels : dict
```
tacaswell: I don't understand why this is coming back as a dict.  Given what the code does, I expect a list or array back.

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
...
+
+    """
+    if labels.shape != images[0].shape[0:]:
+        raise ValueError("Shape of the images should be equal to"
+                         " shape of the label array")
+
+    index = np.unique(labels)[1: ]
+    mean_int = np.zeros((images.shape[0], index.shape[0]))
+
+    for n in range(images.shape[0]):
```
tacaswell: use `for n, img in enumerate(images):` so this will work with things that don't have `shape`, but can be iterated.

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
...
+    mean_int : array
+        mean intensity of each ROI for the set of images as an array
+        shape (number of images in the set, number of labels)
+
+    """
+    if labels.shape != images[0].shape[0:]:
+        raise ValueError("Shape of the images should be equal to"
+                         " shape of the label array")
+
+    index = np.unique(labels)[1: ]
```
tacaswell: I think this should be `index = np.arange(1, np.max(labels) + 1)`.  If you have labels with gaps, you should probably use the relabel method to get the ROIs labeled sequentially again.

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
...
+        mean intensity of each ROI as a dictionary
+        {roi 1: average intensities, roi 2 : average intensities}
+
+    Returns
+    -------
+    combine_mean_int : array
+        combine mean intensities of image sets for each ROI of labeled array
+
+    """
+    return np.vstack(list(mean_int_dict.values()))
```
tacaswell: This will come out in a random order.

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
...
+try:
+    iteritems = dict.iteritems
+except AttributeError:
+    iteritems = dict.items  # python 3
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+def max_counts(images_sets, label_array):
```
tacaswell: This should have ROI in the function name

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
+# ######################################################################
```
tacaswell: Can you rename move this file to the top level and rename it to not name-chcek a specific technique?  I have not seen anything domain specific in here

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def max_counts(images_sets, label_array):
+    """
+    This will determine the highest speckle counts occurred in the required
+    ROI's in required images.
+
+    Parameters
+    ----------
+    images_sets : array
```
tacaswell: So this is a 4-D array (an iterable of iterables which yield 2D images)?

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
...
+    #labels, indices = corr.extract_label_indices(label_array)
+    label_num = np.unique(labels)[1:]
+
+    #intensity_distribution = {}
+
+    #for n in label_num:
+    #    value = (np.ravel(image_array)[indices[labels == n].tolist()])
+    #    intensity_distribution[n] = value
+
+    return {n: image[labels == n] for n in range(1, np.max(labels))}
```
tacaswell: I think there is an off-by-one in here.

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def time_bin_edges(number=2, number_of_images=50):
```
tacaswell: If there are bin edges, then they should start at 0, not 1

skxray/speckle_visibility/speckle_visibility.py
```diff
@@ -0,0 +1,290 @@
...
+                                  pixel_size)
+
+    bin_edges, sums, counts = core.bin_1D(np.ravel(radial_val),
+                                          np.ravel(image), nx)
+    th_mask = counts > threshold
+    ring_averages = sums[th_mask] / counts[th_mask]
+
+    bin_centers = core.bin_edges_to_centers(bin_edges)
+
+    return bin_centers, ring_averages
```
tacaswell: in general `len(bin_centers) != len(ring_averages)`

skxray/speckle_visibility/speckle_visibility.py
sameera2004: The dict come back for separate image sets.
When they take different image sets( ex: 100 images, 200 images, 150 images, etc..) in different places.

skxray/speckle_visibility/speckle_visibility.py
sameera2004: May be the name is confusing, they want to integrate data at different times, first one should be 1 not 0

ericdill:
Is there an example that goes along with this? I remember seeing a 2x2 plot in one of your comments in here. Did you generate that?

sameera2004:
@ericdill No I got them from Andrei, those are the plots he needs for the data sets before doing XSVS

skxray/speckle_visibility/speckle_visibility.py
sameera2004: @tacaswell I am developing analysis softwares for XSVS(X-ray speckle visibility spectroscopy) and it will use some of the functions here and other functions before doing XSVS analysis. Therefore I thought it is better to have it separate XSVS analysis?

skxray/speckle_visibility/speckle_visibility.py
sameera2004: Sure I will add ROI in the max_counts function name

sameera2004:
@ericdill Here's the new python notebook
http://nbviewer.ipython.org/github/sameera2004/my_ipy_notebook/blob/new_ana/Duke_Data_pre_XSVS.ipynb

skxray/speckle_visibility/speckle_visibility.py
sameera2004: In here both index = np.unique(labels)[1: ]
and  index = np.arange(1, np.max(labels) + 1), will give the same answer

skxray/speckle_visibility/speckle_visibility.py
sameera2004: Sure thanks I will fix it ('off-by-one')

skxray/speckle_visibility/speckle_visibility.py
sameera2004: yes len(bin_centers) != len(ring_averages)

sameera2004:
Fixed the documentation and changed some of function names and fixed the bugs

Ready for review

skxray/speckle_visibility/speckle_visibility.py
tacaswell: That is very contrary to expectations.  If they are not the same length how do you till with averages go with which bin?

skxray/speckle_visibility/speckle_visibility.py
tacaswell: Right, but one of the goals of this whole project is to reuse as much code between techniques as possible.  Putting general use code in to domain specific files (two levels deep) works against that goal.

sameera2004:
added same mask to bin centers(which used to ring averages), fixed the documentation,
ready for review

skxray/speckle_visibility/speckle_visibility.py
sameera2004: @tacaswell Ok sure, I renamed and moved the file to top level, ready for PR review

sameera2004:
Moved the file to top level, updated the python notebook, ready for review

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,317 @@
...
+except AttributeError:
+    iteritems = dict.items  # python 3
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+def roi_max_counts(images_sets, label_array):
+    """
+    This will determine the highest speckle counts occurred in the required
```
tacaswell: The docstring should not mention speckles.  

'Return the brightest pixel in any ROI in any image in the image set' or something like that?

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,317 @@
...
+    ROI's in required images.
+
+    Parameters
+    ----------
+    images_sets : array
+        sets of images as an array
```
tacaswell: What is the expeceted dimensionality of this input?

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,317 @@
...
+
+    label_array : array
+        labeled array; 0 is background.
+        Each ROI is represented by a distinct label (i.e., integer).
+
+    Returns
+    -------
+    max_counts : int
+        maximum speckle counts
```
tacaswell: maximum pixel value

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,317 @@
...
+    """
+
+    if labels.shape != image.shape:
+        raise ValueError("Shape of the image data should be equal to"
+                         " shape of the labeled array")
+
+    return {n: image[labels == n] for n in range(1, np.max(labels)+1)}
+
+
+def time_series(number=2, number_of_images=50):
```
tacaswell: This should be 'geometric_bin_edges' or something like that.

It also should not have any default values for the args.

tacaswell: And this should probably be in `core.py`

sameera2004: ok sure, I will move to core, and take out default values and rename it

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,317 @@
...
+    """
+    This will provide the geometric series for the integration.
+    Last values of the series has to be less than or equal to number
+    of images
+    ex: number_of_images = 100
+    number = 2, time_series =  1, 2, 4, 8, 16, 32, 64
```
tacaswell: Is this meant to be used as 

```python
bin_edges = geometric_bin_edges(2, 50)
for f, b in zip(bin_edges[:-1], bin_edges[1:]):
    do_something(images[f:b])
```

```python
prev = 0
for width is bin_edges
    do_something(images[prev:prev + width])
    prev += width
```

?

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,317 @@
...
+    ------
+    time_series : list
+        time binning
+
+    Note
+    ----
+    :math ::
+     a + ar + ar^2 + ar^3 + ar^4 + ...
+
+     a - first term in the series
```
tacaswell: You document `a` here, but provide no way to change it in the code.

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,317 @@
...
+    mean_intensity : array
+        mean intensity of each ROI for the set of images as an array
+        shape (len(images), number of labels)
+
+    """
+    if labels.shape != images[0].shape[0:]:
+        raise ValueError("Shape of the images should be equal to"
+                         " shape of the label array")
+
+    index = np.unique(labels)[1:]
```
tacaswell: You do not want want to do this on the uniqueified labels, as this means you may not be able to lookup in the output the mean of a given ROI.

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,317 @@
...
+                                          np.ravel(image), nx)
+    th_mask = counts > threshold
+    ring_averages = sums[th_mask] / counts[th_mask]
+
+    bin_centers = core.bin_edges_to_centers(bin_edges)[th_mask]
+
+    return bin_centers, ring_averages
+
+
+def roi_speckle_count_waterfall(images, labels, num=1):
```
tacaswell: Drop speckle from the name.

tacaswell: and can we call this `roi_kymograph` https://en.wikipedia.org/wiki/Kymograph which is the proper name for these kinds of plots.

sameera2004: sure I will call it as roi_kymograph  and drop 'speckle'

skxray/speckle_analysis.py
sameera2004: These are integration times. e.g.: they take images in every 1ms for 100 images, 
In XSVS, they want to see the scattering patterns with longer integration times like 1ms,2ms, 4ms,8ms, 16ms, 32ms, 64ms (Because they have only 100 images they can't go 128ms) or (1ms, 10ms, 100ms like that), 
This function has to create that time series

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def mean_intensity_sets(images_set, labels):
+    """
+    Mean intensities for ROIS' of the labeled array for different image sets
+
+    Parameters
+    ----------
+    images : array
+        iterable of 4D arrays
```
tacaswell: so

```python
for img in images:
    assert img.ndim == 4
```

?

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,279 @@
...
+    -------
+    mean_int_labels : dict
+        average intensity of each ROI as a dictionary
+        shape len(images_sets)
+        eg: 2 image sets,
+        {image set 1 : (len(images in image set 1), number of labels),
+        image set 2 : (len(images in image set 2), number of labels)}
+
+    """
+    return {n+1: mean_intensity(images_set[n],
```
tacaswell: Why not

```python
{n+1: mean_intensity(img, labels) for n, img in enumerate(images_set)}
```

?

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,279 @@
...
+        image set 2 : (len(images in image set 2), number of labels)}
+
+    Returns
+    -------
+    combine_mean_int : array
+        combine mean intensities of image sets for each ROI of labeled array
+        shape (len(images in all image sets), number of labels)
+
+    """
+    return np.vstack(list(mean_int_dict.values()))
```
tacaswell: This will come out in a random order in python3.  I assume this is to deal with the output from `mean_intensity_sets`, why not just have that return a list to begin with?

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,279 @@
...
+        to the pixel list
+        {ROI 1 : intensities of the pixels of ROI 1, ROI 2 : intensities of
+         the pixels of ROI 2}
+    """
+
+    if labels.shape != image.shape:
+        raise ValueError("Shape of the image data should be equal to"
+                         " shape of the labeled array")
+
+    return {n: image[labels == n] for n in range(1, np.max(labels)+1)}
```
tacaswell: This should probably return a list of ndarrays

tacaswell: eh, never mind that

tacaswell: never mind the never mind.

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,279 @@
...
+        maximum pixel counts
+    """
+    max_cts = 0
+    for img_set in images_sets:
+        for img in img_set:
+            max_cts = max(max_cts, maximum(img, label_array))
+    return max_cts
+
+
+def roi_pixel_values(image, labels):
```
tacaswell: This should take an optional argument `index` and follow the convention of the scipy functions.  If it is `None` return all of the labels, if it is a list just work on those.

```python
def roi_pixel_values(image, labels, index=None):
    if index is None:
        index = np.arange(1, np.max(labels) + 1)

    out = [image[labels == n] for n in index]
    return out, index
```
or 
```
    out = {n: image[labels == n] for n in index}
    return out
```

Not clear which way is better

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,279 @@
...
+                                          np.ravel(image), nx)
+    th_mask = counts > threshold
+    ring_averages = sums[th_mask] / counts[th_mask]
+
+    bin_centers = core.bin_edges_to_centers(bin_edges)[th_mask]
+
+    return bin_centers, ring_averages
+
+
+def roi_kymograph(images, labels, num=1):
```
tacaswell: num should be required.

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,279 @@
...
+        required ROI
+
+    Returns
+    -------
+    roi_kymograph : array
+
+    """
+    roi_kymograph = []
+    for n, img in enumerate(images):
+        roi_kymograph.append(list(roi_pixel_values(img,
```
tacaswell: With the changes above this turns into

```python
roi_pixel = roi_pixel_list(img, labels, [num, ])[num]
```

```python
[roi_pixel,], index = roi_pixel_list(img, labels, [num, ])
```

skxray/core.py
```diff
@@ -1180,3 +1180,44 @@ def multi_tau_lags(multitau_levels, multitau_channels):
     lag_steps = np.append(lag_steps, np.array(lag))
 
     return tot_channels, lag_steps
+
+
+def geometric_bin_edges(common_ratio, number_of_images, first_term=1):
```
tacaswell: I am still not sure on the use case for this

sameera2004: geometric_bin_edges, use case is for XSVS analysis, I will put new PR soon for that

tacaswell: Can you sketch out how you would use the output of this function?

tacaswell: Can you open that next PR up with just the function signatures and docstrings as soon as possible?

sameera2004:
@tacaswell,  I fixed this PR, ready for review

sameera2004:
Changed the function name to geometric_series (geometric_bin_edges) ready for review

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,311 @@
...
+    return roi_pix, index
+
+
+def mean_intensity_sets(images_set, labels):
+    """
+    Mean intensities for ROIS' of the labeled array for different image sets
+
+    Parameters
+    ----------
+    images_set : array
```
tacaswell: Is this a ragged array which means that this is an object type array.  If each of these is a different length then the input should probably be a list (and the return should be a list of the same length).

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,311 @@
...
+        shapes is: (len(images_sets), )
+        one images_set is iterable of 2D arrays dimensions are: (rr, cc)
+
+    labels : array
+        labeled array; 0 is background.
+        Each ROI is represented by a distinct label (i.e., integer).
+
+    Returns
+    -------
+    mean_intensity_dict : dict
```
tacaswell: It seems odd to me to take in an array and return a dictionary.

tacaswell: It seems odd to me to take in an array and return a dictionary.

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,311 @@
...
+    return mean_intensity_dict, index_list
+
+
+def mean_intensity(images, labels, index=None):
+    """
+    Mean intensities for ROIS' of the labeled array for set of images
+
+    Parameters
+    ----------
+    images : array
```
tacaswell: The dtype and the description do not match.

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,311 @@
...
+        {image set 1 : len(images in image set 1),
+        image set 2 : len(images in image set 2)}
+
+    index_list : list
+        labels list for each image set
+
+    """
+    mean_intensity_dict = {}
+    index_list = []
+    for n in range(len(images_set)):
```
tacaswell: I think this can be re-written as

```python
return tuple(map(list, zip(*[mean_intensity(im, labels) for im in images_set])))
```

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,311 @@
...
+
+    Returns
+    -------
+    combine_mean_int : array
+        combine mean intensities of image sets for each ROI of labeled array
+        shape (len(images in all image sets), number of labels)
+
+    """
+    if np.all(map(lambda x: x == index_list[0], index_list)):
+        combine_mean_intensity = np.vstack(list(mean_int_dict.values()))
```
tacaswell: you do this here and save the result.

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+def combine_mean_intensity(mean_int_dict, index_list):
+    """
+    Combine mean intensities of the images(all images sets) for each ROI
+    if the labels list of all the images are same
+
+    Parameters
+    ----------
+    mean_int_dict : dict
```
tacaswell: Given that there is a natural order to these values, I think these make more sense as a list than as a dict.  The power of dictionaries is that they give you very fast random access to un-ordered data (or data that if you did order it would be very spare, think reserving space in an array for every valid combination of letters and numbers).  The way that dict is implemented is as a hash table.  The first step is to take your key and reduce it to an integer (hash) which is used to essentially index into an array.  The draw back of this is that the order of the keys is the hash is (nominally) random, thus, when you iterate over the dictionary you can not assume anything about the order.

In this case the application is for looking at the same data set integrated with different intervals which has a very definite ordering.

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,311 @@
...
+        combine mean intensities of image sets for each ROI of labeled array
+        shape (len(images in all image sets), number of labels)
+
+    """
+    if np.all(map(lambda x: x == index_list[0], index_list)):
+        combine_mean_intensity = np.vstack(list(mean_int_dict.values()))
+    else:
+        raise ValueError("Labels list for the image sets are different")
+
+    return np.vstack(list(mean_int_dict.values()))
```
tacaswell: and then do it again to return ?

The order of this will be arbitrary.  In python 2 it will be the same from run-to-run, but in python 3.4 this order will be random between python processes.

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+def circular_average(image, calibrated_center, threshold=0, nx=100,
```
tacaswell: Don't we already have at least on implemenation of this?  I was doing radial integration when I was working on the tilt calibration.

sameera2004: no we don't have a function for this

skxray/speckle_analysis.py
sameera2004: Changed it to a list

skxray/speckle_analysis.py
sameera2004: Yes,  Thanks

skxray/speckle_analysis.py
sameera2004: Changed the description

sameera2004:
fixed the code, ready for review

sameera2004:
@tacaswell ready for review

skxray/speckle_analysis.py
```diff
@@ -0,0 +1,305 @@
...
+
+def mean_intensity(images, labels, index=None):
+    """
+    Mean intensities for ROIS' of the labeled array for set of images
+
+    Parameters
+    ----------
+    images : array
+        Intensity array of the images
+        dimensions are: [num_img][num_rows][num_cols]
```
tacaswell: Can you write this as would come back from shape `(num_img, num_rows, num_cols)`?

sameera2004:
@tacaswell I fixed the documentation,

skxray/core.py
```diff
@@ -1226,3 +1226,44 @@ def multi_tau_lags(multitau_levels, multitau_channels):
     lag_steps = np.append(lag_steps, np.array(lag))
 
     return tot_channels, lag_steps
+
+
+def geometric_series(common_ratio, number_of_images, first_term=1):
```
ericdill: @sameera2004 what is the purpose of this function? It does not appear anywhere else in this PR except for its test

sameera2004: @ericdill All these functions are going to use in XSVS2.
As an example in XSVS they collect 100 images, each take in1ms,
Then they choose ROI and see the photon count probability density  in that ROI for 1ms, 2ms, 4ms, 8s, 16ms, etc
or 1ms, 3ms, 9ms, 27ms,  etc like, 
This will create for 100 images, common ratio=2,  geometric_series =  1, 2, 4, 8, 16, 32, 64

ericdill: Moving this conversation to #299

skxray/speckle_analysis.py
ericdill: did you get an answer on this @tacaswell ?

ericdill:
see #299

skxray/speckle_analysis.py
sameera2004: Earlier it used to return dict, not anymore. It will take different image sets and return mean intensities as a list, both shape len(image sets)

PR #285, opened 2015-06-07
ENH/TST: optimize CDI code,

licode:
The radial_grid from core is for 2D case only, so I still keep the _dist function.

### DIFF
--- a/skxray/cdi.py
+++ b/skxray/cdi.py
@@ -323,9 +323,6 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
     obj_avg = np.zeros_like(diffracted_pattern).astype(complex)
     avg_i = 0
 
-    obj_a = np.zeros_like(obj_avg)
-    obj_b = np.zeros_like(obj_avg)
-
     time_start = time.time()
     for n in range(n_iterations):
         obj_old = np.array(sample_obj)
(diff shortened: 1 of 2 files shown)


skxray/cdi.py
```diff
@@ -364,6 +323,9 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
     obj_avg = np.zeros_like(diffracted_pattern).astype(complex)
     avg_i = 0
 
+    obj_a = np.zeros_like(obj_avg)
```
tacaswell: you replace this ref on lines 334, 335, 337

skxray/cdi.py
```diff
@@ -364,6 +323,9 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
...
+    obj_b = np.zeros_like(obj_avg)
+
     time_start = time.time()
     for n in range(n_iterations):
         obj_old = np.array(sample_obj)
```
tacaswell: do you mutate `sample_obj` in `py_modulus`?

licode: no, that's why I created a new one.

tacaswell: but if you don't mutate it, then you don't need to create a new copy.

licode: Sorry, I mean I mutated sample_obj in py_modulus.

tacaswell: That is a very crucial miss placed negative!

licode:
Other comments on this, @tacaswell ?

tacaswell:
Does it actually go faster?

licode:
Around 5% faster, but the code is much clean.

PR #287, opened 2015-06-15
ENH: add support for XPD reciprocal space translation 

CJ-Wright:
from pixel position to Q, using pyFAI's calibration/geometry object, also PEP8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
   - conda create -n testenv --yes pip nose python=$TRAVIS_PYTHON_VERSION lmfit xraylib numpy scipy scikit-image netcdf4 six
   - source activate testenv
   - python setup.py install
-  - pip install coveralls
+  - pip install coveralls, pyFAI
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@sameera2004 here is the pixel to Q function we were talking about today.

tacaswell:
Can you make sure we have a conda rescipe for pyFai and add it as a dependancy in all of the correct places?

CJ-Wright:
The conda recipe I am not sure about, never done that before.  But I will add pyFAI to the setup.py and travis.yml

CJ-Wright:
@tacaswell are only conda packages allowed in meta.yaml?

tacaswell:
Yes

licode:
pyFAI is a good library to borrow. :+1:

sameera2004:
@CJ-Wright sure thanks, yes it is good if we can use pyFAI for XPD data analysis

PR #289, opened 2015-06-21
Img proc math

giltis:
Minor docstring correction and organizational refactor. Finally changing module name from mathops to more clearly understandable name of arithmetic


tacaswell:
This needs a note in https://github.com/Nikea/scikit-xray/tree/master/doc/api_changes documenting the API change.

danielballan:
Probably have to update setup.py as well.

ericdill:
I would also advocate for turning `image.py` into a package and putting `arithmetic` in that package instead of `image_processing`

giltis:
@ericdill : So you're thinking of simply including a package called 'image' instead of 'image processing' ?

giltis:
@danielballan : There is a very real possibility that I'm missing something here, but I'm not seeing where, or why a change to setup.py should be made.

sameera2004:
But there is a module called 'image.py' already in skxray? It has some image processing tools too, (@tacaswell finding the center of the ring). May be, you can integrate everything into image package?

ericdill:
All of the modules in skxray/ (`feature.py`, `image.py`, `recip.py`, etc...) will eventually turn in to packages.  They are not packages right now because they are not sufficiently large to warrant their own packages. I'd suggest the following rearrangements to `image.py` and `image_processing/arithmetic.py`:

```
skxray/
   image/
      __init__.py (import find_ring_center_acorr_1D here, there will be no api break this way)
      ring.py (move the code from `image.py` here)
      arithmetic.py
```

ericdill:
re: setup.py. As the setup.py file of skxray is currently written, there is no need to change it (https://github.com/Nikea/scikit-xray/blob/master/setup.py#L21).  What @danielballan is likely referring to is how setup.py is done in many of our other libraries (see dataportal, for example https://github.com/NSLS-II/dataportal/blob/master/setup.py#L65) where we are explicitly listing each and every subpackage in the library.  Because you added a new subpackage `image_processing` you would need to update the setup.py file if we were listing all subpackages explicitly.

PR #291, opened 2015-06-24
Img proc api

giltis:
Per our conversations about where to place, and how to include the compilations of image processing tools for segmentation, analysis and quantification, this PR adds tools for image filtering, morphological operations, and registration to the API, and (via VTTools) makes them accessible for use in VisTrails for our users.


tacaswell:
I have no problems with this.  Should add a test file which just does an `import foo` for all of these.   If we want to be really cranky check that their name spaces match the expected all and will fail if we add or remove anything from them.  We don't need to do functionality checks as they are done else where (either by us or by their third-party sources).

ericdill:
I would argue that this sort of 'helpful namespacing' is actually harmful (re: pylab).  If we are going to reduce the barrier to new users, we should be providing awesome examples/demos in python scripts and ipython notebooks!  Closing for now.

PR #294, opened 2015-06-29
Cdi plot

licode:
Add plotting function to cdi reconstruction.

### DIFF
--- a/skxray/cdi.py
+++ b/skxray/cdi.py
@@ -284,9 +284,11 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
         number of iterations to run.
         default is 1000.
     plot_function : function, optional
-        plotting function
+        This is a callback function that expects to receive these
+        four objects: sample_obj, obj_error, diff_error, sup_error
     plot_step : int, optional
-        step interval for plotting
+        define plotting frequency, i.e., if plot_step = 10, plot results
+        after every 10 iterations.
 
     Returns
     -------
@@ -366,7 +368,7 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
                     sup_error[n] = np.sum(sup_old)
                     sup_old = np.array(sup)
 
-        if plot_function and np.mod(n_iterations, plot_step) == 0:
+        if plot_function and n_iterations % plot_step == 0:
             plot_function(sample_obj, obj_error, diff_error, sup_error)
 
         if n > start_avg*n_iterations:
--- a/skxray/cdi.py
+++ b/skxray/cdi.py
@@ -45,7 +45,6 @@ import six
 import numpy as np
 import time
 from scipy.ndimage.filters import gaussian_filter
-import matplotlib.pyplot as plt
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


skxray/cdi.py
```diff
@@ -281,6 +283,10 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
     n_iterations : int, optional
         number of iterations to run.
         default is 1000.
+    plot_function : function, optional
+        plotting function
```
ericdill: Note that this is a callback function that expects to receive these four objects: `sample_obj, obj_error, diff_error, sup_error`

skxray/cdi.py
```diff
@@ -281,6 +283,10 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
...
+    plot_step : int, optional
+        step interval for plotting
```
ericdill: what is the step interval?

skxray/cdi.py
```diff
@@ -360,6 +366,9 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
                     sup_error[n] = np.sum(sup_old)
                     sup_old = np.array(sup)
 
+        if plot_function and np.mod(n_iterations, plot_step) == 0:
```
ericdill: could also do `n_iterations % plot_step == 0` instead of `np.mod`

licode:
Done. thanks!

skxray/cdi.py
```diff
@@ -45,6 +45,7 @@
 import numpy as np
 import time
 from scipy.ndimage.filters import gaussian_filter
+import matplotlib.pyplot as plt
```
tacaswell: Do not need this import here

skxray/cdi.py
```diff
@@ -239,7 +240,8 @@ def generate_disk_support(sup_radius, shape_v):
 def cdi_recon(diffracted_pattern, sample_obj, sup,
               beta=1.15, start_avg=0.8, pi_modulus_flag='Complex',
               sw_flag=True, sw_sigma=0.5, sw_threshold=0.1, sw_start=0.2,
-              sw_end=0.8, sw_step=10, n_iterations=1000):
+              sw_end=0.8, sw_step=10, n_iterations=1000,
+              plot_function=None, plot_step=10):
```
tacaswell: I would not call it `plot_function` as the function could do _anything_.

licode:
Done!

skxray/cdi.py
```diff
@@ -281,6 +282,12 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
     n_iterations : int, optional
         number of iterations to run.
         default is 1000.
+    cb_function : function, optional
+        This is a callback function that expects to receive these
+        four objects: sample_obj, obj_error, diff_error, sup_error
```
tacaswell: You should also list the expected types.  Imaging reading this doc string with out having read the function, but wanting to write a callback to plug into this.

skxray/cdi.py
```diff
@@ -281,6 +282,12 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
...
+    plot_step : int, optional
```
tacaswell: change the 'plot' here -> 'cb'?

tacaswell:
Two minor doc/naming comments, but :+1: to merge as soon as those are fixed.

licode:
Updated.

PR #295, opened 2015-07-08
one time correlation experimental data fitted with ISF equation 

sameera2004:


sameera2004:
@tacaswell I rebased this branch , test are not passing because some lines are not covered (coverage issue)

sameera2004:
http://nbviewer.ipython.org/github/sameera2004/my_ipy_notebook/blob/master/Time_Correlation/ISF_fitting_repeat.ipynb

Ready for review

skxray/core/correlation.py
```diff
@@ -367,3 +380,130 @@ def extract_label_indices(labels):
...
+    baseline : float, optional
+        baseline of one time correlation
+        equal to one for ergodic samples
+
+    Returns
+    -------
+    g2 : array
+        normalized intensity-intensity time autocorreltion
+
+    Note :
```
tacaswell: Needs to be 

```python
Notes
-----
```

skxray/core/correlation.py
```diff
@@ -367,3 +380,130 @@ def extract_label_indices(labels):
...
+    :math ::
+        g1(q, t) = e^{-\\Gamma t}
+
+    :math ::
+        g2(q, t) = \\beta e^{-2\\Gamma t} + g_\\infty
+
+    These implementation are based on published work. [1]_
+
+     References
+    ----------
```
tacaswell: spacing is off

skxray/core/correlation.py
```diff
@@ -367,3 +380,130 @@ def extract_label_indices(labels):
...
+    """
+    return beta*np.exp(-2*relaxation_rate*lags) + baseline
+
+
+def _residual_auto_corr(params, lags, g2_data, *eps_data):
+    """
+    This will provide difference between experiment data and the model
+    Parameters
+    ----------
+    params : dict or Parameters
```
tacaswell: Should this be `dict of Paramaters`?

skxray/core/correlation.py
```diff
@@ -367,3 +380,130 @@ def extract_label_indices(labels):
...
+    eps_data : array
+
+    Returns
+    -------
+    final_result : array
+        minimized fitting to results using least square model
+
+    """
+    result = minimize(_residual_auto_corr, params, args=(x, data, eps_data))
+    return data + result.residual
```
tacaswell: Why do you add the residual to the data?

skxray/core/correlation.py
sameera2004: In lmfit  Parameters - (This is little more than a fancy dictionary)

ericdill: Can you change it to `lmfit.Parameters` then?

skxray/core/correlation.py
sameera2004: @tacaswell To get the final result after minimizing. Discussed this issue with @danielballan and removed that function because using lmfit Model class fitting can be done directly without another function.
(http://cars9.uchicago.edu/software/python/lmfit/model.html#model.Model)
example
-----------
from lmfit import Model
mod = Model(auto_corr_scat_factor)
fitted_result = mod.eval(lags=lags, beta=0.2234, relaxation_rate = 6.2567,  baseline=1.0)

sameera2004:
Ipython notebook for one time correlation experimental data fitted with ISF equation

http://nbviewer.ipython.org/github/sameera2004/my_ipy_notebook/blob/master/Time_Correlation/XPCS_ISF_fitting_using_lmfit_model.ipynb

sameera2004:
Thanks Tom

licode:
Is lmfit used in correlation.py? I didn't find it implemented yet, maybe in a new PR?

sameera2004:
Not implemented, users can use lmfit to fit the experimental data with the intermediate scattering factor(ISF) equation. In the ipython notebook I showed it as an example.

Before I used to put residual functions using lmfit  to get the experiment data and the model. using parameter class, (http://lmfit.github.io/lmfit-py/parameters.html). 
After discussing with @danielballan, I took out those functions and only keep the model. (http://lmfit.github.io/lmfit-py/model.html#the-model-class)

@licode Therefore, following line in the correlation.py can be removed,
from lmfit import minimize, Model, Parameters

licode:
I see. Because in the first ipython notebook example, corr.fit_auto_corr was implemented as a fitting function in correlation.py, but I couldn't find that function.

sameera2004:
@licode yes I took that out

ericdill:
Can you put in a PR against scikit-xray-examples with that ipython notebook?

sameera2004:
@ericdill sure I will

licode:
It is better to define that function in ipython notebook if you took it out from source code, so users can understand which function is used for fitting. If that function is standard in the field, I suggest to include it in lineshapes.py.

PR #297, opened 2015-07-15
Invert the library (All the changes)

ericdill:


skxray/io/__init__.py
```diff
@@ -32,5 +32,24 @@
 # IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE   #
 # POSSIBILITY OF SUCH DAMAGE.                                          #
 ########################################################################
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
```
tacaswell: ditch unicode literals here?

PR #298, opened 2015-07-17
ENH: Turn debug logger into a progressbar

ericdill:
- also add clint to the meta.yaml

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -25,6 +25,7 @@ requirements:
     - scikit-image
     - lmfit
     - netcdf4
+    - clint
 
 test:
   requires:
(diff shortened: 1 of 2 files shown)


ericdill:
This PR adds a progress bar to the long running dpc code, as seen in the following screenshot
screenshot from 2015-07-17 12 16 53

danielballan:
Neat. And, wheee more light useful dependencies! The README page for clint is hilarious.

danielballan:
I assume Travis is cranky because clint needs to go in the .travis.yml.

ericdill:
interesting note: clint is from the guy that maintains requests

ericdill:
re offline conversation with @tacaswell , this progressbar thing should instead just be log-level messages. to do that properly, we should implement something like the logging PR that is currently against trackpy at soft-matter/trackpy#184

PR #303, opened 2015-07-26
ENH: Adding codecov to CI

ericdill:


ericdill:
Self merging because it is a small PR that affects virtually nothing

PR #304, opened 2015-08-05
WIP: Img seg tool xfer to modified skxray

giltis:
Note: WIP

Per Eric's previous comments, moving tools specifically implemented for ease of use in VisTrails to VTtools. Tools designed to assist in plotting and vis are migrating to xray-vision. Tools to remain in skxray are being migrated to image.py, and the API structure is being used to simplify and differentiate based on function, functionality...


skxray/core/thresholding.py
```diff
@@ -0,0 +1,282 @@
...
+                         "resulting array fill value = {2}".format(
+                         thresh_value, np.amin(src_data), False))
+    vox_abv_min_value = (thresh_value_min <= src_data)
+    vox_abv_max_value = (thresh_value_max <= src_data)
+    output = vox_abv_min_value - vox_abv_max_value
+    return output
+
+
+def thresh_adapt(src_data, kernel_size, filter_type='gaussian'):
+    """
```
danielballan: What does this add to `skimage.threshold_adaptive`?

skxray/core/image.py
```diff
@@ -97,3 +97,58 @@ def _corr_ax1(input_image):
...
+        array
+    bin_edges : array
+        1xN array containing the edge values for each bin
+        NOTE: the length of this array is 1 larger than the length of the
+        hist array (e.g. len(bin_edges) = len(hist) + 1)
+    """
+
+    hist, bin_edges = np.histogram(src_data,
+                                   bins=num_bins,
+                                   density=pd_function)
```
danielballan: Everything below this line could be:

```
bin_avg = np.mean([bin_edges[1:], bin_edges[:-1]])
```

danielballan:
My thoughts, with @giltis over my shoulder here:

1. All of these should move to VTTools. They are simple functional wrappers of numpy functions.
skxray/segment/arithmetic/__init__.py
skxray/segment/arithmetic/logic.py
skxray/segment/arithmetic/math.py
skxray/segment/transformation.py

2. These import functions for scipy and skimage. In an earlier discussion we opted not to do this kind of thing. It aids "discovery" for newbies, but it risks confusing users about where they should go to look for docs or report docs. Instead of importing external tools, we will instead demonstrate them in our our docs.

    skxray/segment/filtering.py
    skxray/segment/morphology.py
    skxray/segment/registration.py <-- @giltis says he intends to put SimpleITK stuff in here also

3. skxray/segment/thresholding.py is an API to skxray/core/thresholding.py which has 7 functions.
These three wrap numpy and should be moved to VTTools.

    thresh_globalGT
    thresh_globalLT
    thresh_bounded

    `thresh_adapt` seems to add nothing to `skimage.threshold_adaptive`. @giltis will check this.

    These three wrap a skimage function that returns a threshold *value*, and they apply that threshold to the image. At minimum, we should credit that the underlying algorithm is from skimage. We should also try to understand why there is inconsistency in skimage (return threshold value vs. return applied threshold) and whether there is a good reason for it.

    thresh_otsu
    thresh_yen
    thresh_isodata

4. skxray/segment/histogram.py exposes two functions in skxray/core/image.py. One (`rescale_intensity`) is not implemented yet. The other is `hist_make`.

    `np.histogram` returns `(counts, bin_edges)`. `hist_make` returns `(counts, bin_edges, bin_centers)`. If we include this, I think it should have a different name and make more explicit its relationship to `np.histogram`.

5. That leaves skxray/segment/__init__.py, which might go away depending on the fates of the segment package's two remaining modules, histogram.py and threshold.py

skxray/core/image.py
```diff
@@ -97,3 +97,58 @@ def _corr_ax1(input_image):
...
+            False -- Histogram data reported simply as "counts" (e.g. Voxel
+                     Count)
+
+    Returns
+    -------
+    hist : array
+        1xN array containing all of the actual bin measurements (e.g. voxel
+        counts)
+    bin_avg : array
+        1xN array containing the average intensity value for each bin.
```
tacaswell: Is the the average (ex, the center) of the bins or the average of the values that are put into this bin?

skxray/core/image.py
```diff
@@ -97,3 +97,58 @@ def _corr_ax1(input_image):
...
+    Parameters
+    ----------
+    src_data : ndarray
+        Can be JxK or IxJxK
+        Specifies the source data set from which you want to evaluate the
+        histogram.
+    num_bins : int
+        Specify the number of bins to include in the histogram as an integer.
+
+    pd_function : bool, optional
```
tacaswell: A boolean flag should not be called 'function'

skxray/core/thresholding.py
```diff
@@ -0,0 +1,282 @@
...
+# Developed by Gabriel Iltis, Oct. 2013
+"""
+This module contains tools for thresholding data sets.
+"""
+
+
+import numpy as np
+import skimage.filter as sk
+
+#def check_saturation (src_data):
```
tacaswell: Don't included large amounts of commented out code.

skxray/core/image.py
giltis: I'm confused as to what you're asking here. Are you asking if it is basically the weighted average bin value? As implemented it is the bin center. This PR was posted mostly so that Dan and I could discuss, and agree upon where tools and functions most appropriately belong. The API focused files are all being moved to VTTools. This tool as well, largely aids in simplifying histogram plotting/histogram analysis, so perhaps this would be more appropriately located in VTTools as well?

skxray/core/image.py
giltis: noted. Thanks, didn't think of that (obviously).
Changed to pd_convert. actually, would changing this to "normalize" be more easily understandable?

PR #306, opened 2015-08-18
Fixed the issues with the tests in Eric's roi branch

sameera2004:


sameera2004:
@ericdill tests are ok, issue with pandas
ERROR: Failure: ImportError (No module named pandas)

sameera2004:
fixed the issue

ericdill:
I am :-1: on including `mean_intensity_sets` as a library function, as it only iterates over a list and calls another library function `mean_intensity`.

sameera2004:
@ericdill I took it out, now ready for review

.travis.yml
```diff
@@ -22,6 +22,7 @@ install:
   - python setup.py install
   - pip install coveralls
   - pip install codecov
+  - pip install pandas
```
ericdill: Can you `conda install pandas` instead? (Add it to line 18)

skxray/core/tests/test_roi.py
```diff
@@ -278,49 +276,35 @@ def test_static_test_sets():
...
-    #  when trying to combine the values
-    assert_raises(ValueError,
-                  lambda: roi.combine_mean_intensity(average_int_sets,
-                                                     index_list2))
+    # get the mean intensities of image sets given as a dictionary
+    roi_data = {}
+    for k, v in sorted(samples.items()):
+        intensity, index_list = roi.mean_intensity(v, label_array)
+        roi_data[k] = intensity
+    average_int_sets = pd.DataFrame(roi_data, index_list)
```
ericdill: Can you rework this test to not use pandas?  I am not sure that it is a good idea to pick up a pandas dependency in the test suite because of this minor use.

sameera2004:
@ericdill I took out the pandas dependency in these test, and rebased this PR. Ready for review.

skxray/core/roi.py
```diff
@@ -357,139 +355,87 @@ def roi_pixel_values(image, labels, index=None):
...
-    ----------
-    images : array
-        Intensity array of the images
-        dimensions are: (num_img, num_rows, num_cols)
-
-    labels : array
-        labeled array; 0 is background.
-        Each ROI is represented by a distinct label (i.e., integer).
-
+    mean_intensity : list
```
tacaswell: This is coming back an as array not a list

skxray/core/roi.py
```diff
@@ -357,139 +355,87 @@ def roi_pixel_values(image, labels, index=None):
...
-    else:
-        raise ValueError("Labels list for the image sets are different")
-
-    return combine_mean_intensity
+        # use a mean that is mask-aware
+        mean_intensity[n] = ndim.mean(img, labeled_array, index=index)
+    # turn the 2-D array back into a list of arrays because the rows and
+    # columns have different units
+    data = [mean_intensity[:, i] for i in range(mean_intensity.shape[1])]
+    roi_labels = ['roi_%s' % idx for idx in index]
```
tacaswell: What is the use for returning a list of strings here?  I think just returning `index` would be more useful here.

skxray/core/roi.py
```diff
@@ -531,7 +473,6 @@ def roi_kymograph(images, labels, num):
     """
     roi_kymo = []
     for n, img in enumerate(images):
-        roi_kymo.append((roi_pixel_values(img,
-                                          labels == num)[0])[0])
+        roi_kymo.append((roi_pixel_values(img, labels == num)[0])[0])
```
tacaswell: You can compute the label mask only once which will have a lot of time.

The two `[0]` makes is a bad code smell that makes it seem like there is something not quite right here.

skxray/core/roi.py
ericdill: I was matching the original API in my refactor of this code. I have no strong feelings either way

skxray/core/roi.py
ericdill: This could probably be done in some sort of horrible list comprehension too. Aren't those always faster?

ericdill: The first zero is because `roi_pixel_values` returns two values: the list of pixels in each roi and the list of roi indices.  The first `[0]` is to grab the list of pixels in each roi and the second `[0]` is to grab the first and only element out of that list of pixels.  I agree that this is pretty ugly. How should that roi_pixel_values API be changed?

skxray/core/roi.py
sameera2004: Sure I will take this out(returning strings)

skxray/core/roi.py
sameera2004: Fixed the issue of mean_intensity coming as array, it returns a list

skxray/core/roi.py
sameera2004: took out one [0] out

sameera2004:
Fixed the issues ready for review

skxray/core/roi.py
tacaswell: I would have just changed the documentation.  Arrays are far more useful than lists of lists

ericdill: Yeah, I'm with Tom on this one. Can you undo the change to the code and update the docs to say that it is coming back as an array?

.travis.yml
```diff
@@ -10,14 +10,17 @@ before_install:
   - chmod +x miniconda.sh
   - ./miniconda.sh -b -p /home/travis/mc
   - export PATH=/home/travis/mc/bin:$PATH
+  - wget https://gist.githubusercontent.com/tacaswell/128bb482f845feb024eb/raw/5cf21dc03a354fc87140d4a75e17cb5c076a0517/.condarc -O /home/travis/.condarc
```
ericdill: why is this being added back in to the travis yaml?

skxray/core/roi.py
```diff
@@ -357,139 +355,86 @@ def roi_pixel_values(image, labels, index=None):
...
-        shape (len(images), number of labels)
-
+        The labels for each element of the `mean_intensity` list
     """
-    if labels.shape != images[0].shape[0:]:
-        raise ValueError("Shape of the images should be equal to"
-                         " shape of the label array")
+    if labeled_array.shape != images[0].shape[0:]:
+        raise ValueError(
+            "`images` shape (%s) needs to be equal to the labeled_array` shape"
```
ericdill: you have an extra backtick ( ` ) after `labeled_array`

skxray/core/roi.py
```diff
@@ -531,7 +472,6 @@ def roi_kymograph(images, labels, num):
     """
     roi_kymo = []
     for n, img in enumerate(images):
-        roi_kymo.append((roi_pixel_values(img,
-                                          labels == num)[0])[0])
+        roi_kymo.append((roi_pixel_values(img, labels == num)[0]))
 
-    return np.matrix(roi_kymo)
+    return np.vstack(roi_kymo)
```
ericdill: can you remove the `roi_` prefix from the `roi_kymo` local variable?

.travis.yml
sameera2004: It is a mistake I will take it out in travis yaml

skxray/core/roi.py
sameera2004: Sure I will make it to return an array instead of a list

sameera2004:
Fixed ready for review

skxray/core/roi.py
```diff
@@ -359,137 +357,83 @@ def roi_pixel_values(image, labels, index=None):
...
-    else:
-        raise ValueError("Labels list for the image sets are different")
-
-    return combine_mean_intensity
+        # use a mean that is mask-aware
+        data[n] = ndim.mean(img, labeled_array, index=index)
+    # turn the 2-D array back into a list of arrays because the rows and
+    # columns have different units
+    mean_intensity = [data[:, i] for i in range(data.shape[1])]
+    return np.asarray(mean_intensity), index
```
ericdill: It is entirely unclear why you are laundering this numpy array through a list and back to a numpy array...

sameera2004:
@ericdill now the mean_intensity only return as a bumpy array, took out the list

skxray/core/roi.py
```diff
@@ -359,137 +357,80 @@ def roi_pixel_values(image, labels, index=None):
...
-
-    mean_intensity = np.zeros((images.shape[0], index.shape[0]))
+        index = list(np.unique(labeled_array))
+        index.remove(0)
+    try:
+        len(index)
+    except TypeError:
+        index = [index]
+    # not sure that this is needed
+    index = np.asarray(index)
```
ericdill: This line is not needed. The index argument to `ndim.mean` just needs to be a sequence, not a numpy array

skxray/core/utils.py
```diff
@@ -602,32 +602,30 @@ def bin_1D(x, y, nx=None, min_x=None, max_x=None):
...
-    Make a grid of radial positions.
+    """Convert a cartesian grid (x,y) to the radius relative to some center
 
     Parameters
     ----------
     center : tuple
         point in image where r=0; may be a float giving subpixel precision.
         Order is (rr, cc).
-
     shape: tuple
```
ericdill: add a whitespace between `shape` and `:`

skxray/core/utils.py
```diff
@@ -602,32 +602,30 @@ def bin_1D(x, y, nx=None, min_x=None, max_x=None):
...
-        pixel coordinates. Order is (rr, cc).
+        pixel coordinates.
+        Order is (rr, cc).
 
     Returns
     -------
     r : array
-        The L2 norm of the distance of each pixel from the calibrated center.
+        The distance of each pixel from `center`
+        Shape of the array is equal to `shape`
```
ericdill: `Shape of the return value is equal to the `shape` input parameter

skxray/core/utils.py
ericdill: Also add 
```
pixel_size : sequence, optional
    The physical size of the pixels.
    len(pixel_size) should be the same as len(shape)
    defaults to (1,1)
```

ericdill:
Great, thanks for all the work!

sameera2004:
Sure, Thank you for all the help!

PR #310, opened 2015-08-21
Rename Nikea to scikit-xray

ericdill:


conda-recipe/find_modules.py
```diff
@@ -0,0 +1,20 @@
...
+        continue
+    print(tup)
+    if '__init__.py' in tup[2]:
+        package = '.'.join(tup[0].split(os.sep)[1:])
+        to_import.append(package)
+        for mod in tup[2]:
+            if mod == '__init__.py':
+                continue
+            to_import.append(package + '.' + mod.split('.')[0])
+
```
licode: This is a smart way to find modules. Do we need to worry about files in tup[1]?

ericdill: tup[1] lists the directories that are in tup[0]

licode: Ah, got it.

PR #316, opened 2015-08-25
BLD/CLN: Isolate dependencies.

danielballan:
* Do a little more to isolate parts of the code that require particular dependencies from those that work just fine with numpy alone.
* Add a `requirements.txt `and an `optional-requirements.txt` to clarify dependencies for users.
* Also split essential and optional dependencies in docs.
* Pin the version of lmfit to 0.8.1. We are broken on 0.8.2 and 0.8.3 (latest) and no one has noticed. @licode can you look at this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,9 @@ before_install:
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
   - conda update conda --yes
-  - conda create -n testenv --yes pip nose python=$TRAVIS_PYTHON_VERSION lmfit xraylib numpy scipy scikit-image netcdf4 six coverage
+  - conda create -n testenv --yes pip nose python=$TRAVIS_PYTHON_VERSION xraylib numpy scipy scikit-image netcdf4 six coverage
   - source activate testenv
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -15,8 +15,9 @@ before_install:
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
   - conda update conda --yes
-  - conda create -n testenv --yes pip nose python=$TRAVIS_PYTHON_VERSION lmfit xraylib numpy scipy scikit-image netcdf4 six coverage
+  - conda create -n testenv --yes pip nose python=$TRAVIS_PYTHON_VERSION xraylib numpy scipy scikit-image netcdf4 six coverage
   - source activate testenv
+  - pip install lmfit==0.8.1
```
ericdill: why 0.8.1 and not 0.8.3?

danielballan: Drink some coffee and read the PR again. :- )

ericdill: This means that we are pinning 0.0.4 to lmfit 0.8.1, correct?

PR #318, opened 2015-08-25
Update docs

ericdill:
~~Supersedes #316~~


ericdill:
This is good to go. I'll push the v0.0.4 tag whenever this gets merged

doc/resource/api/core.rst
```diff
@@ -1,79 +0,0 @@
-====================
- :mod:`core` Module
-====================
```
danielballan: I don't think this should have been nuked. I can't find a replacement elsewhere.

danielballan:
Other that file that I don't think should have been removed, feel free to self-merge. Tagging suggestion:

```
git commit --allow-empty -m "REL: v0.0.4rc1"
git tag v0.0.4rc1
git push upstream --tags
```

If everything seems to work (check `__version__`) then repeat, dropping the rc1.

ericdill:
good catch

PR #325, opened 2015-09-08
fitting for xsvs data 

sameera2004:
Negative binomial distribution
Poisson distribution 
Gamma distribution

### DIFF
--- a/skxray/core/speckle_fitting.py
+++ b/skxray/core/speckle_fitting.py
@@ -1,0 +1,175 @@
+# ######################################################################
+# Copyright (c) 2014, Brookhaven Science Associates, Brookhaven        #
+# National Laboratory. All rights reserved.                            #
+#                                                                      #
+# Original code:                                                       #
+# @author: Andrei Fluerasu, Brookhaven National Laboratory and         #
+# Pawel Kwasniewski, European Synchrotron Radiation Facility           #
+#                                                                      #
+# Developed at the NSLS-II, Brookhaven National Laboratory             #
+# Developed by Sameera K. Abeykoon, June 2015                          #
+#                                                                      #
+# Redistribution and use in source and binary forms, with or without   #
+# modification, are permitted provided that the following conditions   #
+# are met:                                                             #
+#                                                                      #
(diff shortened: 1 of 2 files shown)


skxray/core/speckle_fitting.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    co_eff = np.exp(gammaln(bin_edges + M) -
+                    gammaln(bin_edges + 1) - gammaln(M))
+
+    poisson_gamma = co_eff * np.power(M / (K + M), M) * np.power(K / (M + K),
+                                                                 bin_edges)
+    return poisson_gamma
+
+
+def poisson_distribution(bin_edges, K):
```
ericdill: Why are you not using scipy.stats.poisson?

skxray/core/speckle_fitting.py
```diff
@@ -0,0 +1,175 @@
...
+    nbinom_distribution() function Notes
+
+    :math ::
+        P(K) = \frac{<K>^K}{K!}\exp(-K)
+    """
+    poisson_dist = np.exp(-K) * np.power(K, bin_edges)/gamma(bin_edges + 1)
+    return poisson_dist
+
+
+def gamma_distribution(bin_edges, K, M):
```
ericdill: Does this do anything that `scipy.stats.gamma` does not do?

http://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.stats.gamma.html

ericdill:
Don't all of these functions already exist in scipy?

licode:
I suggest to put these functions into lineshapes https://github.com/scikit-xray/scikit-xray/blob/master/skxray/core/fitting/lineshapes.py if they are not included in scipy.

danielballan:
+1 to Li's comment

sameera2004:
@Li yes in the future I will add these fitting functions to lineshapes https://github.com/scikit-xray/scikit-xray/blob/master/skxray/core/fitting/lineshapes.py
@ericdill yes it looks like most of these functions already exist in scipy. Thanks! 
Looks like two are not there, anyway I will check back in scipy if so I will put new PR,

PR #339, opened 2015-09-19
MNT: Add example notebook links to tutorial page

ericdill:


licode:
Looks good except that powder calibration is still broken?

ericdill:
Yes, it is broken. The underlying math is also wrong, so it is a great deal more work to get this working than simply updating some function calls

Issue #338, opened 2015-09-19
build skxray v0.0.5 for linux


ericdill:
I am having an odd issue with my docker image (github.com/ericdill/conda-builder) that I have been using to build debian 7 conda binaries.  Need to get that sorted out.  

It is roughly this:

```
root@55e0815ff4ff:/local-src/conda/conda-prescriptions/recipes# conda install scipy
Fetching package metadata: ......
Solving package specifications: .
Package plan for installation in environment /opt/conda:

The following NEW packages will be INSTALLED:

    numpy:    1.9.2-py27_2
    openblas: 0.2.14-2
    scipy:    0.16.0-np19py27_1

Proceed ([y]/n)? y

Linking packages ...
[      COMPLETE      ]|###############################################################################################################################################################################| 100%
root@55e0815ff4ff:/local-src/conda/conda-prescriptions/recipes# python -c "from scipy.special import comb, gamma"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/opt/conda/lib/python2.7/site-packages/scipy/special/__init__.py", line 601, in <module>
    from ._ufuncs import *
ImportError: libgfortran.so.3: cannot open shared object file: No such file or directory
```

This problem is affecting the test portion of the conda-build for scikit-xray as the `import skxray.diffraction` fails with the same traceback:

```
===== testing package: scikit-xray-v0.0.5-nppy27_0 =====
import: u'skxray'
import: u'skxray.diffraction'
Traceback (most recent call last):
  File "/opt/conda/conda-bld/test-tmp_dir/run_test.py", line 28, in <module>
    import skxray.diffraction
  File "/opt/conda/envs/_test/lib/python2.7/site-packages/skxray/diffraction.py", line 47, in <module>
    from skxray.core.fitting import Lorentzian2Model
  File "/opt/conda/envs/_test/lib/python2.7/site-packages/skxray/core/fitting/__init__.py", line 43, in <module>
    from .background import snip_method
  File "/opt/conda/envs/_test/lib/python2.7/site-packages/skxray/core/fitting/background.py", line 45, in <module>
    import scipy.signal
  File "/opt/conda/envs/_test/lib/python2.7/site-packages/scipy/signal/__init__.py", line 280, in <module>
    from .bsplines import *
  File "/opt/conda/envs/_test/lib/python2.7/site-packages/scipy/signal/bsplines.py", line 12, in <module>
    from scipy.special import comb, gamma
[...]

PR #340, opened 2015-09-21
Use scipy=0.15.1

ericdill:
See #338


ericdill:
Closes #338

ericdill:
As a hotfix, are there any objections to using scipy=0.15.1 until there is bandwidth to sort out why I cannot build with scipy=0.16.0?

PR #343, opened 2015-09-21
Dockerfile

chuckie82:
Added a Dockerfile for running skxray on any machine with Docker installation.
Currently, only installs requirements.txt


ericdill:
Cool!

PR #344, opened 2015-10-01
Remove coveralls badge and add gitter badge

ericdill:
- coveralls service has been intermittent recently
- gitter is cool and we should make more use of it


tacaswell:
@ericdill Please be careful about not pushing branches to upstream

ericdill:
@tacaswell if it makes you feel any better I did not mistakenly push this branch to upstram, I edited it in the browser and made a PR from github.com. I did not realize that the branch would be created on upstream though!

tacaswell:
ah, fair enough.

PR #348, opened 2015-11-23
Fitting functions for XSVS

sameera2004:


ericdill:
closed and reopened to force travis to use the new master branch that has lmfit pinned to 0.8.3 on travis

skxray/core/fitting/lineshapes.py
```diff
@@ -385,3 +391,100 @@ def compton(x, compton_amplitude, coherent_sct_energy,
     counts += value
 
     return counts
+
+
+def gamma_dist(bin_values, K, M):
```
ericdill: is it important that `K` and `M` are capitalized?

tacaswell: wikipedia says κ and θ  If we drop 2.7 we could use those as the variables!

https://en.wikipedia.org/wiki/Gamma_distribution

tacaswell:
:+1: modulo the concerns about the names.

skxray/core/fitting/lineshapes.py
sameera2004: It is not important to keep K and M capitalized, sure we can change to  κ and θ when we drop 2.7 .

tacaswell: I was mostly being obnoxious with by unicode comment.

ericdill:
lgtm

Issue #92, opened 2014-08-21
2-time correlation


ericdill:
# Description

Part of X-ray photon correlation spectroscopy (XPCS).  See [pyXPCS](https://github.com/NSLS-II/pyXPCS)

PR #352, opened 2015-12-02
Two Time Correlation

sameera2004:


sameera2004:
weblink for the python notebook
http://nbviewer.ipython.org/github/sameera2004/my_ipy_notebook/blob/master/Time_Correlation/XPCS%20Two-time%20SAXS%20example.ipynb

ericdill:
closed/opened to restart travis now that we have test coverage again

ericdill:
closes #92

sameera2004:
This PR is ready for review, please go through the two_time_corr function, I will fix the merge conflicts

skxray/core/correlation.py
```diff
@@ -402,3 +351,366 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+            for i in range(-nshift+1, nshift+1):
+                two_time[int(tind1+i),
+                           int(tind2+i)] = (tmp_binned/(pi_binned *
+                                                        fi_binned))*num_pixels
+        else:
+            two_time[tind1, tind2] = tmp_binned/(pi_binned *
+                                                   fi_binned)*num_pixels
+
+
+def _help_process(level, num_bufs, buf_no, i, buf, label_array):
```
tacaswell: This needs a better name.

sameera2004: @tcaswell sure, I will change the name of this function. anymore comments?

tacaswell: I thought you were on vacation!  I only started to read through the code so more comments should be coming.

sameera2004: Only checking my e-mails still having jet lag, wake up around 4.30am (Sri Lankan time), better than last week ( 2.30am) So nothing to do until others wake up. Sure leave comments , I will fix after I come back,

skxray/core/correlation.py
```diff
@@ -402,3 +351,366 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+
+    if np.any(num_pixels == 0):
+        raise ValueError("Number of pixels of the required roi's"
+                         " cannot be zero, "
+                         "num_pixels = {0}".format(num_pixels))
+
+    return label_array, indices, num_rois, num_pixels
+
+
+def one_time_from_two_time(two_time_corr):
```
tacaswell: This will certainly be useful for testing as well as science!

tacaswell:
@sameera2004 @ericdill The code for the two correlation functions looks very similar with the exception of storing a 2D correlation result and a different inner function.  Is there a way for the two functions to be merged so we only have one copy of the subtle multi-tua averaging code?

skxray/core/correlation.py
```diff
@@ -402,3 +351,366 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+                                  current_img_time, level=level,
+                                  buf_no=cur[level]-1)
+                level += 1
+
+                # Checking whether there is next level for processing
+                processing = level < num_levels
+
+    for q in range(np.max(labels)):
+        x0 = two_time[:, :, q]
+        two_time[:, :, q] = (np.tril(x0) + np.tril(x0).T
```
tacaswell: Is there data in the upper triangle that is being over-written here?

It is probably better to return the un-patched up result and provide a function to do this as needed.

If there is not data in the upper triangle would `x0 + np.tril(x0, -1)` would be faster.

skxray/core/correlation.py
```diff
@@ -402,3 +351,366 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+    for i in range(i_min, min(img_per_level[level], num_bufs)):
+        (t_index, tmp_binned, pi_binned,
+         fi_binned) = _help_process(level, num_bufs, buf_no, i, buf,
+                                    label_array)
+
+        tind1 = (current_img_time - 1)
+
+        tind2 = (current_img_time - lag_steps[t_index] - 1)
+
+        if not isinstance(current_img_time, int):
```
tacaswell: When would this not be an integer?

skxray/core/correlation.py
```diff
@@ -402,3 +351,366 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+            nshift = 2**(level-1)
+            for i in range(-nshift+1, nshift+1):
+                two_time[int(tind1+i),
+                           int(tind2+i)] = (tmp_binned/(pi_binned *
```
tacaswell: You can do this computation outside of the loop.

skxray/core/correlation.py
sameera2004: This doesn't works for multi tau, but helpful when people are doing two time correlation for all the frames. They can get the one time correlation data using this method.

skxray/core/correlation.py
sameera2004: There is no data in the upper triangle, so I will this way

skxray/core/correlation.py
sameera2004: Sure, I will

sameera2004:
Added new PR, closing this due to merge conflicts

PR #368, opened 2015-12-10
added multi_tau_auto_corr_partial_data function

bfrosik:
The new function is a modified copy of multi_tau_auto_corr function. It supports partial data input when calculation correlation. Each iteration takes previous results.

### DIFF
--- a/skxray/core/correlation.py
+++ b/skxray/core/correlation.py
@@ -328,31 +328,32 @@ def multi_tau_auto_corr_partial_data(num_levels, num_bufs, labels, images, previ
     # level has twice the delay times of the next lower one. To save
     # needless copying, of cyclic storage of images in buf is used.
 
+
+    if num_bufs % 2 != 0:
+        raise ValueError("number of channels(number of buffers) in "
+                            "multiple-taus (must be even)")
+
+    if hasattr(images, 'frame_shape'):
+        # Give a user-friendly error if we can detect the shape from pims.
+        if labels.shape != images.frame_shape:
+            raise ValueError("Shape of the image stack should be equal to"
+                                " shape of the labels array")
+
+    # get the pixels in each label
+    label_mask, pixel_list = roi.extract_label_indices(labels)
+
+    num_rois = np.max(label_mask)
+
+    # number of pixels per ROI
+    num_pixels = np.bincount(label_mask, minlength=(num_rois+1))
+    num_pixels = num_pixels[1:]
+
+    if np.any(num_pixels == 0):
+            raise ValueError("Number of pixels of the required roi's"
+                            " cannot be zero, "
+                            "num_pixels = {0}".format(num_pixels))
+
     if previous is None:
-        if num_bufs % 2 != 0:
-            raise ValueError("number of channels(number of buffers) in "
-                             "multiple-taus (must be even)")
-
-        if hasattr(images, 'frame_shape'):
-            # Give a user-friendly error if we can detect the shape from pims.
-            if labels.shape != images.frame_shape:
-                raise ValueError("Shape of the image stack should be equal to"
-                                 " shape of the labels array")
-
-        # get the pixels in each label
-        label_mask, pixel_list = roi.extract_label_indices(labels)
-
-        num_rois = np.max(label_mask)
-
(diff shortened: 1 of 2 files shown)


skxray/core/correlation.py
```diff
@@ -253,6 +253,211 @@ def multi_tau_auto_corr(num_levels, num_bufs, labels, images):
...
+        maximum lag step to compute in each generation of
+        downsampling
+
+    labels : array
+        labeled array of the same shape as the image stack;
+        each ROI is represented by a distinct label (i.e., integer)
+
+    images : iterable of 2D arrays
+        dimensions are: (rr, cc)
+
```
tacaswell: Please document `previous`

skxray/core/correlation.py
```diff
@@ -253,6 +253,211 @@ def multi_tau_auto_corr(num_levels, num_bufs, labels, images):
...
+                # on previous call above.
+                _process(buf, G, past_intensity_norm,
+                         future_intensity_norm, label_mask,
+                         num_bufs, num_pixels, img_per_level,
+                         level=level, buf_no=cur[level]-1,)
+                level += 1
+
+                # Checking whether there is next level for processing
+                processing = level < num_levels
+
```
tacaswell: I think the `yield` should be here so we can get the updated `g2` as it runs?

skxray/core/correlation.py
```diff
@@ -253,6 +253,211 @@ def multi_tau_auto_corr(num_levels, num_bufs, labels, images):
...
+        # to track how many images processed in each level
+        img_per_level = np.zeros(num_levels, dtype=np.int64)
+
+        start_time = time.time()  # used to log the computation time (optionally)
+
+    else:
+        G = previous[2]
+        past_intensity_norm = previous[3]
+        future_intensity_norm = previous[4]
+        buf = previous[5]
```
tacaswell: python has nice unpacking, these lines can be written as 

```python
G, past_intensity_norm, future_intensity_norm, buf = previous[2:]
```

skxray/core/correlation.py
```diff
@@ -253,6 +253,211 @@ def multi_tau_auto_corr(num_levels, num_bufs, labels, images):
...
+    """
+    # In order to calculate correlations for `num_bufs`, images must be
+    # kept for up to the maximum lag step. These are stored in the array
+    # buffer. This algorithm only keeps number of buffers and delays but
+    # several levels of delays number of levels are kept in buf. Each
+    # level has twice the delay times of the next lower one. To save
+    # needless copying, of cyclic storage of images in buf is used.
+
+    if previous is None:
+        if num_bufs % 2 != 0:
```
tacaswell: This whole block of code should probably be pulled out into a function that can be re-used by both this an the existing multi tau function.

skxray/core/correlation.py
```diff
@@ -253,6 +253,211 @@ def multi_tau_auto_corr(num_levels, num_bufs, labels, images):
...
+    # the normalization factor
+    if len(np.where(past_intensity_norm == 0)[0]) != 0:
+        g_max = np.where(past_intensity_norm == 0)[0][0]
+    else:
+        g_max = past_intensity_norm.shape[0]
+
+    # g2 is normalized G
+    g2 = (G[:g_max] / (past_intensity_norm[:g_max] *
+                       future_intensity_norm[:g_max]))
+
```
sameera2004: I think 'yield' can come here and the tot_channels and lag_steps are taken out from this function. Because we can find it separately

skxray/core/correlation.py
sameera2004: Yes same block is need for two time correlation, So I have taken this out and made small helper function for that. I think this can use that too

skxray/core/correlation.py
tacaswell: but we want to get a partial g2 for every image.

ericdill:
closed/opened to restart travis now that we have test coverage again

sameera2004:
@ericdill Thanks for fixing the test coverage

tacaswell:
@bfrosik Thanks for your work on this!

PR #370, opened 2015-12-11
Document making the docs

ericdill:


ericdill:
The docs are now live http://scikit-beam.github.io/scikit-beam/

danielballan:
Thanks!

PR #371, opened 2015-12-11
Fixing up the skxray test suite.  It was not actually running on travis

ericdill:
- backing out changes to the .travis.yml file that disabled running tests.  
- fixing up new tests that were failing because we missed that the tests were not being run.  
- turning on python 3.5 on travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
   - conda config --set always_yes true
   - conda update conda
-  - conda create -n testenv pip nose python=$TRAVIS_PYTHON_VERSION numpy scipy scikit-image six coverage
-  - conda install -n testenv -c scikit-xray xraylib lmfit=0.8.3 netcdf4
+  - conda config --add channels scikit-xray
+  - conda create -n testenv pip nose python=$TRAVIS_PYTHON_VERSION numpy scipy scikit-image six coverage xraylib lmfit=0.8.3 netcdf4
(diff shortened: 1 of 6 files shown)


ericdill:
Ok, tests are passing now.  Whew.  This should go in immediately since our tests are not being run on the master branch.  Therefore **all** PRs that are coming in look like they are ok.

.travis.yml
```diff
@@ -15,16 +16,19 @@ install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
   - conda config --set always_yes true
   - conda update conda
-  - conda create -n testenv pip nose python=$TRAVIS_PYTHON_VERSION numpy scipy=0.15.1 scikit-image six coverage
-  - conda install -n testenv -c scikit-xray xraylib lmfit=0.8.3 netcdf4
+  - conda config --add channels scikit-xray
```
danielballan: I preferred the clarify of two lines, telling me which packages come from us and which come from Continuum.

ericdill: I can see your point.  However, travis is unecessarily thrashing because of this two line install.  it is first installing  numpy 1.10 on line 18.  On line 19, because we dont have numpy 1.10 builds of some packages on our scikit-xray channel, conda is installing numpy 1.9 and then it has to reinstall a bunch of packages that were compiled against numpy 1.10 instead of 1.9.  Travis takes ~25% longer to run with these installs on two lines instead of one.

See [here] (https://travis-ci.org/scikit-beam/scikit-beam/jobs/96243412#L360).

PR #373, opened 2015-12-11
Doc rewrite

tacaswell:


ericdill:
What is the best way to do this?  Should I just merge this and then people can propose edits via PR if there is any contention?

danielballan:
Let's let it sit for a week so it is easy for people to leave line comments.

tacaswell:
The docs also do not currently build, that should be fixed before it is
merged.

ericdill:
Am I remembering correctly that we decided to move this to the top-level scikit-beam organization docs?

ericdill:
cc @tacaswell @danielballan

doc/index.rst
```diff
@@ -1,41 +1,149 @@
...
 
+The :ref:`vision<vision>` of `scikit-beam` is to provide simple
+functions useful for the X-ray, Neutron and Electron communities.  The
+primary goal of the scikit-beam project is to provide a centralized
+repository for algorithms that are used in these three scientific
+domains.  `scikit-beam` functions accept and return standard Python
+and numpy datatypes, so they integrate well with other packages from
+the scientific Python community.  Further, the modular design of
+scikit-beam allows its components to be easily reused in ways not
+envisioned by the authors.
```
tacaswell: @chrisvam Comments on this paragraph?

tacaswell:
@chrisvam The inventory of what we have implemented has been side-tracked my making sure the docs build with out errors.

doc/index.rst
chrisvam: Looks good.  My only thought is that we do have classes that maintain state (for performance) so maybe "functions" should be changed to "functions/classes"?

chrisvam:
Hi Tom,

No worries about the delay.  You and Dan should have received a link to a google-doc (that you can edit) that is my zeroth-order suggestion for what to send to the silx people.  In a nutshell:  I think we should see if we can agree on a hopefully-simple list of requirements for software we can all work on.  And I believe skbeam fulfills the ones I have written down so far.

danielballan:
Normal Travis builds passed. Only the doc build failed, on warning-as-error from sphinx. Merging as is and basing further work on master.

PR #374, opened 2015-12-11
Add fast, incremental cython-based histogram

ericdill:
This is in its final stages of review. If anyone else wants to make suggestions or comment, now is the time!

### DIFF
--- a/skxray/core/accumulators/histogram.pyx
+++ b/skxray/core/accumulators/histogram.pyx
@@ -11,6 +11,7 @@ from ..utils import bin_edges_to_centers
 import logging
 logger = logging.getLogger(__name__)
 
+DEF MAX_DIMENSIONS = 10
 
 ctypedef fused xnumtype:
     np.int_t
@@ -30,6 +31,9 @@ cdef void* _getarrayptr(np.ndarray a):
 
 
 class Histogram:
+
+    _always_use_fillnd = False      # FIXME remove this
+
     def __init__(self, binlowhigh, *args):
         """
 
@@ -48,11 +52,9 @@ class Histogram:
         -----
         The right most bin is half open
         """
-        if len(args) > 1:
-            raise NotImplementedError(
-                "This class does not yet support higher dimensional histograms "
-                "than 2D"
-            )
+        if 1 + len(args) > MAX_DIMENSIONS:
+            emsg = "Cannot create histogram of more than {} dimensions."
+            raise ValueError(emsg.format(MAX_DIMENSIONS))
         logger.debug('binlowhigh = {}'.format(binlowhigh))
         logger.debug('args = {}'.format(args))
         nbins = []
@@ -83,6 +85,7 @@ class Histogram:
         """
         self._values.fill(0)
 
+
     def fill(self, *coords, weights=1):
         """
 
@@ -110,9 +113,12 @@ class Histogram:
                 emsg = "Coordinate arrays must have the same length."
                 raise ValueError(emsg)
         if len(weights) != 1 and len(weights) != nexpected:
-            emsg = "Weights must be scalar or have the same length as coordinates."
+            emsg = ("Weights must be scalar or have the same length "
+                    "as coordinates.")
             raise ValueError(emsg)
-
+        if self._always_use_fillnd:
+            self._fillnd(coords, weights)
+            return
         if len(coords) == 1:
             # compute a 1D histogram
             self._fill1d(coords[0], weights)
@@ -121,7 +127,7 @@ class Histogram:
             self._fill2d(coords[0], coords[1], weights)
         else:
             # do the generalized ND histogram
-            raise NotImplementedError()
+            self._fillnd(coords, weights)
         return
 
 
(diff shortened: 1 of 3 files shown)


ericdill:
I am getting an occasional out of bounds error on the upper end of the binning code that I would imagine is due to floating point precision or something?

Here is the code needed to replicate what I'm seeing:

```python
import numpy as np
from skxray.core.accumulators.histogram import Histogram

h = Histogram([10, 0, 10.1])
for i in range(1000):
    x = np.random.random(1000000)*40
    w = 1
    h.fill(x, weights=w)

    h.reset()
```

Running the above code should generate an IndexError along the lines of:

```
$ python timings.py
Traceback (most recent call last):
  File "timings.py", line 10, in <module>
    h.fill(x, weights=w)
  File "skxray/core/accumulators/histogram.pyx", line 118, in skxray.core.accumulators.histogram.Histogram.fill (skxray/core/accumulators/histogram.c:3752)
    self._fill1d(coords[0], weights)
  File "skxray/core/accumulators/histogram.pyx", line 141, in skxray.core.accumulators.histogram.Histogram._fill1d (skxray/core/accumulators/histogram.c:6160)
    data[xidx] += weight[wstride * i]
IndexError: Out of bounds on buffer access (axis 0)
```

or 
```
$ python timings.py
Traceback (most recent call last):
  File "timings.py", line 30, in <module>
    timethis('x, y, w = gendata(); h.fill(x, y, weights=w)'))
  File "timings.py", line 16, in timethis
    return np.mean(timeit.repeat(stmt, number=10, repeat=10, globals=gg))
  File "/Users/edill/mc/envs/skbeam35/lib/python3.5/timeit.py", line 218, in repeat
    return Timer(stmt, setup, timer, globals).repeat(repeat, number)
  File "/Users/edill/mc/envs/skbeam35/lib/python3.5/timeit.py", line 206, in repeat
    t = self.timeit(number)
  File "/Users/edill/mc/envs/skbeam35/lib/python3.5/timeit.py", line 178, in timeit
    timing = self.inner(it, self.timer)
  File "<timeit-src>", line 6, in inner
  File "skxray/core/accumulators/histogram.pyx", line 129, in skxray.core.accumulators.histogram.Histogram.fill (skxray/core/accumulators/histogram.c:3936)
    self._fill2d(coords[0], coords[1], weights)
  File "skxray/core/accumulators/histogram.pyx", line 171, in skxray.core.accumulators.histogram.Histogram._fill2d (skxray/core/accumulators/histogram.c:11577)
    data[xidx][yidx] += weight[wstride * i]
IndexError: Index out of bounds (axis 0)
```

If I surround the offending block in the cython code with a try/except block that catches this index error and prints out the offending value, we see this:

```
In [5]: %run timings.py
Having trouble indexing value=10.100000053772984. xidx=10
Having trouble indexing value=10.100000191044401. xidx=10
Having trouble indexing value=10.099999919957048. xidx=10
Having trouble indexing value=10.10000006734455. xidx=10
Having trouble indexing value=10.100000236211049. xidx=10
Having trouble indexing value=10.099999932871162. xidx=10
Having trouble indexing value=10.099999950901681. xidx=10
Having trouble indexing value=10.09999997167224. xidx=10
Having trouble indexing value=10.100000280200572. xidx=10
Having trouble indexing value=10.100000367277815. xidx=10
Having trouble indexing value=10.100000102750485. xidx=10
Having trouble indexing value=10.100000378813313. xidx=10
Having trouble indexing value=10.100000201542208. xidx=10
Having trouble indexing value=10.10000000442588. xidx=10
```

and the diff
```diff
++        try:
++            for i in range(xlen):
++                xidx = find_indices(xval[i], low, high, binsize)
++                if xidx == -1:
++                    continue
++                w = weight[wstride * i]
++                data[xidx] += w
++        except IndexError:
++            print("Having trouble indexing value=%s. xidx=%s" % (xval[i], xidx))
++            raise
```

Can anyone else replicate this?

ordirules:
About the floating point errors, I can try later this evening when I have time.

I have a question about the benchmark. Have you tried setting _always_use_fillnd = True  and running the 1d and 2d histograms again for the benchmark? (I remember you commented about this somewhere but I cannot seem to find it)

I tried my  2D simulations and they did not work setting ```_always_use_fillnd = True```
Ideally, using fillnd or fill2d should yield the same result. As far as I can read from the fillnd code, it looks like it should work and no indices seem to be swapped or anything (@pavoljuhas, very clever to use merge sort to keep indice order :-) ). I'll try to dig in tonight.

ordirules:
Okay, I found the bug... It took me long enough... Eric, try this in the 2d histogram tester (I can't commit this in in any clean way, easier you just type this in):
```        [[ten, nine], xf, yi, wi]
```

remember that phase space exploration thing we talked about? My random walk helped find this one ;-)
(basically the only combination I happened to try did not work with my 2d simulations)

I think this suggests we'll need to be a little more rigorouos on testing fillnd. What do you think we should do?

This puzzled me for quite a while, I saw in your code that you tested fillnd with np.histogram2d and for a second I actually thought I had been sucked into some anarchistic parallel universe.

pavoljuhas:
@ericdill - about [floating point issues](https://github.com/scikit-beam/scikit-beam/pull/374#issuecomment-164211809)  Looks like _lows, _highs, _binsizes have only `float32` precision, but the values are `float64`.  I'd guess an out of bound doubles get rounded to in-bond floats.  Try to use float64 for _lows, _highs, _binsizes.

ordirules:
looks good. tested it on my simulations and results are sensible. Also tested a 3D version and it works perfectly!

ericdill:
@pavoljuhas good idea. Changing all floats to doubles does indeed solve this problem.  Should I still keep the underlying `_values` array as a `float` or should this also be a `double`?

ericdill:
Ah, I meant to cc @ordirules and @chrisvam on this too

ericdill:
@ordirules Can you submit your histogramming notebooks to scikit-xray-examples?

ordirules:
yeah, it would be good to document. I assume I just send a PR to scikit-xray-examples, is that correct?

ericdill:
@ordirules Indeed

ordirules:
ok thanks, submitted. I'll try to clean it up later and add a sample usage of the 3D version, to at least document that some form of visual testing on higher order historgams was performed.

ericdill:
I had to locally merge this and fix setup.py.  @stuwilkins deleted `setup_ext.py` which we were making use of to get the cython code compiled.  Thanks for all your work @ordirules @pavoljuhas @chrisvam !

PR #383, opened 2015-12-17
DOC: update scikit-xray -> scikit-beam

ericdill:
Update to reflect the org name change from scikit-xray to scikit-beam
and the repo name change from scikit-xray to scikit-beam. Note that
there is still a major overhaul needed to rename the skxray python
package to skbeam


ericdill:
Non-controversial PR. Self-merging to get our badges working again

PR #384, opened 2015-12-18
Histogram adjust dtype

pavoljuhas:
Use uniform numerical dtype for internal floating point numpy arrays in Histogram.  Also clean up unused C-arrays from the cython part.

### DIFF
--- a/skxray/core/accumulators/histogram.pyx
+++ b/skxray/core/accumulators/histogram.pyx
@@ -175,10 +175,8 @@ class Histogram:
     def _fillnd(self, coords, np.ndarray[wnumtype, ndim=1] weight):
         # allocate pointer arrays per each supported numerical types
         cdef np.int_t* aint_ptr[MAX_DIMENSIONS]
-        cdef int aint_stride[MAX_DIMENSIONS]
         cdef int aint_count = 0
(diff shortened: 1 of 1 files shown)


skxray/core/accumulators/histogram.pyx
```diff
@@ -184,31 +184,32 @@ class Histogram:
         numtypeindex = {tp : i for i, tp in enumerate(numtypes)}
         ctypeindices = [numtypeindex.get(x.dtype, 999) for x in coords]
         coordsorder = np.argsort(ctypeindices, kind='mergesort')
-        istrides = np.asarray(self._values.strides, dtype=np.int32)[coordsorder]
+        istrides = np.asarray(self._values.strides)[coordsorder]
         istrides //= self._values.itemsize
-        cdef int [:] dataindexstrides = istrides
+        cdef int dataindexstrides[MAX_DIMENSIONS]
+        cdef int i
+        for i in range(self.ndims):
```
ericdill: is this `i` index the one that is being defined above on line `190`?  I have not looked at the C code to see if it is actually using this variable.

pavoljuhas: Yes, it is used in the c file at the loop on line 191, 192.

ericdill:
Looks good to me. Thanks!

PR #385, opened 2015-12-28
TST: Clean up tests

ericdill:


ericdill:
Leaving open for one more day just in case anyone has comments. Will self-merge tomorrow (1/5/16) otherwise

PR #389, opened 2016-01-05
ENH: Automagically build the docs on travis

ericdill:


ericdill:
This will automagically build the docs on travis once it is merged.  It will only build the docs on python 3.5 after travis is triggered with a new commit in *master*.  It does not build the docs on pull requests.

ericdill:
Here is an [example of the built docs] (http://scikit-beam.github.io/scikit-beam/) for my [scikit-beam:auto-build-docs-on-travis branch] (https://github.com/scikit-beam/scikit-beam/tree/auto-build-docs-on-travis) and [the travis job] (https://travis-ci.org/scikit-beam/scikit-beam/jobs/100410815) that built those specific docs

danielballan:
Sweet, that was fast! I can't see any problems with this. It is IMO cleaner than any approaches I have seen before. Good to merge?

ericdill:
Yeah turned out to be somewhat easier than I anticipated.  The **only** reason not to merge now would be that it is my personal OAuth token that is being used to push to scikit-beam.  I can't think of any reason why this matters, but if you know of a reason why we should do it differently (a la create a new github account for the express purposes of pushing to github from travis) then I can do that too!  Thoughts?

danielballan:
You can create a token (not a separate account) with limited permission. Want to give that a try?

Sent from Outlook Mobile<https://aka.ms/qtex0l>




On Wed, Jan 6, 2016 at 6:42 AM -0800, "Eric Dill" <notifications@github.com<mailto:notifications@github.com>> wrote:


Yeah turned out to be somewhat easier than I anticipated. The only reason not to merge now would be that it is my personal OAuth token that is being used to push to scikit-beam. I can't think of any reason why this matters, but if you know of a reason why we should do it differently (a la create a new github account for the express purposes of pushing to github from travis) then I can do that too! Thoughts?

ericdill:
The token that I am using already has the minimal permissions set. My point was more along the lines that this is a token from my account, not a token from a shared account that we all would have access to

ericdill:
Admittedly I am unaware of any sort of convention in the open source community regarding whose account is used to upload packages to pypi, whose account has push access from travis to github, etc... I am happy to continue using my account to do the pushing to github from travis part, but if there is a convention that I am unaware of, let's just follow that instead!

danielballan:
Oh I see. This is how it is done AFAIK. We all have push access to replace the token as needed.

Sent from Outlook Mobile<https://aka.ms/qtex0l>




On Wed, Jan 6, 2016 at 7:58 AM -0800, "Eric Dill" <notifications@github.com<mailto:notifications@github.com>> wrote:


Admittedly I am unaware of any sort of convention in the open source community regarding whose account is used to upload packages to pypi, whose account has push access from travis to github, etc... I am happy to continue using my account to do the pushing to github from travis part, but if there is a convention that I am unaware of, let's just follow that instead!

ericdill:
Oh ok great.  Then this is good to go! Thanks for having a look :)

danielballan:
Sorry for the delay, @ericdill. :fire: at BNL.

ericdill:
No worries. Thanks for getting this in!

PR #390, opened 2016-01-07
ENH: Programmatically convert all notebooks

ericdill:
~~When complete, this PR is going to run all notebooks in scikit-beam-examples and then convert them into html to incorporate into the docs.~~

I am punting on running the notebooks for now.  That is significantly more challenging.  However, this PR does add the very useful feature that we do not have to maintain the list of tutorials in `tutorial.rst` or `index.rst` any longer.  So that's cool.


doc/tutorial/Makefile
```diff
@@ -1,8 +1,4 @@
...
-
-	ipython nbconvert --to rst ../../../scikit-beam-examples/demos/1_time_correlation/*.ipynb
-	ipython nbconvert --to rst ../../../scikit-beam-examples/demos/dpc/*.ipynb
-	ipython nbconvert --to rst ../../../scikit-beam-examples/demos/reciprocal_space/*.ipynb
-	ipython nbconvert --to rst ../../../scikit-beam-examples/demos/speckle/*.ipynb
-	ipython nbconvert --to rst ../../../scikit-beam-examples/demos/xrf/*.ipynb
-	ipython nbconvert --to rst ../../../scikit-beam-examples/demos/*.ipynb
+	git clone https://github.com/scikit-beam/scikit-beam-examples /tmp/scikit-beam-examples
+	python write_main_index.py
+	ipython nbconvert --config convert_config.py --to rst
```
licode: interesting. so this line will convert all the ipython notebook according to configuration file convert_config.py.

ericdill: Yup. I was pretty excited to discover this configuration

doc/tutorial/write_main_index.py
```diff
@@ -0,0 +1,38 @@
...
+            if line.endswith('\n'):
+                line = line[:-1]
+            # find the line where I want to write the list of examples
+            if 'REPLACE_WITH_LIST_OF_EXAMPLES' in line:
+                line = ['   tutorial/%s' % os.path.splitext(notebook_path)[0]
+                        for notebook_path in notebooks]
+                new_file.extend(line)
+            else:
+                new_file.append(line)
+
```
licode: replace line 32-34 with one line
```
new_file.append(line)
```

licode:
looks good to me. Smarter way to organize tutorial examples.

doc/tutorial/write_main_index.py
ericdill: No, it is correct as is currently written.  On line 32, the variable `line` is a list.  By calling `new_file.extend(line)`, I am adding all the elements in the list `line` to the other list, keeping `new_file` as a flat list.  If I just do `new_file.append`, then I will end up with a list of lists.  That being said, I see how this code looks confusing.  I will update the variable names

licode:
```
 for line in f.readlines():
        if line.endswith('\n'):
            line = line[:-1]
        # find the line where I want to write the list of examples
        if 'REPLACE_WITH_LIST_OF_EXAMPLES' in line:
            line = ['   tutorial/%s' % os.path.splitext(notebook_path)[0]
                    for notebook_path in notebooks]

        new_file.append(line)
```

ericdill:
@licode Does that clarify what the inner loop of `write_main_index.py` is doing?

doc/tutorial/write_main_index.py
```diff
@@ -0,0 +1,37 @@
...
+# read in the template
+files_to_convert = {'../index.rst.tmpl': '../index.rst',
+                    '../tutorial.rst.tmpl': '../tutorial.rst'}
+for template_file, output_file in files_to_convert.items():
+    with open(template_file, 'r') as f:
+        new_file = []
+        for single_line in f.readlines():
+            single_line = single_line[:-1]
+            # find the line where I want to write the list of examples
+            if 'REPLACE_WITH_LIST_OF_EXAMPLES' in line:
```
licode: line or single_line?

ericdill: ah good catch

PR #391, opened 2016-01-08
WIP: Two time

sameera2004:


skxray/core/correlation.py
```diff
@@ -198,8 +197,8 @@ def _init_state(num_levels, num_bufs, labels):
     )
 
 
-def lazy_multi_tau(image_iterable, num_levels, num_bufs, labels,
-                   internal_state=None):
+def lazy_one_time(image_iterable, num_levels, num_bufs, labels,
```
sameera2004: @ericdill I have to change this name because we are giving multi tau capabilities for both one and two. Making it multi tau is upto the user, with using num_levels and num_bufs. If they don't want multi tau(brute force) they can choose num_levels=1 and num_bufs=number of images(or how many images they want to use for correlation). 
 However, this function is for one time , therefore I made this name change I will make changes in the notebook too.

skxray/core/correlation.py
```diff
@@ -405,3 +404,338 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
     return beta * np.exp(-2 * relaxation_rate * lags) + baseline
 
 
+two_time_results = namedtuple(
```
ericdill: I think that the `results` object can be shared between the one-time and two-time implementations

skxray/core/correlation.py
ericdill: Ah ok, that makes sense.  Can you basically add what you said about not wanting multi tau to the docstring of this function? That seems like very useful information.  Maybe add it at the top.  Something like:

```
"""Generator implementation of 1-time multi-tau correlation

If you do not want multi-tau correlation, set num_levels to 1 and
num_bufs to the number of images you wish to correlate
"""
```

skxray/core/correlation.py
```diff
@@ -405,3 +404,338 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+    'two_timecorrelation_results',
+    ['two_time', 'two_time_internal_state']
+)
+
+_two_time_internal_state = namedtuple(
```
ericdill: Is there any reason why the internal state object also cannot be shared between one-time and two-time?

skxray/core/correlation.py
sameera2004: Sure, I will add this doctring about multi_tau

skxray/core/correlation.py
sameera2004: Yes still I am looking for it, sure I will try to make it, I thing we can

skbeam/core/correlation.py
```diff
@@ -405,3 +406,357 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
     return beta * np.exp(-2 * relaxation_rate * lags) + baseline
 
 
+def two_time_corr(labels, images, num_frames, num_bufs, num_levels=1):
+    """Wraps generator implementation of multi-tau two time correlation
+
+    This function computes two-time correlations.
+    Original code : @author: Yugang Zhang
```
ericdill: I am not aware of `@author` meaning anything special.  Can you just have this say `Original code written by Yugang Zhang`

skbeam/core/correlation.py
```diff
@@ -336,18 +334,21 @@ def lazy_multi_tau(image_iterable, num_levels, num_bufs, labels,
...
-        yield results(g2, lag_steps[:g_max], s)
+        yield results(g2, s.lag_steps[:g_max], s)
 
 
 def multi_tau_auto_corr(num_levels, num_bufs, labels, images):
     """Wraps generator implementation of multi-tau
 
-    See docstring for lazy_multi_tau
+    Original code(in Yorick) for multi tau auto correlation
+    @author: Mark Sutton
```
ericdill: I am not aware of `@author` meaning anything special.  Can you just have this say `Original code written by Mark Sutton`

skbeam/core/correlation.py
```diff
@@ -190,18 +191,22 @@ def _init_state(num_levels, num_bufs, labels):
...
         future_intensity,
         img_per_level,
-        label_mask,
+        label_array,
         track_level,
         cur,
         pixel_list,
-        label_mapping,
+        num_pixels,
+        lag_steps,
```
ericdill: why did you add `num_pixels` and `lag_steps`?

skbeam/core/correlation.py
```diff
@@ -121,7 +119,7 @@ def _one_time_process(buf, G, past_intensity_norm, future_intensity_norm,
 
 results = namedtuple(
     'correlation_results',
-    ['g2', 'lag_steps', 'internal_state']
+    ['correlation_results', 'lag_steps', 'internal_state']
```
ericdill: Why was this renamed from `g2` to `correlation_results`?

skbeam/core/correlation.py
```diff
@@ -405,3 +406,357 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+        to track processing each level
+    cur : array
+
+    """
+    if num_bufs % 2 != 0:
+        raise ValueError("There must be an even number of `num_bufs`. You "
+                         "provided %s" % num_bufs)
+    label_array, pixel_list = extract_label_indices(labels)
+
+    # map the indices onto a sequential list of integers starting at 1
```
ericdill: Why are these being mapped onto a list starting at 1 instead of 0?

ericdill: Err, sorry, this is supposed to be on the line below this comment.

To elaborate a little... It seems that by mapping these onto a list starting at 1 (instead of at 0), it means that every time you do a `np.bincount`, you have to add a `[1:]` to the end of the resulting array.

ericdill:
I'll give this a closer read tomorrow.  Time to enjoy the second half of the Redskins/Packers game!

skbeam/core/correlation.py
sameera2004: I renamed result object from g2 to correlation_results to use for both one_time and two_time

skbeam/core/correlation.py
sameera2004: Sure I will take out @author from both places in correlation.py

skbeam/core/correlation.py
sameera2004: I am trying to stop repeating things use for both 1 and 2 correlation. therefore, I put num_pixels and lag_steps inside _input function and added it here.

sameera2004:
@ericdill Sure Thanks today I will put a test for this function

skbeam/core/correlation.py
ericdill: is `g2` a one-time correlation thing only?  Is there a conventional name for the results from two time correlation?

skbeam/core/correlation.py
```diff
@@ -405,3 +407,355 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+        pixel_list,
+        num_pixels,
+        lag_steps,
+        two_time,
+        count_level,
+        #current_img_time,
+    )
+
+
+def _inputs(num_bufs, num_levels, labels):
```
ericdill: can you rename this to `_validate_and_transform_inputs` or something equally verbose

skbeam/core/correlation.py
sameera2004: Usually `g2` is for one-time, but somie papers they call two-time `g2`. I don't find the conventional name for the results from two-time correlation. Anyway, Tomorrow I will discuss this with Andrei and Yugang.

skbeam/core/correlation.py
sameera2004: Sure I will rename the `_inputs` function

skbeam/core/correlation.py
ericdill: Ah ok.  My primary complaint with calling this `correlation_results` is that it is just pretty long.  That being said, we should just call this whatever Andrei/Yugang want it to be called.  When you meet with them, can you also verify that the output is what they want to see?

skbeam/core/correlation.py
ericdill: Ah ok, fair enough. Thanks for clarifying

skbeam/core/tests/test_correlation.py
```diff
@@ -102,7 +115,20 @@ def test_lazy_multi_tau():
     for second_half_result in gen_second_half:
         pass
 
-    assert np.all(full_result.g2 == second_half_result.g2)
+    assert np.all(full_result.correlation_results ==
+                  second_half_result.correlation_results)
+
+
+def test_two_time_corr():
+    y = []
```
ericdill: call `setup()` as the first line here. I am not sure how nosetests determines the order in which tests are run.  If this test is ever run first, there will be cryptic errors.

skbeam/core/tests/test_correlation.py
```diff
@@ -102,7 +115,20 @@ def test_lazy_multi_tau():
...
+    for i in range(50):
+        y.append(img_stack[0])
+    two_time, lag_steps = two_time_corr(rois, np.asarray(y),
+                                             50, num_bufs=50,
+                                             num_levels=1)
+
+    assert (np.all(two_time)==True)
```
ericdill: this should be `assert np.all(two_time)`. You don't need the `==True`, or the parentheses around the whole statement

ericdill:
Looks good to me. Let's chat after your meeting with Andrei and Yugang tomorrow.  Depending on how that meeting goes, I could see this being merged tomorrow.  Thanks for the effort @sameera2004 !

sameera2004:
@ericdill  Thank you very much. Sure I will let you know, and Thank for making the one-time correlation function more efficient.

skbeam/core/correlation.py
```diff
@@ -405,3 +408,359 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+    # generate a time frame for each level
+    time_ind = {key: [] for key in range(num_levels)}
+
+    for img in images:
+        s.cur[0] = (1 + s.cur[0]) % num_bufs  # increment buffer
+
+        s.count_level[0] = 1 + s.count_level[0]
+
+        # get the current image time
+        s = s._replace(current_img_time=(s.current_img_time + 1))
```
ericdill: This is creating a new namedtuple for every iteration.  If it is critical that `current_img_time` is tracked, make it a single element list, then you can just do `s.current_img_time[0] += 1`

Why is it important that we track `current_img_time`?

sameera2004: In two time you need to track the current_img_time because it will use to find the t1 and t2 values for the two time correlation(two times). The final results store in the results two_time(t1, t2)

Check from line 636  ..., It shows how it use.

tacaswell: I prefer the way @sameera2004 is doing it, it fits the functional model better and creating tuples should be very cheap.

sameera2004:
@ericdill This PR is ready for review, looks fine, 

Yesterday I discussed with Andrei and Yugang, they told we can keep both one time and two time,
`results = namedtuple(
    'correlation_results',
    ['g2', 'lag_steps', 'internal_state'])'

Because both one-time and two-time they are finding `g2' but different normalization, in some papers they call two time it C(12) or g12, but keeping g2 is fine.

skbeam/core/correlation.py
```diff
@@ -405,3 +408,358 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+                level += 1
+
+                # Checking whether there is next level for processing
+                processing = level < num_levels
+
+    for q in range(np.max(s.label_array)):
+        x0 = (s.two_time)[:, :, q]
+        (s.two_time)[:, :, q] = (np.tril(x0) + np.tril(x0).T -
+                                 np.diag(np.diag(x0)))
+    yield results(s.two_time, s.lag_steps, s)
```
ericdill: This is a different pattern than `lazy_one_time`.  `lazy_one_time` yields after every image. Can you make this function just yield the internal state, `s`, after every image and then provide a helper function to transform the internal state into the results object?

skbeam/core/tests/test_correlation.py
```diff
@@ -102,7 +115,20 @@ def test_lazy_multi_tau():
...
+
+def test_two_time_corr():
+    setup()
+    y = []
+    for i in range(50):
+        y.append(img_stack[0])
+    two_time, lag_steps = two_time_corr(rois,
+                                        np.asarray(y), 50,
+                                        num_bufs=50, num_levels=1)
+    assert np.all(two_time)
```
ericdill: I would like this test to be more like the lazy_one_time test, where the first half of the images are correlated, then the second half are correlated with the internal state from the first half. Then the results of the second half correlation can be compared with the results of running the correlation with the full stack.  That is one effective way of testing to make sure we are capturing the correct amount of state from the lazy_two_time function

skbeam/core/correlation.py
ericdill: Ok, thanks for weighing in here.  I'll defer to you on this one.

skbeam/core/tests/test_correlation.py
sameera2004: Added another test to take care of 1st half images correlated and then do the second half of images with the 1st half internal state and the compare with full set of images.

sameera2004:
@ericdill fixed those issues ready for review

sameera2004:
I have changed the tests again to show multi tau two time and ready for merge!

Here you can find the notebook created using real data

http://nbviewer.ipython.org/github/sameera2004/my_ipy_notebook/blob/master/Time_Correlation/XPCS%20Two-time%20SAXS%20pipeline_new_version.ipynb

sameera2004:
@ericdill anymore comments?, Can you please merge this?

ericdill:
Thanks for putting this notebook up. It is nice to see this code being used!

However, there is a lot of code in this notebook and it does not seem that you are using any of the tools that you are writing and have written for CHX.  I have a number of questions about this notebook.

1. Why is the `LazyEigerHandler` code in this notebook? Can't you just import from [eiger-io] (https://github.com/NSLS-II-CHX/eiger-io)?
2. Why is the code for `show_label_array_on_image` in this notebook? Can't you use `xray_vision.mpl_plotting.roi:show_label_array_on_image`?
3. Why is **all** of the correlation code in this notebook? Why are you not using the code in scikit-beam that you are writing?

skbeam/core/correlation.py
```diff
@@ -405,3 +408,358 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+    namedtuple
+        A `results` object is yielded after every image has been processed. This
+        `reults` object contains:
+        - the normalized correlation, `two_time`
+        - the times at which the correlation was computed, `lag_steps`
+        - and all of the internal state, `final_state`, which is a
+          `correlation_state` namedtuple
+
+    Returns
+    -------
```
ericdill: This generator does not `return` and `yield`

sameera2004:
In the CHX computers, still we have skxary 

1. ` LazyEigerHandler` may be this I can use eiger-io, but it was not install last Dec, after that I didn't try
2. 'Show_label-array_on_image` also new in `X-ray-vision` to my knowledge that version of `X-ray-vision` is not install
3. The correlation code is same,

I logging to jupyterhub and run this code, all the chi-pipelines has to modify(and cleaned )after installing new `skbeam`, this notebook is only en example no plan to add it chi-pipelines before cleaning

skbeam/core/correlation.py
sameera2004: I took the `return` out

sameera2004:
@ericdill I fixed `lazy_two_time` documentation,

 I will update the new CHX-pipelines when the `skbeam` is installed.

ericdill:
But the point of making example notebooks is to use the code in scikit-beam, not to just copy it out and use it somewhere else.  Why can't you use the code in your branch to run the chx pipeline?

ericdill:
Oh, I didn't see your other comment, sorry.  You should be able to create your own kernel for jupyterhub and use whatever environment you need.  If you need help installing new libraries to use with jupyterhub, let's get together and I can help you figure that out.

skbeam/core/correlation.py
```diff
@@ -405,3 +409,356 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+                s.cur[level] = 1 + s.cur[level] % num_bufs
+                s.count_level[level] = 1 + s.count_level[level]
+
+                s.buf[level, s.cur[level] - 1] = (s.buf[level - 1, prev - 1] +
+                                                  s.buf[level - 1,
+                                                  s.cur[level - 1] - 1])/2
+
+                t1_idx = (s.count_level[level] - 1) * 2
+
+                current_img_time = ((s.time_ind[level - 1])[t1_idx]
```
ericdill: Is this `current_img_time` different from `s.current_img_time`?

skbeam/core/tests/test_correlation.py
```diff
@@ -88,13 +101,43 @@ def test_lazy_multi_tau():
     assert np.average(full_result.g2-1) < 0.01
 
     # run the correlation on the first half
-    gen_first_half = lazy_multi_tau(
+    gen_first_half = lazy_two_time(rois, img_stack[:stack_size//2], stack_size,
+                                   num_bufs=stack_size, num_levels=1)
```
ericdill: Can `num_frames` be longer than the length of the image stack?  Here `len(img_stack)` is only half of `stack_size`

skbeam/core/correlation.py
```diff
@@ -405,3 +409,356 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+def _init_state_two_time(num_levels, num_bufs, labels, num_frames):
+    """Initialize a stateful namedtuple for two time correlation
+
+    Parameters
+    ----------
+    num_levels : int
+    num_bufs : int
+    labels : array
+        Two dimensional labeled array that contains ROI information
+
```
ericdill: You are missing the docstring for the `num_frames` parameter

skbeam/core/correlation.py
```diff
@@ -405,3 +409,356 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+    Parameters
+    ----------
+    labels : array
+        labeled array of the same shape as the image stack;
+        each ROI is represented by a distinct label (i.e., integer)
+    images : array
+        dimensions are: (rr, cc), iterable of 2D arrays
+    num_frames : int
+        number of images to use
+        default is number of images
```
ericdill: `num_frames` has no default value in this code path.  Is it actually an optional parameter? If so, you should add some code somewhere that looks like 

```python
if not num_frames:
    num_frames = len(images)
```

but that will not work if `images` is a generator, so maybe this parameter is not actually optional?  I'm not sure

skbeam/core/correlation.py
```diff
@@ -405,3 +409,356 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+    num_bufs : int, must be even
+        maximum lag step to compute in each generation of
+        downsampling
+        default is number of images
```
ericdill: Same with this parameter. It is not optional, so noting that there is a default does not make sense.  If it is optional, make the code reflect this. Otherwise remove this line

skbeam/core/tests/test_correlation.py
sameera2004: YEs `num_frames` can be longer then you will get like following , not enough images to fill,

screen shot 2016-01-14 at 10 18 57 am

skbeam/core/correlation.py
```diff
@@ -405,3 +409,356 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+
+    The longest lag time computed is num_levels * num_bufs.
+    ** see comments on multi_tau_auto_corr
+
+    Parameters
+    ----------
+    labels : array
+        labeled array of the same shape as the image stack;
+        each ROI is represented by a distinct label (i.e., integer)
+    images : array
```
ericdill: Here you say that this parameter is an `array`, but in the explanation of the parameter you say it is an `iterable of 2D arrays`. If it can be an iterable, then this type should be `iterable`, not `array`

skbeam/core/correlation.py
sameera2004: Yes, you don't have to track it, this will only use in multi_tau method, current_img_time will be calculated here, it doesn't need prior values for it.

skbeam/core/correlation.py
```diff
@@ -405,3 +409,356 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+                _two_time_process(s.buf, s.two_time, s.label_array, num_bufs,
+                                  s.num_pixels, s.img_per_level, s.lag_steps,
+                                  current_img_time, level=level,
+                                  buf_no=s.cur[level]-1)
+                level += 1
+
+                # Checking whether there is next level for processing
+                processing = level < num_levels
+
+    for q in range(np.max(s.label_array)):
```
ericdill: This generator implementation is different from `lazy_one_time`.  You should `yield s` after **every** image and then provide a helper function to convert `s` into `g2 and `lag_steps`.  Lines 557-568 should be converted into something like this:

```python
                level += 1

                # Checking whether there is next level for processing
                processing = level < num_levels
        yield s

def two_time_state_to_results(state):
    """Convert the internal state of the two time generator into usable results

    Parameters
    ----------
    state : namedtuple
        The internal state that is yielded from `lazy_two_time`

    Returns
    -------
    results : namedtuple
        A results object that contains the two time correlation results and the
        lag steps
    """
    for q in range(np.max(state.label_array)):
        x0 = (state.two_time)[:, :, q]
        (state.two_time)[:, :, q] = (np.tril(x0) + np.tril(x0).T -
                                     np.diag(np.diag(x0)))
    return results(state.two_time, state.lag_steps, state)
```

You will likely need to update the tests accordingly.  

It would also make sense to remove `state` from the `results` namedtuple, but that would then require making similar changes to the `lazy_one_time` generator and all code paths, so you are welcome to punt on doing that in this PR and I can do it later.

skbeam/core/tests/test_correlation.py
ericdill: Ok great, thanks for clarifying

skbeam/core/correlation.py
ericdill: Ok great, thanks for clarifying

skbeam/core/tests/test_correlation.py
tacaswell: Given how sparse that array is I can not help but think there are better ways to store it

ericdill: Probably. Is there an array size (i.e., some rr x cc) where it really starts to matter? I wonder if we are even making arrays that are big enough that we care

skbeam/core/correlation.py
sameera2004: I added this function and updated the tests.
 I will remove the `state` from the name tuple, and `lazy_one_time` too, any way I will add more changes to both one and two time correlation functions while I am working on removing bad_images. 

@ericdill can you please review this and merge this PR

ericdill:
Ugh sorry I was commenting on a single commit, not the overall diff -_-. Let me delete these comments and try again...

skbeam/core/correlation.py
```diff
@@ -405,3 +409,364 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+    This function computes two-time correlation
+    Original code : author: Yugang Zhang
+
+    See docstring for lazy_two_time
+    """
+    gen = lazy_two_time(labels, images, num_frames, num_bufs, num_levels)
+    for result in gen:
+        pass
+    final_result = two_time_state_to_results(result)
+    return final_result.g2, final_result.lag_steps
```
ericdill: Why not just return `two_time_state_to_results(result)`?

skbeam/core/correlation.py
```diff
@@ -405,3 +409,364 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+        maximum lag step to compute in each generation of
+        downsampling
+    num_levels : int, optional
+        how many generations of downsampling to perform, i.e.,
+        the depth of the binomial tree of averaged frames
+        default is one
+
+    Yields
+    ------
+    internal_state - tuple
```
ericdill: Change the `-` to a `:`

skbeam/core/correlation.py
```diff
@@ -405,3 +409,364 @@ def auto_corr_scat_factor(lags, beta, relaxation_rate, baseline=1):
...
+
+def _init_state_two_time(num_levels, num_bufs, labels, num_frames):
+    """Initialize a stateful namedtuple for two time correlation
+
+    Parameters
+    ----------
+    num_levels : int
+    num_bufs : int
+    labels : array
+        Two dimensional labeled array that contains ROI information\
```
ericdill: Delete the `\` at the end of this line

skbeam/core/tests/test_correlation.py
```diff
@@ -64,22 +66,66 @@ def setup():
...
-    assert np.all(g2 == gen_result.g2)
-    assert np.all(lag_steps == gen_result.lag_steps)
+    assert np.all(g2 == gen_result_one.g2)
+    assert np.all(lag_steps == gen_result_one.lag_steps)
+
+    full_gen_two = lazy_two_time(rois, img_stack, stack_size,
+                                 num_bufs, num_levels)
+    for gen_result_two in full_gen_two:
+        pass
+    final_gen_result_two = two_time_state_to_results(gen_result_two)
```
ericdill: To make these variables clearer, rename `gen_result_two` to `gen_state_two` and `final_gen_result_two` to `gen_result_two`

skbeam/core/tests/test_correlation.py
```diff
@@ -64,22 +66,66 @@ def setup():
 def test_lazy_vs_original():
     setup()
     # run the correlation on the full stack
-    full_gen = lazy_multi_tau(
+    full_gen_one = lazy_one_time(
         img_stack, num_levels, num_bufs, rois)
-    for gen_result in full_gen:
+    for gen_result_one in full_gen_one:
```
ericdill: This is not a result, rename to `gen_state_one`

ericdill: Yes it is, my mistake

skbeam/core/tests/test_correlation.py
```diff
@@ -64,22 +66,66 @@ def setup():
...
 
-def test_lazy_multi_tau():
+def test_lazy_two_time():
     setup()
     # run the correlation on the full stack
-    full_gen = lazy_multi_tau(
-        img_stack, num_levels, num_bufs, rois)
+    full_gen = lazy_two_time(rois, img_stack, stack_size,
+                             stack_size, 1)
+    for full_result in full_gen:
```
ericdill: This is not a result, rename to `full_state`

sameera2004:
@ericdill I fixed it( changed from result to state, took out #, and changed the `-` to `:' in the documentation

ericdill:
Ok, we are almost there. Thanks for your patience @sameera2004 .  I am looking at the coverage stats. You are missing a test that raises the value error on line 740. Can you make sure to exercise that code path?  The line in question can be found [here] (https://github.com/scikit-beam/scikit-beam/pull/391/files#diff-6044560e5d90cac1c5ed37302acc83b4R740)

sameera2004:
I added test to check the num_buf is even or not. As I remember that test was there in the old one_time tests!

ericdill:
Other than that last comment, I think that this looks good.  The last thing that I would like you to do is to use this code on actual data and make sure that you can resume the two time correlation with the internal_state object.  Do you know what I mean?

sameera2004:
@ericdill I took out the lambda from the tests.  I was able to use real data and resume the two time correlation with internal_state object.

There is a example notebook , in skbeam_examples ( I made a PR for that Two time correlation example notebook #46)

http://nbviewer.ipython.org/github/sameera2004/scikit-beam-examples/blob/two_time_corr/demos/time_correlation/two-time-correlation.ipynb

ericdill:
Fantastic. Thank you for verifying that for me!

ericdill:
Thanks for all your hard work on this @sameera2004 and @yugangzhang. Looks good to me!

ericdill:
@sameera2004 Can you submit that last notebook as a PR against scikit-beam-examples?

sameera2004:
@ericdill Thanks!  there is a PR against scikit-beam-examples,
https://github.com/scikit-beam/scikit-beam-examples/pull/46

PR #392, opened 2016-01-09
ENH: Convert DPC code into a generator that operates on partial results

ericdill:
Still need to make an example notebook with this code.

- [x] Working notebook example with streaming DPC data scikit-beam/scikit-beam-examples#45

### DIFF
--- a/skbeam/core/dpc.py
+++ b/skbeam/core/dpc.py
@@ -280,7 +280,7 @@ def recon(gx, gy, scan_xstep, scan_ystep, padding=0, weighting=0.5):
 
 
 def dpc_runner(ref, image_sequence, start_point, pixel_size, focus_to_det,
-               scan_rows, scan_cols, scan_xstep, scan_ystep, energy,
+               scan_rows, scan_cols, scan_xstep, scan_ystep,
                solver='Nelder-Mead', roi=None, bad_pixels=None,
                dpc_state=None):
     """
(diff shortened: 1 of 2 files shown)


ericdill:
@tacaswell Could you have a look at this since you were intimately involved in the curation of the original DPC code from Cheng?  Other than some code formatting, I think that this is ready to go.

skbeam/core/dpc.py
```diff
@@ -366,52 +357,45 @@ def dpc_runner(ref, image_sequence, start_point, pixel_size, focus_to_det,
...
+        The internal state that `dpc_runner` requires for each iteration.
+        Can be passed to reconstruct_phase_from_partial_info which, along
+        with some additional info, will produce the final phase image
 
     References: text [1]_
     .. [1] Yan, H. et al. Quantitative x-ray phase imaging at the nanoscale by
     multilayer Laue lenses. Sci. Rep. 3, 1307; DOI:10.1038/srep01307 (2013).
 
     """
+    def initialize_state(scan_rows, scan_cols, ref, roi, bad_pixels):
```
tacaswell: should this function be public?

ericdill: I don't know.  I had the same question.  The reason why I have it internal to lazy_dpc is because I can't think of why you would want to initialize the state yourself.  Seems like the only use would be to pass it in to the lazy_dpc function.

tacaswell: One reason maybe for distributed jobs where you may want to send the data to the remote machine using some random serialization and you will need to reconstruct the state before resuming the computation.

ericdill: Ah, that is a fair point.  In the interest of not over-architecting, it sounds like we should just cross that bridge when the time comes?

ericdill:
@tacaswell bump. This look ok?

PR #395, opened 2016-01-13
WIP: added new function to get one_time_correlation data from two_tim…

sameera2004:
…e_corrrelation

### DIFF
--- a/skbeam/core/correlation.py
+++ b/skbeam/core/correlation.py
@@ -599,7 +599,7 @@ def _two_time_process(buf, g2, label_array, num_bufs, num_pixels,
         image data array to use for two time correlation
     g2: array
         two time correlation matrix
-        shape (number of labels(ROI), number of images, number of images)
+        shape (number of labels(ROI), number of frames, number of frames)
     label_array: array
         Elements not inside any ROI are zero; elements inside each
         ROI are 1, 2, 3, etc. corresponding to the order they are specified
@@ -806,13 +806,13 @@ def one_time_from_two_time(two_time_corr):
     ----------
     two_time_corr : array
         matrix of two time correlation
-        shape (number of labels(ROI's), number of images, number of images)
+        shape (number of labels(ROI's), number of frames, number of frames)
 
     Returns
     -------
     one_time_corr : array
         matrix of one time correlation
-        shape (number of labels(ROI), number of labels(ROI's))
+        shape (number of labels(ROI's), number of frames)
     """
 
     one_time_corr = np.zeros((two_time_corr.shape[0], two_time_corr.shape[2]))
--- a/skbeam/core/correlation.py
+++ b/skbeam/core/correlation.py
@@ -794,3 +794,29 @@ def _validate_and_transform_inputs(num_bufs, num_levels, labels):
     return (label_array, pixel_list, num_rois, num_pixels,
             lag_steps, buf, img_per_level, track_level, cur,
             norm, lev_len)
+
+
+def one_time_from_two_time(two_time_corr):
+    """
+    This will provide the one-time correlation data from two-time correlation
+    data. 
+
+    Parameters
+    ----------
+    two_time_corr : array
+        matrix of two time correlation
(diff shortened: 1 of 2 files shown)


ericdill:
This now has merge conflicts. It also needs a test

sameera2004:
Sure, I will fix it and add a test to that

sameera2004:
This PR is ready for review, added tests

skbeam/core/correlation.py
```diff
@@ -769,3 +769,29 @@ def _validate_and_transform_inputs(num_bufs, num_levels, labels):
...
+def one_time_from_two_time(two_time_corr):
+    """
+    This will provide the one-time correlation data from two-time correlation
+    data. 
+
+    Parameters
+    ----------
+    two_time_corr : array
+        matrix of two time correlation
+        shape (number of images, number of images, number of labels(ROI))
```
ericdill: I think your shape is backwards. Shouldn't this be `(num ROIs, num images, num images)`?  Same below in the `Returns` section

ericdill: After in-person discussion with @sameera2004 , we have concluded that the two-time correlation results, that is `g2`, should have a shape of `(num ROI, num images, num images)` which will require touching the `_init_state_two_time()` function, and the `two_time_state_to_results()` function and probably the tests and the examples.... (sorry :cry: )

sameera2004:
@ericdill I changed the `two-time correlation results`, that is `g2`, to have a shape of `(num ROI, num images, num images)` and fixed the tests too. They are passing and this PR is ready for review.

skbeam/core/tests/test_correlation.py
```diff
@@ -44,9 +44,12 @@
                                      auto_corr_scat_factor,
                                      lazy_one_time,
                                      lazy_two_time, two_time_corr,
-                                     two_time_state_to_results)
+                                     two_time_state_to_results,
```
ericdill: This is imported twice

ericdill:
Did you consider having the one time from two time function return the same thing as `multi_tau_auto_corr`?

ericdill:
Err have it return the correlation and the lag steps?

ericdill:
Oh wait you already have that info from the two time results

sameera2004:
@ericdill No `one_time_from_two_time_function` will not return the same results as `multi_tau_auto_corr`. If you `brute force`(not `multi tau`) using `multi tau auto corr` the results will have same shape.

sameera2004:
@ericdill I fixed the PR ready to review

skbeam/core/tests/test_correlation.py
```diff
@@ -198,6 +200,28 @@ def test_bad_images():
...
+    y_dim = 10
+    stack = 10
+    imgs = np.random.randint(1, 3, (stack, x_dim, y_dim))
+    roi = np.zeros_like(imgs[0])
+    # make sure that the ROIs can be any integers greater than 1.
+    # They do not have to start at 1 and be continuous
+    roi[0:x_dim//10, 0:y_dim//10] = 5
+    roi[x_dim//10:x_dim//5, y_dim//10:y_dim//5] = 3
+
+    two_time = two_time_corr(roi, imgs, stack,
```
ericdill: It is a little difficult to understand exactly what `two_time[0]` means in this test, as you need to understand trace the results of `two_time_corr` through about 3 function calls to understand what exactly it returns.  Having just realized that it is difficult to figure out what `two_time[0]` is, I just put in #407 to update the docs for `two_time_corr`.  Could you also change `two_time` to be `g2, lag_steps, _state` to make the meaning of the return value of `two_time_corr` a little clearer in this test?

ericdill:
Hi, one last minor comment. Then it is good to go!

sameera2004:
@ericdill i fixed this PR ready for review. For your PR #407 I left a minor comment, other than it is good to go.

ericdill:
Awesome, thanks @sameera2004 !

sameera2004:
@ericdill Thank you very much

PR #396, opened 2016-01-14
DOC: Add two-time to whats-new

ericdill:


licode:
Looks good to me. This is a good way to handle updates in new release.

PR #397, opened 2016-01-14
DOC: Move whats-new up one level

ericdill:
This way it will automatically build without having to add a new file
for each release. Which I forgot to do this time. So I am making it more
eric-proof in the future


ericdill:
After this is merged, http://scikit-beam.github.io/scikit-beam/whats_new.html will always show whatever is in the `whats_new.rst` file.  This project is not active enough that having individual files for each release makes sense yet.  It might eventually, but once we get to **that** point, we can make individual files for each release.

sameera2004:
Looks good.

PR #398, opened 2016-01-15
MNT: Gotta make the output folder

ericdill:
Ugh it is so hard to get the docs to work, since travis only builds and 
pushes the docs if it is on the master branch.  This PR should build the docs
every time on python 3.5 and if the docs fail, then travis will fail.  It will
only push the docs if travis is not building a PR and is on the scikit-beam org.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,16 +36,21 @@ install:
 
 script:
   - python run_tests.py
-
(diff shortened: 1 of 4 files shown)


ericdill:
Self-merging because I really want to get the docs auto-building

PR #401, opened 2016-01-24
Removing bad images from xsvs analysis

sameera2004:


tacaswell:
How does the fair on corrolating `np.ones(10000, 50, 50)` with some bad frames thrown in?

sameera2004:
Here I am showing the a notebook example for `one time correlation`.
I was taking the same image by `repeating for 10000 times`. (starting from `60` cell in the notebook.)
Then we should get the `same values for the one time correlation`.
 Then I introduce bad frames different steps 10, 100, 200, 500, 1000, 2500, 5000. The results are reasonable there is no change until 1000 bad images but 2500 and 5000 you will lose last one time correlation results. http://nbviewer.jupyter.org/github/sameera2004/my_ipy_notebook/blob/master/Time_Correlation/Check_remove_bad_imgs_using_same_image.ipynb

sameera2004:
Ready for review

skbeam/core/speckle.py
```diff
@@ -237,20 +240,29 @@ def _process(num_roi, level, buf_no, buf, img_per_level, labels,
...
-        roi_data = buf[level, buf_no][labels == label]
-        spe_hist, bin_edges = np.histogram(roi_data, bins=bin_edges,
-                                           density=True)
-        spe_hist = np.nan_to_num(spe_hist)
-        prob_k[level, j] += (spe_hist -
-                             prob_k[level, j])/(img_per_level[level])
-
-        prob_k_pow[level, j] += (np.power(spe_hist, 2) -
-                                 prob_k_pow[level, j])/(img_per_level[level])
+    # check if there are any bad images, using bad_to_nan function
```
ericdill: Did you mean to use the `bad_to_nan` function somewhere in this code?

skbeam/core/speckle.py
```diff
@@ -237,20 +240,29 @@ def _process(num_roi, level, buf_no, buf, img_per_level, labels,
...
+    # in mask.py all the bad images are converted into np.nan arrays,
+    # therefore check for np.nan arrays
+    if np.isnan(buf[level, buf_no]).any():
+        track_bad[level] += 1
```
ericdill: I would change this to:

```python
if np.isnan....
    track_bad
    return

for j, label in enumerate(u_labels):
    ...
```

skbeam/core/speckle.py
sameera2004: bad_to_nan function is a separate function in mask.py. If there is bad images, (and defining bad is up to the user), user can use bad_to_nan function for images that will be used in XSVS or XPCS analysis. Therefore, if they used it for the images, this XSVS function has to check and remove that image and track it, if the user didn't use it(all the images are good) then the old xsvs analysis. No change

ericdill: Thank you for the clarification.  I misunderstood the intent of the code comment because I interpreted this code comment as "This code is going to check for bad images using the bad_to_nan function".  Would you change this code comment to simply say "Check if there are any bad images, represented as an array filled with np.nan"?  I think that would capture your intent.

sameera2004:
@ericdill, I fixed this PR it is ready for review

ericdill:
Can you note in the docstring of the `xsvs` function that bad images need to be represented as an array filled with `np.nan`?

ericdill:
Is there an XSVS notebook somewhere?

ericdill:
Couple of minor comments @sameera2004 . Looks good otherwise. Thanks

sameera2004:
There is a xsvs notebook in skbeam-examples but removing bad images part is not included. Added a doctoring to xsvs function about bad images.

@ericdill this PR is ready to review

ericdill:
I was unable to find the xsvs notebook in scikit-beam-examples. Which notebook is it?

ericdill:
Otherwise, looks good. Thanks!

sameera2004:
Thanks. @ericdill Here is the notebook https://github.com/scikit-beam/scikit-beam-examples/blob/master/demos/speckle/X-ray_Speckle_Visibility_Spectroscopy.ipynb

ericdill:
Oh, because it is under demos/speckle, not demos/time_correlation. Duh!  Thanks :smile:

PR #405, opened 2016-01-29
User peak

licode:
Implement user peak, so users have more flexible control of escape peak or pileup peaks.

### DIFF
--- a/skbeam/core/fitting/xrf_model.py
+++ b/skbeam/core/fitting/xrf_model.py
@@ -420,13 +420,19 @@ class ParamController(object):
                 if constraint is not None:
                     self._element_strategy[param_name] = constraint
                 self.params.update({param_name: new_pos})
-        else:
+        elif 'user' in element.lower():  # for user peak
+            param_name = '_'.join((element, param_suffix))
+            new_pos = PARAM_DEFAULTS[kind].copy()
+            if constraint is not None:
+                self._element_strategy[param_name] = constraint
+            # update parameter in place
+            self.params.update({param_name: new_pos})
+        else:   # for pileup peak
             linename = 'pileup_'+element.replace('-', '_')
             param_name = linename + param_suffix  # as in lmfit Model
             new_pos = PARAM_DEFAULTS[kind].copy()
             if constraint is not None:
                 self._element_strategy[param_name] = constraint
-
             # update parameter in place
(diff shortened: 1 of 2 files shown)


tacaswell:
Can you explain the purpose of this a bit more?

licode:
There are several reasons to add user peaks:
- User-peaks can be used as a quick replacement for a pileup peak. Pileup peaks are generated due to detector inefficiency. If users are confident that certain peak is not an element peak, but from detector defects, users can simply add user-peaks instead of spending long time working on pileup peaks.
- User-peaks can be used to represent escape peaks. 
- User-peaks can be used to represent other unknown peaks due to detector defects or scattering process.  
- This feature has been required by scientists at HXN for XRF fitting.

licode:
anyone check on this? @ericdill @tacaswell ?

ericdill:
Sure thing. I'll have a look shortly @licode

skbeam/core/fitting/xrf_model.py
```diff
@@ -420,13 +420,19 @@ def add_param(self, kind, element, constraint=None):
                 if constraint is not None:
                     self._element_strategy[param_name] = constraint
                 self.params.update({param_name: new_pos})
-        else:
+        elif 'user' in element.lower():  # for user peak
```
ericdill: Can you add a note about this in the docstring for this function?  It is not immediately evident from the current docstring that `add_param` supports the notion of a `user_peak`

ericdill:
Are you using this in PyXRF?  I'm curious to see how this is being used in the wild. Or is the plan to get it in to scikit-beam and then update PyXRF to use it?

ericdill:
Looks like you are missing coverage on all of the if blocks in the new `elif 'user' ...` block in xrf_model.py.  
The lines are 

    840, 851, 856, 862, 868

Can you make sure to hit those if blocks in the tests?

ericdill:
Only a couple of minor comments regarding test coverage and adding documentation to the `add_param` function. Otherwise, looks good.

licode:
Yes, this function was implemented at PyXRF already. Users can select the user peak from the scroll-down window the same as they choose element peaks.  I will add more coverage.

ericdill:
Great, thanks for updating the docs. Looks good to me.

PR #407, opened 2016-02-03
DOC: Clarify docstrings in correlation.py

ericdill:


skbeam/core/correlation.py
```diff
@@ -469,9 +480,14 @@ def lazy_two_time(labels, images, num_frames, num_bufs, num_levels=1,
...
     Yields
     ------
-    internal_state : tuple
-      all of the internal state, `final_state`, which is a
-      `correlation_state` namedtuple
+    namedtuple
+        A `results` object is yielded after every image has been processed.
+        This `reults` object contains, in this order:
+        - `g2`: the normalized correlation
+          shape is (len(lag_steps), len(lag_steps), num_rois)
```
sameera2004: when the PR #395 merged this returning g2 shape will be (num_rois, len(lag_steps), len(lag_steps))

ericdill:
@sameera2004 good catch on the order of the axes in the two time results. I've fixed it and pushed up a new commit. This should be good to go.

sameera2004:
@ericdill Thank you for making correlation.py docs strings more clear. Looks good!

PR #415, opened 2016-02-14
MNT: Invoke coverage with the 'coverage run`

ericdill:
Just listened to a podcast today with the author of the `coverage` package 
who recommends that you invoke coverage via `coverage run <test_script.py`.
This is also what is listed on the home page of the [coverage docs] 
(https://coverage.readthedocs.org/en/coverage-4.0.3/).  The podcast can 
be found [here] (https://coverage.readthedocs.org/en/coverage-4.0.3/)
 for those of you that are interested.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,5 @@
 [run]
-source=scikit-beam
+source=skbeam
 [report]
 omit =
     */python?.?/*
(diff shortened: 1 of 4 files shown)


ericdill:
This is a minor change that just makes coverage stats better. Self merging.

PR #422, opened 2016-02-24
Convert reciprocal space and 3d grinder to OpenMP

stuwilkins:
This request converts the c routines to use OpenMP (as opposed to pthreads) and uses scatter gather for the grinder.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,8 @@ def c_ext():
         return []
     # compile the extension on OSX and Linux.
     return [Extension('skbeam.ext.ctrans', ['src/ctrans.c'],
-                      define_macros=[('USE_THREADS', None)])]
+                      extra_compile_args=['-fopenmp'],
+                      extra_link_args=['-lgomp'])]
 
 
 def cython_ext():
--- a/skbeam/core/recip.py
+++ b/skbeam/core/recip.py
@@ -57,7 +57,7 @@ logger = logging.getLogger(__name__)
 
 def process_to_q(setting_angles, detector_size, pixel_size,
                  calibrated_center, dist_sample, wavelength, ub,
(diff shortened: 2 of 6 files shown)


stuwilkins:
Note : DO NOT MERGE YET! Submitting for discussion and comments.

stuwilkins:
Attn @ericdill, This now uses OpenMP and is ready for trying on windows.

ericdill:
Thanks for pinging me again. I saw this earlier and was waiting for it to pass tests. I suppose the most effective way for you to debug this on windows is going to be to get appveyor running CI for us on this project.  Do you have any idea what is required to get OpenMP going on Windows?

licode:
I think openMP is included in most of the standard compilers (gcc/g++) in windows, so there is no more installation needed. http://openmp.org/wp/openmp-compilers/ 
However, it seems that openMP is not associated with standard gcc compiler in mac (installed with Xcode automatically). So I usually manually install another gcc using macport or brew in order to use openMP. 
This may cause some installation issue on mac?

stuwilkins:
Thanks @licode this must be a cython problem as well when compiling extension modules no?

Should we put some compiler directives around the openmp lines so that we can compile without (and then use single core?)

licode:
Yes, that will be helpful. cython works fine, although it may also include openMP in the backend.

licode:
Just sent a PR to fix_ctrans. Please let me know how it goes.

stuwilkins:
I think this is now ready for review and merging. Once that is done, I can work on the Windows bit, but this version is very light so might compile out of the box.

Attn @tacaswell @ericdill @licode.

licode:
Thanks. I will take a look at this.

licode:
Looks good to me.

PR #429, opened 2016-03-16
add mask option

yugangzhang:
mask is required to make a correct normalization of circular average.
@ericdill or @sameera2004  please take a look.

### DIFF
--- a/skbeam/core/roi.py
+++ b/skbeam/core/roi.py
@@ -397,6 +397,53 @@ def mean_intensity(images, labeled_array, index=None):
         mean_intensity[n] = ndim.mean(img, labeled_array, index=index)
     return mean_intensity, index
 
+
+
+def bin_1D(x, y, nx=None, min_x=None, max_x=None):
+    """
+    Bin the values in y based on their x-coordinates
+
+    Parameters
+    ----------
+    x : array
+        position
+    y : array
+        intensity
+    nx : integer, optional
+        number of bins to use defaults to default bin value
+    min_x : float, optional
+        Left edge of first bin defaults to minimum value of x
+    max_x : float, optional
+        Right edge of last bin defaults to maximum value of x
+
+    Returns
+    -------
+    edges : array
+        edges of bins, length nx + 1
+
+    val : array
+        sum of values in each bin, length nx
+
+    count : array
+        The number of counts in each bin, length nx
+    """
+    # handle default values
+    if min_x is None:
+        min_x = np.min(x)
+    if max_x is None:
+        max_x = np.max(x)
+    if nx is None:
+        nx = int(max_x - min_x) 
+    # use a weighted histogram to get the bin sum
+    bins = np.linspace(start=min_x, stop=max_x, num=nx+1, endpoint=True)
+    val, _ = np.histogram(a=x, bins=bins, weights=y)
+    # use an un-weighted histogram to get the counts
+    count, _ = np.histogram(a=x, bins=bins)
+    # return the three arrays
+    return bins, val, count
+
+
+
 def circular_average(image, calibrated_center, threshold=0, nx=None,
                      pixel_size=(1, 1),  min_x=None, max_x=None, mask=None):
     """Circular average of the the image data
@@ -413,7 +460,7 @@ def circular_average(image, calibrated_center, threshold=0, nx=None,
         default is zero
     nx : int, optional
         number of bins in x
-        defaults is None, gives the pixel number of the max radius 
+        defaults is 100 bins
     pixel_size : tuple, optional
         The size of a pixel (in a real unit, like mm).
         argument order should be (pixel_height, pixel_width)
@@ -422,8 +469,6 @@ def circular_average(image, calibrated_center, threshold=0, nx=None,
         Left edge of first bin defaults to minimum value of x
     max_x : float, optional number of pixels
         Right edge of last bin defaults to maximum value of x
-    mask: array, bool type
-        same shape as image
     Returns
     -------
     bin_centers : array
@@ -432,29 +477,25 @@ def circular_average(image, calibrated_center, threshold=0, nx=None,
         Radial average of the image. shape is (nx, ).
     """
     radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size) 
-    ps =np.min( pixel_size )
-    if max_x is None:max_x = np.max(radial_val)/ps
-    if min_x is None:min_x = np.min(radial_val)/ps        
-    if nx is None:        
-        nx = int(max_x - min_x)        
-    if mask is not None:          
+    if mask is not None:  
         mask = np.array( mask, dtype = bool)
         binr = radial_val[mask]
         image_mask =     np.array( image )[mask]        
     else:        
         binr = np.ravel( radial_val ) 
-        image_mask = np.ravel(image)         
-    bin_edges, sums, counts = utils.bin_1D( binr,
+        image_mask = np.ravel(image) 
+    binr_ =   binr /(np.sqrt(pixel_size[1]*pixel_size[0] ))
+    bin_edges, sums, counts = bin_1D(      binr_,
                                            image_mask,
-                                           nx,
+                                           nx=nx,
                                            min_x=min_x,
-                                           max_x=max_x)   
-    
+                                           max_x=max_x) 
     th_mask = counts > threshold
     ring_averages = sums[th_mask] / counts[th_mask]
     bin_centers = utils.bin_edges_to_centers(bin_edges)[th_mask]
     return bin_centers, ring_averages 
 
+ 
 
 
 
--- a/skbeam/core/roi.py
+++ b/skbeam/core/roi.py
@@ -397,9 +397,8 @@ def mean_intensity(images, labeled_array, index=None):
         mean_intensity[n] = ndim.mean(img, labeled_array, index=index)
     return mean_intensity, index
 
-
-def circular_average(image, calibrated_center, threshold=0, nx=100,
-                     pixel_size=(1, 1),  min_x=None, max_x=None):
+def circular_average(image, calibrated_center, threshold=0, nx=None,
+                     pixel_size=(1, 1),  min_x=None, max_x=None, mask=None):
     """Circular average of the the image data
     The circular average is also known as the radial integration
     Parameters
@@ -414,7 +413,7 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
         default is zero
     nx : int, optional
         number of bins in x
-        defaults is 100 bins
+        defaults is None, gives the pixel number of the max radius 
     pixel_size : tuple, optional
         The size of a pixel (in a real unit, like mm).
         argument order should be (pixel_height, pixel_width)
@@ -423,6 +422,8 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
         Left edge of first bin defaults to minimum value of x
     max_x : float, optional number of pixels
         Right edge of last bin defaults to maximum value of x
+    mask: array, bool type
+        same shape as image
     Returns
     -------
     bin_centers : array
@@ -430,18 +431,31 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
     ring_averages : array
         Radial average of the image. shape is (nx, ).
     """
-    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size)
-
-    bin_edges, sums, counts = utils.bin_1D(np.ravel(radial_val),
-                                           np.ravel(image), nx,
+    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size) 
+    ps =np.min( pixel_size )
+    if max_x is None:max_x = np.max(radial_val)/ps
+    if min_x is None:min_x = np.min(radial_val)/ps        
+    if nx is None:        
+        nx = int(max_x - min_x)        
+    if mask is not None:          
+        mask = np.array( mask, dtype = bool)
+        binr = radial_val[mask]
+        image_mask =     np.array( image )[mask]        
+    else:        
+        binr = np.ravel( radial_val ) 
+        image_mask = np.ravel(image)         
+    bin_edges, sums, counts = utils.bin_1D( binr,
+                                           image_mask,
+                                           nx,
                                            min_x=min_x,
-                                           max_x=max_x)
+                                           max_x=max_x)   
+    
     th_mask = counts > threshold
     ring_averages = sums[th_mask] / counts[th_mask]
-
     bin_centers = utils.bin_edges_to_centers(bin_edges)[th_mask]
+    return bin_centers, ring_averages 
 
-    return bin_centers, ring_averages
+
 
 
 def kymograph(images, labels, num):


skbeam/core/roi.py
```diff
@@ -397,9 +397,8 @@ def mean_intensity(images, labeled_array, index=None):
         mean_intensity[n] = ndim.mean(img, labeled_array, index=index)
     return mean_intensity, index
 
-
-def circular_average(image, calibrated_center, threshold=0, nx=100,
-                     pixel_size=(1, 1),  min_x=None, max_x=None):
+def circular_average(image, calibrated_center, threshold=0, nx=None,
+                     pixel_size=(1, 1),  min_x=None, max_x=None, mask=None):
```
sameera2004: @yugangzhang yes I agree with you `mask is required to make a correct normalization of circular average.`. But you can take care of that outside of this function. that means do this step before using the circular average
`image_mask =     np.array( image )[mask] `

yugangzhang:
No you can't. I that case, you just count more zeroes!

danielballan:
@yugangzhang You can see that adding this argument would set a precedent in the package: this problem will come up in many places, and we'll want to be consistent about it. We should consider our options:

1. Adopt a standard that `np.nan` means "masked out"
2. Accept an optional `mask` argument
3. Use numpy masked array

I mention (3) for completeness, but in fact numpy wants to phase these out, so we should avoid that.

I would say that (1) is more common, but it has a downside. Numpy has no integer NaN, so using `np.nan` means that all masked images must be float images.

Thoughts?

yugangzhang:
@danielballan I think that the np.nan is potentially a good idea. But, the current util.bin1D doesn't work for histogram of nan. I did all the test for four cases, including no mask, bool masked array, nan masked array, and mask options. I put it on the https://github.com/yugangzhang/TestCode/blob/master/Test_Circular_Average_YG.ipynb
I also share this notebook as an example in the jupterhub.

tacaswell:
@tjlane and company should probably chime in here as they also have significant masking issues.

tjlane:
@tacaswell et al, yes we make very heavy use of masks. Will be absolutely essential.

You'll notice that in some radial averaging code I wrote I added a mask as a separate argument:
https://github.com/slaclab/timescans/blob/master/timescans/algorithms.py#L13

As @danielballan says, using nans is a decent idea, but we could also consider keeping a separate mask array around. The former has the advantage that you do not need a 2nd object. The latter is a bit more explicit. May also play nicer with external code/analysis routines. Finally, separate mask objects separate the data and the mask -- you can play with different masking techniques and keep your data constant.

Therefore I have a soft pref for a separate mask array that gets passed as necessary.

danielballan:
@tjlane That's a convincing argument.

ericdill:
There has been some ongoing conversations regarding masking.  Is there any place where those conversations are being tracked?  Based on this week's skbeam dev call, I remember someone was going to start working on dual implementations of a function to look at using mask kwarg vs a numpy masked array as the input, or something?  I guess we should wait to see where that goes before finishing up this PR.  cc @tacaswell

danielballan:
@yugangzhang Thanks for you patience on this. What is the status of this code? Are you using this mask-aware version at CHX currently?

yugangzhang:
@danielballan  Thanks! It works very well now. We are using mask aware version.

danielballan:
OK. As you can see, this hasn't been merged into scikit-beam yet. Am I correct to assume that you have a copy of the code in this PR in chxtools or some other local repo? Is the code in this PR fully to up to date with the code you have been using there?

yugangzhang:
You are correct. We have the code with mask option in our local repo. We use that code to do circular average.

For simplicity, if you need, I just can copy the code here.

danielballan:
That would be great. Force-push to this branch or open a new PR, whichever is easier.

yugangzhang:
The easiest way is just to copy code here

```
def bin_1D(x, y, nx=None, min_x=None, max_x=None):
    """
    Bin the values in y based on their x-coordinates

    Parameters
    ----------
    x : array
        position
    y : array
        intensity
    nx : integer, optional
        number of bins to use defaults to default bin value
    min_x : float, optional
        Left edge of first bin defaults to minimum value of x
    max_x : float, optional
        Right edge of last bin defaults to maximum value of x

    Returns
    -------
    edges : array
        edges of bins, length nx + 1

    val : array
        sum of values in each bin, length nx

    count : array
        The number of counts in each bin, length nx
    """
    # handle default values
    if min_x is None:
        min_x = np.min(x)
    if max_x is None:
        max_x = np.max(x)
    if nx is None:
        nx = int(max_x - min_x) 
    # use a weighted histogram to get the bin sum
    bins = np.linspace(start=min_x, stop=max_x, num=nx+1, endpoint=True)
    val, _ = np.histogram(a=x, bins=bins, weights=y)
    # use an un-weighted histogram to get the counts
    count, _ = np.histogram(a=x, bins=bins)
    # return the three arrays
    return bins, val, count



def circular_average(image, calibrated_center, threshold=0, nx=None,
                     pixel_size=(1, 1),  min_x=None, max_x=None, mask=None):
    """Circular average of the the image data
    The circular average is also known as the radial integration
    Parameters
    ----------
    image : array
        Image to compute the average as a function of radius
    calibrated_center : tuple
        The center of the image in pixel units
        argument order should be (row, col)
    threshold : int, optional
        Ignore counts above `threshold`
        default is zero
    nx : int, optional
        number of bins in x
        defaults is 100 bins
    pixel_size : tuple, optional
        The size of a pixel (in a real unit, like mm).
        argument order should be (pixel_height, pixel_width)
        default is (1, 1)
    min_x : float, optional number of pixels
        Left edge of first bin defaults to minimum value of x
    max_x : float, optional number of pixels
        Right edge of last bin defaults to maximum value of x
    Returns
    -------
    bin_centers : array
        The center of each bin in R. shape is (nx, )
    ring_averages : array
        Radial average of the image. shape is (nx, ).
    """
    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size) 
    if mask is not None:  
        mask = np.array( mask, dtype = bool)
        binr = radial_val[mask]
        image_mask =     np.array( image )[mask]        
    else:        
        binr = np.ravel( radial_val ) 
        image_mask = np.ravel(image) 
    binr_ =   binr /(np.sqrt(pixel_size[1]*pixel_size[0] ))
    bin_edges, sums, counts = bin_1D(      binr_,
                                           image_mask,
                                           nx=nx,
                                           min_x=min_x,
                                           max_x=max_x) 
    th_mask = counts > threshold
    ring_averages = sums[th_mask] / counts[th_mask]
    bin_centers = utils.bin_edges_to_centers(bin_edges)[th_mask]
    return bin_centers, ring_averages 

```

danielballan:
@yugangzheng Haha, well that *is* easy from your side, but it doesn't make it easy for the community to review it. Please check out your `cir_avg` branch of scikit-beam, or a new branch, and push this code so we can look at the changes relative to the current codebase.

sameera2004:
Isn't bin_1D is the same function we have in skbeam.core.utils.py?

CJ-Wright:
We should use a more flexible method of computing the statistics. The statistical concept of *average* is a bit fuzzy and could mean numerous things. scipy.binned_stats might be a better way to go.

yugangzhang:
@danielballan I just push the new code in to my cir_avg branch on yugangzhang/scikit-beam. Thanks!

yugangzhang:
@CJ-Wright Good suggestions. Can you make a test?

skbeam/core/roi.py
```diff
@@ -398,8 +398,54 @@ def mean_intensity(images, labeled_array, index=None):
     return mean_intensity, index
 
 
-def circular_average(image, calibrated_center, threshold=0, nx=100,
-                     pixel_size=(1, 1),  min_x=None, max_x=None):
+
+def bin_1D(x, y, nx=None, min_x=None, max_x=None):
```
sameera2004: @yugangzhang Can you please take out the bin_1D function and use the same function from skbeam.core.utils.?

skbeam/core/roi.py
```diff
@@ -430,18 +476,27 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
...
+    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size) 
+    if mask is not None:  
+        mask = np.array( mask, dtype = bool)
+        binr = radial_val[mask]
+        image_mask =     np.array( image )[mask]        
+    else:        
+        binr = np.ravel( radial_val ) 
+        image_mask = np.ravel(image) 
+    binr_ =   binr /(np.sqrt(pixel_size[1]*pixel_size[0] ))
+    bin_edges, sums, counts = bin_1D(      binr_,
```
sameera2004: use utils.bin_1D here

skbeam/core/roi.py
```diff
@@ -413,7 +460,7 @@ def circular_average(image, calibrated_center, threshold=0, nx=None,
         default is zero
     nx : int, optional
         number of bins in x
-        defaults is None, gives the pixel number of the max radius 
+        defaults is 100 bins
```
sameera2004: Do you want to change the nx default value to 100?

CJ-Wright:
@yugangzhang What do you mean `make a test`?
Basically you need to provide a default average and then call `scipy.stats.binned_statistic` instead of bin1D. My best example is [this code](https://github.com/CJ-Wright/xpd_workflow/blob/integration/xpd_workflow/single_event_integration.py#L136) although it also does the binning, and masking in addition to azimuthal integration.

yugangzhang:
Great!  I would like to ask you for one example. we have to test the results before accepting your codes.

CJ-Wright:
@yugangzhang for tests I would highly suggest reading [chapter 5](https://github.com/CJ-Wright/Masters_Thesis/blob/master/thesis.pdf). (Although I think the rest of it is a gripping read myself.) The code is included in the repo [here](https://github.com/CJ-Wright/Masters_Thesis/tree/master/dp/figures/scripts).

skbeam/core/roi.py
```diff
@@ -430,18 +476,27 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
...
+        binr = np.ravel( radial_val ) 
+        image_mask = np.ravel(image) 
+    binr_ =   binr /(np.sqrt(pixel_size[1]*pixel_size[0] ))
+    bin_edges, sums, counts = bin_1D(      binr_,
+                                           image_mask,
+                                           nx=nx,
                                            min_x=min_x,
-                                           max_x=max_x)
+                                           max_x=max_x) 
     th_mask = counts > threshold
```
CJ-Wright: For best usability I think it would be better to remove this line, it assumes that a threshold mask is desired, which may not be the case. If a threshold mask is desired then that could be handled in the mask argument.

skbeam/core/roi.py
```diff
@@ -430,18 +476,27 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
...
-                                           np.ravel(image), nx,
+    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size) 
+    if mask is not None:  
+        mask = np.array( mask, dtype = bool)
+        binr = radial_val[mask]
+        image_mask =     np.array( image )[mask]        
+    else:        
+        binr = np.ravel( radial_val ) 
+        image_mask = np.ravel(image) 
+    binr_ =   binr /(np.sqrt(pixel_size[1]*pixel_size[0] ))
```
CJ-Wright: Why is the geometric mean preferred over the arithmetical?

skbeam/core/roi.py
```diff
@@ -430,18 +476,27 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
     ring_averages : array
         Radial average of the image. shape is (nx, ).
     """
-    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size)
-
-    bin_edges, sums, counts = utils.bin_1D(np.ravel(radial_val),
-                                           np.ravel(image), nx,
+    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size) 
+    if mask is not None:  
+        mask = np.array( mask, dtype = bool)
```
CJ-Wright: Can't we just assume that the mask has been properly typed, potentially by specifying it in the docstring? If not, would it be possible to use `mask.astype(bool)`?

skbeam/core/roi.py
```diff
@@ -430,18 +476,27 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
     ring_averages : array
         Radial average of the image. shape is (nx, ).
     """
-    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size)
-
-    bin_edges, sums, counts = utils.bin_1D(np.ravel(radial_val),
-                                           np.ravel(image), nx,
+    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size) 
+    if mask is not None:
```
CJ-Wright: Would it not be simpler to let the `mask==None` case create a mask of ones? that way the `binr` and `image_mask` can share the same logic?

CJ-Wright: Additionally, this removes the need for the`image_mask` line, you can just hand `image[mask]` to the binning algorithm, numpy should auto-ravel the array.

skbeam/core/roi.py
```diff
@@ -430,18 +476,27 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
...
+                                           image_mask,
+                                           nx=nx,
                                            min_x=min_x,
-                                           max_x=max_x)
+                                           max_x=max_x) 
     th_mask = counts > threshold
     ring_averages = sums[th_mask] / counts[th_mask]
-
     bin_centers = utils.bin_edges_to_centers(bin_edges)[th_mask]
+    return bin_centers, ring_averages
```
CJ-Wright: Would you also want to return the mask, if you go with the auto-theshold setup? That way we have total provenance and reproducible of the data and analysis, we can always go back and examine the mask or reuse it.

skbeam/core/roi.py
```diff
@@ -430,18 +476,27 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
     ring_averages : array
         Radial average of the image. shape is (nx, ).
     """
-    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size)
-
-    bin_edges, sums, counts = utils.bin_1D(np.ravel(radial_val),
-                                           np.ravel(image), nx,
+    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size)
```
CJ-Wright: Does this work for detectors which are potentially tilted?

skbeam/core/roi.py
ordirules: I agree with @CJ-Wright , I would also prefer to let the geometric mean be done by the user after the circular average and try to keep this data as raw as possible. (It's just a constant factor to be divided after the fact) 

So I also want to know the reasoning? It can't just be set to (1,1) to by pass since you need it to map out the radial_average code. As a user I vote for arithmetic :-P

thanks!

skbeam/core/roi.py
ordirules: But counts is the counts of pixel per radius (so it's after the fact). I think it could potentially still be useful. Depending on your binning, you may want to ignore values that didn't average over enough pixels. I don't know if there is a point to returning the mask from that, just compare  threshold=0 with nonzero maybe?

skbeam/core/roi.py
ordirules: I think it can if you trick it by setting pixel_size to be anisotropic? Could be interesting to try.

skbeam/core/roi.py
```diff
@@ -430,18 +476,27 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
...
-    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size)
-
-    bin_edges, sums, counts = utils.bin_1D(np.ravel(radial_val),
-                                           np.ravel(image), nx,
+    radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size) 
+    if mask is not None:  
+        mask = np.array( mask, dtype = bool)
+        binr = radial_val[mask]
+        image_mask =     np.array( image )[mask]        
+    else:
```
ordirules: @yugangzhang , instead of this double statement, I'd be tempted just to put:
```python
    if mask is not None:
        w = np.where(mask == 1)
        radial_val = radial_val[w]
        image = image[w]
```
and leave everything else unchanged. What do you think? ( very anxious to see this code added in, useful for us :-) )

skbeam/core/roi.py
ordirules: I think you would have to circular average the mask and then find the points that are above this threshold manually then, so this code could save an extra function call.

skbeam/core/roi.py
```diff
@@ -430,18 +476,27 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
     ring_averages : array
```
ordirules: Ignore counts above `threshold`
This should be:
         Ignore counts below `threshold`

skbeam/core/roi.py
CJ-Wright: Sorry, I guess I didn't understand counts. However, that brings up a bigger question: why are you using a binning where all your bins don't have values?

skbeam/core/roi.py
CJ-Wright: Actually you don't need your second line. If the mask is a bool type then you can just use `radial_val = radial_val[mask]` see [my post](https://github.com/scikit-beam/scikit-beam/pull/429#discussion_r73264811)

skbeam/core/roi.py
CJ-Wright: Yes, I mean out of the plane. But why should we be building code we need to "trick"? It might be better to take the array which translates pixel position to radial position in as a parameter. This way no tricking is needed.

CJ-Wright:
I think this might do things a bit more cleanly #459.

ordirules:
@CJ-Wright that is a nice PR with nice features.
However, in my opinion, I think including the r_array in my opinion defeats the purpose of the circular average. Now it's 1D binning with extra options. Note that r_array could now be many things (could be a qx average etc) and thus it no longer makes sense to call it a circular average. I think circular averages are common enough that they deserve their own function call with a simplistic set of arguments (with using calibrated center as opposed to pre-binning being a convenience).

Although I think it strays from the focus of this PR, I think your PR definitely contains many useful features, such as auto binning which could be quite convenient. I think you should definitely keep those ideas in mind. :-)

CJ-Wright:
@ordirules I guess I don't see too much difference between a circular average and 1D binning, given that both seek to reduce a 2D image to its radial average. The point of the `r_array`, which could be obtained from the calibrated center, was that it could correct for detector tilt and rotation. Otherwise the peak shapes will be incorrect. Additionally, unless the binning is pre-specified, an r_array which is not the array of radial distances will have the wrong bin resolution. 

As a point of clarification, the r_array does not pre-bin the image, it simply states where each pixel is in the radial direction from the center. This could be obtained from `utils.radial_grid`, but I would avoid assuming that dependency as `radial_grid` does not correct for detector tilt. Things like [pyFAI](https://github.com/silx-kit/pyFAI) do have these corrections, and give out a similar radial array. Thus the code now supports any radial grid, instead of just the uncorrected one.

skbeam/core/roi.py
danielballan: +1 for making this a separate function. We can build convenience functions out of these modular pieces as need be (probably in downstream utilities like chxtools).

danielballan:
I think @ordirules is making the point that a simple circular average -- specified only by a central point, ignoring tilt and rotation corrections -- is a common enough use case that convenience may be important than maximizing modularity and requiring the user to provide their own r bins. I would also guess that, down the road, we can squeeze performance optimizations from this simple case that we can't get from the general case with arbitrary r bins.

I think we need both this and #459.

CJ-Wright:
@danielballan @ordirules Would it be valid to make a `simple_circular_average`? This would wrap the circular_average and provide the rbins from the central point.

danielballan:
Just to keep everyone in the loop: in an offline conversation, @CJ-Wright and I came to the conclusion that we can approach this in two steps. We'll merge this PR with the current signature -- once the implementation and return value has been ironed out, as discussed in the inline comments. Then we'll revisit #459 and see what parts of `circular_average` might be refactored to utilize it.

skbeam/core/roi.py
ordirules: That's what I mean with "np.asarray(mask,bool)" which should have been np.array(mask,dtype=bool), sorry. I agree with you though, the bool is probably faster.
Anyway, I'm just trying to argue that there doesn't need to be an else statement here. Just if there's a mask, do something extra to the array, if not, leave it alone and leave the rest of the code alone. We shouldn't slow down the code that's been used by others without mask if we can, I think.

CJ-Wright: I'm not certain how slow the masking code will slow everything down. I think the call should be fairly fast, although if the profiling shows it to be slow, we can start speeding it up. However, I expect the binning to be much slower than the mask generation.

ordirules: you're right, the binning is very likely the bottleneck, sorry (got carried away with just wanting the least lines of added code possible :-)).

danielballan:
These commits were included in #462 and merged. Closing. Thanks again, @yugangzhang!

PR #430, opened 2016-03-21
array projections

chrisvam:
Hey everyone,

This is a work-in-progress, but I thought I would start getting comments early.  It is code from TJ, Mikhail and myself that does radial projections of a 2D image by calling a class that can do more general projections (e.g. for multi-panel detectors or for more unusual projections).  In the more general case the user must specify the "bin_value" for each pixel that will be used to be create the projection histogram.

There are some subtleties with x/y flips that I'm still sorting out to try to make it intuitive for someone who is looking at a matplotlib plot (which is shown in row-column order (vs. cartesian order)).  Those flips are causing it to crash with non-square images.

Also, the testing is currently just a place-holder: not a real test.

chris


skbeam/core/accumulators/arrayProjector.py
```diff
@@ -0,0 +1,155 @@
...
+        # this with arrays (matrices).  numpy.meshgrid (version 1.6) uses
+        # cartesian (non-matrix) coordinates, so we need to do the flip
+
+        xsize   = columns
+        ysize   = rows
+
+        xc    = xsize/2                  if xc    is None else xc
+        yc    = ysize/2                  if yc    is None else yc
+        rmin  = 0                        if rmin  is None else rmin
+        rmax  = math.sqrt(xc**2 + yc**2) if rmax  is None else rmax
```
tjlane: might as well use np.sqrt() -- I think we could get rid of the math module import

skbeam/core/accumulators/arrayProjector.py
```diff
@@ -0,0 +1,155 @@
...
+        if phimax is not None:
+            weights[phipix>phimax] = 0
+        if rmin is not None:
+            weights[rpix<rmin] = 0
+        if rmax is not None:
+            weights[rpix>rmax] = 0
+
+        super(RadialProjector,self).__init__(rpix, weights, nbins=nbins, norm=norm)
+
+    def __call__(self,image):
```
tjlane: I think this function definition is redundant... the subclass will automatically inherit this method from the parent and it will work OK.

tjlane:
@chrisvam I can help write tests/demo nose this afternoon if that's useful

skbeam/core/accumulators/arrayProjector.py
```diff
@@ -0,0 +1,155 @@
...
+            weights = image.flatten()
+        else:
+            weights = image.flatten() * self.weights.flatten()
+            if not (image.shape == self.weights.shape):
+                raise ValueError('`image` and `weights` must have the same shape')
+
+        bin_values = np.bincount(self._bin_assignments.flatten(), weights=weights)
+
+        if self.norm:
+            bin_values /= self._normalization_array
```
ericdill: Add `from __future__ import division` so that this will *always* be float division

skbeam/core/accumulators/arrayProjector.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        Parameters
+        ----------
+        bin_values : np.ndarray
+            For each pixel, this is the value of that determines which bin
+            a pixel's intensity will contribute to
+        weights : np.ndarray
+            A weight for each pixel with the same shape as bin_values (can be zero to ignore a pixel)
+        nbins : int
+            The number of bins to employ. If `None` guesses a good value.
```
ericdill: docstring and code are inconsistent. There is no `if nbins is None:` handling in the `__init__`.  Given the current code, this should probably just read `The number of bins to employ`.  I am 50/50 on adding heuristics to guess a good value of nbins. numpy recently picked up an `auto` keyword for its histogram function.  You may want to consider using their machinery for guessing an appropriate number of bins. https://github.com/numpy/numpy/blob/master/numpy/lib/function_base.py#L300

skbeam/core/accumulators/arrayProjector.py
```diff
@@ -0,0 +1,155 @@
...
+        self.binvalRange = self.bin_values.max() - self.bin_values.min()
+        self.bin_width = self.binvalRange / (float(nbins) - 1)
+
+        self._bin_assignments = np.floor( (self.bin_values - self.bin_values.min()) / self.bin_width ).astype(np.int32)
+        if weights is None:
+            self._normalization_array = (np.bincount( self._bin_assignments.flatten() ) \
+                                             + 1e-100).astype(np.float)
+        else:
+            self._normalization_array = (np.bincount( self._bin_assignments.flatten(), weights=self.weights.flatten() ) \
+                                             + 1e-100).astype(np.float)
```
ericdill: Why add an infinitesimally small number here?

skbeam/core/accumulators/arrayProjector.py
chrisvam: agreed

skbeam/core/accumulators/arrayProjector.py
```diff
@@ -0,0 +1,155 @@
...
+
+        if not (image.shape == self.bin_values.shape):
+            raise ValueError('`image` and `bin_values` must have the same shape',image.shape,self.bin_values.shape)
+
+        if self.weights is None:
+            weights = image.flatten()
+        else:
+            weights = image.flatten() * self.weights.flatten()
+            if not (image.shape == self.weights.shape):
+                raise ValueError('`image` and `weights` must have the same shape')
```
ericdill: Can you add the shapes of `image` and `weights` here too?

skbeam/core/accumulators/arrayProjector.py
chrisvam: Agreed.  Tried that earlier, and it failed, but must have been because of another reason since it works now.

ericdill:
Did you mean to have a `self.bin_values += bin_values` somewhere in the `__call__` function so that this function accumulates the histogram for each image?

skbeam/core/accumulators/arrayProjector.py
```diff
@@ -0,0 +1,155 @@
...
+        """   
+        # flip these to make it more intuitive for users who will use
+        # this with arrays (matrices).  numpy.meshgrid (version 1.6) uses
+        # cartesian (non-matrix) coordinates, so we need to do the flip
+
+        xsize   = columns
+        ysize   = rows
+
+        xc    = xsize/2                  if xc    is None else xc
+        yc    = ysize/2                  if yc    is None else yc
```
ericdill: if this is supposed to be int division, this will need to be changed to `xsize // 2` once `from __future__ import division` is added

skbeam/core/accumulators/arrayProjector.py
chuckie82: I would replace the infinitesimally small number to **np.finfo(np.float).eps**

skbeam/core/accumulators/arrayProjector.py
```diff
@@ -0,0 +1,155 @@
...
+        -----------
+        rows,columns:  shape of image to be projected
+        xc,yc:         location (in pixels) of origin (default: center of image)
+        rmin,rmax:     radial range to include in projection, in pixels (default: no limits)
+        phimin,phimax: phi range to include in projection, in degrees (default: no limits)
+        nbins:         number of bins in projected histogram
+        weights:       np.ndarray.  weight to be applied to each pixel in image.  this can
+                       be used as a mask if the weight is set to zero.
+        norm:          boolean indicating whether bin entries in the projected histogram should be divided
+                       by weights (number of pixels, in the case where the weights are 1).
```
ericdill: Sphinx is not going to render this docstring correctly.  The rest of the docstrings look good though.

skbeam/core/accumulators/arrayProjector.py
ericdill: Well that's pretty nifty.  What is `eps`?  Is that substantially different than `np.finfo(np.float).resolution`?

tjlane: IIRC (as the original author of this code lol) the small number avoids div by zero (ie, enforces the convention 0/0 = 0)

chuckie82: `eps` is the smallest number you can add to make a difference, which is what the code is doing. 1e-100 will be too small.

`resolution` will give a decimal point resolution, which isn't quite what we want.

tjlane:
@chrisvam @ericdill 

Upon closer review, It looks like there are two meanings of `bin_values` in the code -- one I put in, and one Chris added. I *think* there is no conflict because Chris' `bin_values` are in the self namespace, and mine are not. Chris' variable `self.bin_values` indicates the value of each pixel to use to bin (e.g. radius), while mine indicated the value of the final histogram for that particular bin.

Both names are pretty ambiguous. Hence Eric's confusion above!

Suggested name changes:
* `self.bin_values` to `bin_variable` or `binning_axis_value`
* `bin_values` to `histogram_values` (or keep same)

Better suggestions very welcome...

skbeam/core/accumulators/arrayProjector.py
chrisvam: I thought about this line.  It is somewhat non-ideal (in other pieces of code we have disabled numpy divide-by-zero warnings for specific lines of code).  I guess the goal is to protect the novice user from scary looking warning messages and give them understandable results that don't include nan's.  It's certainly debatable, but my best guess is that we should include the offset (although I'll likely switch to Chuck's "eps").

skbeam/core/accumulators/arrayProjector.py
chrisvam: You're right.  I think I'll take out the default value.

skbeam/core/accumulators/arrayProjector.py
chrisvam: You're right.  I thought about this logic a little more, and I think we don't need these rmax/rmin lines anymore.  Will remove them.

skbeam/core/accumulators/arrayProjector.py
chrisvam: Agreed.  Done.

chrisvam:
Eric, for your self.bin_values += bin_values comment:  I think that should be left up to the users.  This code will project 1 image.  The user can decide outside of this if they want to bump a histogram.

TJ: I changed the second "bin_values" to "histogram".

chrisvam:
@chuckie82 @tjlane @ericdill I think I've improved this quite a bit, in particular I feel better about the user-controlled switch from "cartesian" (where column corresponds to x) to "matrix" (where row corresponds to x).  Any further suggestions?  I guess we still have to decide on a directory.

I would  also still like to put in scikit-beam examples.

chrisvam:
@danielballan @tacaswell Sending this pull request to you instead, since I heard Eric is busy warming bottles, etc.  Do you think this pull request needs more work?  The one thing I'm not sure of is where to put it.  I put it in accumulators since it's integrating, fundamentally, but I'm not sure what you precisely meant by accumulators.

I also want to add some related stuff to scikit-beam-examples.

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+    Project a 2D image onto a radial axis.
+    """
+
+    def __init__(self, xsize, ysize, nbins, xc=None, yc=None, rmin=None,
+                 rmax=None, phimin=None, phimax=None, weights=None,
+                 norm=True, cartesian=True):
+        """
+        Parameters:
+        -----------
+        xsize,ysize:   shape of image in pixels
```
tacaswell: please use numpy docstring format

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
+#!/usr/bin/env python
```
tacaswell: do not need this line

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+
+from __future__ import division
```
tacaswell: should have the standard set of future import

```
from __future__ import (absolute_import, division, print_function,
                        unicode_literals)
```

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+import numpy as np
+
+"""
+authors: Mikhail Dubrovin, TJ Lane, Christopher O'Grady
```
tacaswell: we do not have notes like this on any other file

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        Parameters
+        ----------
+        bin_values : np.ndarray
+            For each pixel, this is the value of that determines which bin
+            a pixel's intensity will contribute to
+        weights : np.ndarray
+            A weight for each pixel with the same shape as bin_values (set
+            to zero to ignore a pixel)
+        nbins : int
```
tacaswell: please put this in order of the parameters, `norm` is missing.

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class Projector(object):
+
+    def __init__(self, bin_values, nbins, weights=None, norm=True):
+        """
+        Parameters
+        ----------
+        bin_values : np.ndarray
+            For each pixel, this is the value of that determines which bin
```
tacaswell: Is this an array of ints such that `0 <= bv < nbins` for all `bv in bin_values` or an array of arbitrary floats which will be made discrete in the function call?

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+            self._ibv = np.s_[:]  # slice that returns arr unchanged
+
+        bv = self.bin_values[self._ibv]
+        wt = self.weights[self._ibv]
+
+        binvalRange = bv.max() - bv.min()
+        # add a little bit to not have the bv.max() value create an extra bin
+        # in the np.floor statement below
+        binvalRange *= 1.0+1e-9
+        self.bin_width = binvalRange / (float(self.nbins))
```
tacaswell: do not need the float cast.

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+
+        self._bin_assignments = np.floor((bv - bv.min()) /
+                                         self.bin_width).astype(np.int32)
+        smallval = np.finfo(np.float).eps
+        if weights is None:
+            self._normalization_array = \
```
tacaswell: please use `()` continuation not `\` line continuation.

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+import numpy as np
+
+"""
+authors: Mikhail Dubrovin, TJ Lane, Christopher O'Grady
+"""
+
+
+class Projector(object):
+
+    def __init__(self, bin_values, nbins, weights=None, norm=True):
```
tacaswell: This should take `vmin` and `vmax` as inputs instead of always doing 'tight' binning.  Matching the `np.histogram` api would be a good idea.

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+
+from __future__ import division
+import numpy as np
+
+"""
+authors: Mikhail Dubrovin, TJ Lane, Christopher O'Grady
+"""
+
+
+class Projector(object):
```
tacaswell: This does not seem like the right name to me, I expected this to be related to back projection or some such.

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+            self._normalization_array = \
+                (np.bincount(self._bin_assignments.flatten()) + smallval) \
+                .astype(np.float)
+        else:
+            self._normalization_array = \
+                (np.bincount(self._bin_assignments.flatten(),
+                             weights=wt.flatten()) + smallval) \
+                .astype(np.float)
+
+        assert self.nbins == self._bin_assignments.max()+1, \
```
tacaswell: please use a `if ... raise ...` pattern here.  assert statements can be ignored by the interpreter.

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+
+        binvalRange = bv.max() - bv.min()
+        # add a little bit to not have the bv.max() value create an extra bin
+        # in the np.floor statement below
+        binvalRange *= 1.0+1e-9
+        self.bin_width = binvalRange / (float(self.nbins))
+
+        self._bin_assignments = np.floor((bv - bv.min()) /
+                                         self.bin_width).astype(np.int32)
+        smallval = np.finfo(np.float).eps
```
tacaswell: Please put this as a constant at module level.

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+        # _ibv = included bin values
+        # this is used to select only unmasked pixels in the analysis
+        # (i.e. those with weight!=0)
+        if weights is not None:
+            self._ibv = (weights != 0)
+        else:
+            self._ibv = np.s_[:]  # slice that returns arr unchanged
+
+        bv = self.bin_values[self._ibv]
+        wt = self.weights[self._ibv]
```
tacaswell: if weights is `None` this will raise?

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+        # in the np.floor statement below
+        binvalRange *= 1.0+1e-9
+        self.bin_width = binvalRange / (float(self.nbins))
+
+        self._bin_assignments = np.floor((bv - bv.min()) /
+                                         self.bin_width).astype(np.int32)
+        smallval = np.finfo(np.float).eps
+        if weights is None:
+            self._normalization_array = \
+                (np.bincount(self._bin_assignments.flatten()) + smallval) \
```
tacaswell: it looks like `smallval` is being used here to prevent divide by 0 warnings later?  It is probably better to replace the 0s by `np.nan` rather than add a small value to every normalization

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+        self.nbins = nbins
+        self.norm = norm
+
+        # _ibv = included bin values
+        # this is used to select only unmasked pixels in the analysis
+        # (i.e. those with weight!=0)
+        if weights is not None:
+            self._ibv = (weights != 0)
+        else:
+            self._ibv = np.s_[:]  # slice that returns arr unchanged
```
tacaswell: should use `self._ibv = np.ones_like(bin_values, dtype='bool')`.  As it is in the next step when you use these to slice the bin_values and weights you get different shaped arrays out.  In this code path the same of weight/bin_values is unchanged, in the other code path you get a 1D array.  By always using a bool array you will eliminate the need for the `flatten` calls later in the code

skbeam/core/accumulators/projector.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @property
+    def bin_centers(self):
+        """
+        Returns:
+        --------
+        bin_centers : ndarray, float
+            The center of each bin.
+        """
+        return (np.arange(self.nbins) + 0.5) * self.bin_width + \
```
tacaswell: no `\` line continuations

skbeam/core/accumulators/projector.py
chrisvam: done

skbeam/core/accumulators/projector.py
chrisvam: done

skbeam/core/accumulators/projector.py
chrisvam: removed

skbeam/core/accumulators/projector.py
chrisvam: fixed

skbeam/core/accumulators/projector.py
chrisvam: Yes.  I have tried to make that more explicit in the doctoring now.

skbeam/core/accumulators/projector.py
chrisvam: removed

skbeam/core/accumulators/projector.py
chrisvam: remove all \.

skbeam/core/accumulators/projector.py
chrisvam: I thought about it, and I removed the assert statements:  I believe I was checking conditions that should never happen if I wrote the code correctly.

skbeam/core/accumulators/projector.py
chrisvam: Following one of your later comments (setting 0 values to NaNs) I think I was able to remove this constant entirely.

skbeam/core/accumulators/projector.py
chrisvam: You are correct.  I have fixed this error and added tests where weights=None.

skbeam/core/accumulators/projector.py
chrisvam: Done.

skbeam/core/accumulators/projector.py
chrisvam: I have restructured the code to eliminate the np.s_[:] and reduce the number of flatten() calls (all except one are in the calls to numpy.bincount)

skbeam/core/accumulators/projector.py
chrisvam: Done.

skbeam/core/accumulators/projector.py
chrisvam: I have tried to use numpy docstring format now.

tacaswell:
This looks much better!  Sorry for leaving so many comments :disappointed: .

One minor suggestion is that you could merge `self._norm` and `self._normalization_array` to something like

```python
if norm:
    self._norm = make_array_logic
else:
    self._norm = None
```

and in the `__call__` do something like

```python
if self._norm is not None:
    histogram /= self._norm
```

It just eliminates on piece of state and ensures that the attributes that the object has are always consistent.

chrisvam:
@tacaswell On the contrary: your comments are excellent, and very important if we want to make this world-class software (an extremely difficult task).  Thanks for those.  Your "norm" comment is also correct.  I have changed that.

skbeam/core/accumulators/projector.py
chrisvam: I'm not married to this and happy to discuss.  Perhaps on the call tomorrow?

skbeam/core/accumulators/projector.py
chrisvam: I've thought about this and will argue somewhat to the contrary:  currently the user can specify the equivalent of vmin/vmax arguments (and indeed many more complex "masks") by setting elements of the "weights" argument to zero.  I suggest that adding vmin/vmax is a more restrictive way of doing the same thing and makes the interface less clean.  But happy to discuss, of course.  Perhaps on the call tomorrow?

skbeam/core/accumulators/projector.py
chrisvam: @tacaswell Howzabout BinImage as a name?  (and RadialBinImage for the specific radial case).  "Bin" could conceivably be confused with "binary" but I think most photon science users would not do that.

One thing I do need your input on:  what directory should this go in?

licode: This is a very good question. I think we can put this at the accumulators directory for now. But it will be helpful if we can think about better structure to organize tools at skbeam, especially when the package is growing larger.

chrisvam: @licode @tacaswell Hi guys, following Li's advice I have left the code in accumulators.  Following Tom's advice I have renamed the classes BinImage and RadialBinImage.  Let me know if you have further thoughts.  Thanks!

skbeam/core/accumulators/projector.py
chrisvam: @tacaswell Hi Tom, I thought about this comment more in light of what you said during the mtg on Monday and I believe you are correct:  the user needs a way to control the binning more easily than with just the "weights" array.  I have added that in the most recent version I just uploaded.

CJ-Wright:
What does this provide above and beyond [scipy.stats.binned_statistic](http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.stats.binned_statistic.html)?

chrisvam:
Hi Chris,

Yes, the BinImage interface is very close to the binned_statistic interface, so your point is worth serious consideration.  I believe what we gain is performance:  most of the time scipy/numpy provide routines that don't maintain any state.  By maintaining state in the BinImage class (e.g. computing the bin-values and limits only once) BinImage appears to be a factor of 3 faster than binned_statistic according to some quick test code I wrote (code/results below).  This test tries to emulate the way it would typically be used at LCLS: binning a mega-pixel image, with masking, at the 120Hz shot rate.  The CPU consumption of image-binning has been an issue for several experiments here.

Also, for completeness, the proposed RadialBinImage class provides an easier user-interface for radial integrations compared to binned_statistic.

As an aside, I think this "state" idea is the same reason we need skbeam/core/accumulators/Histogram.pyx: the available python histogramming code that I'm aware of doesn't maintain state, so makes it difficult/expensive to increment one bin (for example) when a new shot is analyzed.

chris

pc91011:accumulators cpo$ python ~/junk/speed.py 100
binned_statistic time: 4.53855204582 BinImage Time: 1.30998492241 Ratio: 3.46458342244
pc91011:accumulators cpo$ 

`from skbeam.core.accumulators.bin_image import BinImage
import time
import numpy as np
from scipy.stats import binned_statistic
import sys
niter = int(sys.argv[1])

bin_values = np.random.rand(1000000)
mask = np.zeros_like(bin_values)
mask[bin_values>0.5] = 1
bins = 100
bi = BinImage(bin_values,False,bins,weights=mask)
image = np.random.rand(1000000)

start = time.time()
for i in range(niter):
    bi(image)
binimagetime = time.time()-start

bv = bin_values[mask==1]
start = time.time()
for i in range(niter):
    binned_statistic(bv,image[mask==1],statistic='sum',bins=bins)
binstattime = time.time()-start

print 'binned_statistic time:',binstattime,'BinImage Time:',binimagetime,'Ratio:',binstattime/binimagetime
`

skbeam/core/accumulators/bin_image.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def __init__(self, bin_values, norm, nbins, min=None, max=None,
+                 weights=None):
+        """
+        Parameters
+        ----------
+        bin_values : np.ndarray
+            For each pixel, this is the value that determines which bin
+            a pixel's intensity will contribute to.  It can be an arbitrary
+            array of floats/ints which will be discretized into nbins equally
```
CJ-Wright: Are equally sized bins appropriate for an integration to Q?

skbeam/core/accumulators/bin_image.py
```diff
@@ -0,0 +1,219 @@
...
+            These can be set to zero to ignore a pixel, similar to a mask.
+            If not provided, each pixel is assigned a weight of 1.
+        """
+
+        self._bin_values = bin_values
+        self._weights = weights
+        self._nbins = nbins
+        self._norm = None
+        self._min = min
+        self._max = max
```
CJ-Wright: You might want to rename your min and max variables, they shadow the existing python min and max functions

CJ-Wright:
Chris,
Thanks for the reply.
I have a few questions:
 1. Does your method allow for different types of operations on the binned pixels, like binned statistic allows for median, mean, and other functions?
 1. Can you do a more full profile of your code vs. binned_statistic? I am very curious where the speed up is coming from.

skbeam/core/accumulators/bin_image.py
chrisvam: Even thought I work on LCLS, I have more of a high-energy-physics background, so I can't comment so well on the Q question.  I will say, however, that in principle the user can define BinImage bin_values however he/she likes to get whatever binning is desired.

skbeam/core/accumulators/bin_image.py
chrisvam: Absolutely correct.  I have renamed them to vmin/vmax in the commit just now.

chrisvam:
Chris Wright wrote:

> Does your method allow for different types of operations on the binned pixels, like binned statistic allows for median, mean, and other functions?

It doesn't at the moment, but I guess it could in the future, like binned_statistic.  To make that possible, perhaps I should switch the "norm" bool keyword to be like "statistic" string in binned_statistic?  I'd have to think for a bit to see if it makes sense in the same way:  the BinImage "weights" kwarg behaves a little different than binned_statistic.

> Can you do a more full profile of your code vs. binned_statistic? I am very curious where the speed up is coming from.

Sure ... I've included a small profiling example and output below.  It looks like the major culprit is a call to np.digitize that happens on every call to binned_statistic.  I also read the binned_statistic code a little: unsurprisingly, it calls np.bincount, the same as BinImage, but there is indeed quite a bit of setup (which is visible by comparing the profile output of the two below).

chris

p.s. after looking at the profile output I realized flatten() was copying an array, and I could avoid that with resize(-1).  that is fixed in my latest commit.

`from skbeam.core.accumulators.bin_image import BinImage
import numpy as np
from scipy.stats import binned_statistic
import sys
niter = int(sys.argv[1])

bin_values = np.random.rand(1000000)
mask = np.zeros_like(bin_values)
mask[bin_values>0.5] = 1
bins = 100
bi = BinImage(bin_values,False,bins,weights=mask)
image = np.random.rand(1000000)

bv = bin_values[mask==1]

def testbinimage():
    for i in range(niter):
        bi(image)

def testbinnedstatistic():
    for i in range(niter):
        binned_statistic(bv,image[mask==1],statistic='sum',bins=bins)

import cProfile
cProfile.run('testbinimage()',sort='tottime')
cProfile.run('testbinnedstatistic()',sort='tottime')
`

`         304 function calls in 1.188 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      100    0.965    0.010    1.136    0.011 bin_image.py:100(__call__)
      100    0.170    0.002    0.170    0.002 {numpy.core.multiarray.bincount}
        1    0.052    0.052    1.188    1.188 speed.py:16(testbinimage)
      100    0.001    0.000    0.001    0.000 {method 'reshape' of 'numpy.ndarray' objects}
        1    0.000    0.000    1.188    1.188 <string>:1(<module>)
        1    0.000    0.000    0.000    0.000 {range}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


         8504 function calls in 4.633 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      100    2.822    0.028    2.822    0.028 {numpy.core.multiarray.digitize}
        1    0.859    0.859    4.633    4.633 speed.py:20(testbinnedstatistic)
      100    0.403    0.004    0.403    0.004 {method 'round' of 'numpy.ndarray' objects}
      100    0.184    0.002    3.739    0.037 _binned_statistic.py:257(binned_statistic_dd)
      100    0.118    0.001    0.118    0.001 {numpy.core.multiarray.bincount}
      800    0.072    0.000    0.072    0.000 {numpy.core.multiarray.array}
      500    0.059    0.000    0.059    0.000 {method 'reduce' of 'numpy.ufunc' objects}
      100    0.050    0.001    0.050    0.001 {numpy.core.multiarray.zeros}
      100    0.034    0.000    3.774    0.038 _binned_statistic.py:14(binned_statistic)
      100    0.012    0.000    0.012    0.000 {numpy.core.multiarray.where}
      800    0.003    0.000    0.003    0.000 {numpy.core.multiarray.arange}
      100    0.002    0.000    0.003    0.000 function_base.py:9(linspace)
      100    0.001    0.000    0.001    0.000 {method 'round' of 'numpy.generic' objects}
      500    0.001    0.000    0.073    0.000 numeric.py:476(asanyarray)
      200    0.001    0.000    0.002    0.000 shape_base.py:9(atleast_1d)
      100    0.001    0.000    0.001    0.000 function_base.py:1116(diff)
      100    0.001    0.000    0.073    0.001 shape_base.py:61(atleast_2d)
      200    0.001    0.000    0.001    0.000 {numpy.core.multiarray.empty}
      200    0.001    0.000    0.405    0.002 fromnumeric.py:2694(around)
     1200    0.001    0.000    0.001    0.000 {len}
      200    0.001    0.000    0.031    0.000 {method 'min' of 'numpy.ndarray' objects}
      100    0.001    0.000    0.001    0.000 fromnumeric.py:712(sort)
      100    0.000    0.000    0.000    0.000 {method 'copy' of 'numpy.ndarray' objects}
      100    0.000    0.000    0.002    0.000 {method 'prod' of 'numpy.ndarray' objects}
      100    0.000    0.000    0.000    0.000 {method 'reshape' of 'numpy.ndarray' objects}
      100    0.000    0.000    0.001    0.000 numeric.py:1970(isscalar)
      200    0.000    0.000    0.031    0.000 _methods.py:28(_amin)
      100    0.000    0.000    0.000    0.000 {numpy.core.multiarray.result_type}
      300    0.000    0.000    0.000    0.000 {method 'append' of 'list' objects}
      200    0.000    0.000    0.000    0.000 {method 'argsort' of 'numpy.ndarray' objects}
      100    0.000    0.000    0.000    0.000 {method 'fill' of 'numpy.ndarray' objects}
      100    0.000    0.000    0.026    0.000 {method 'max' of 'numpy.ndarray' objects}
      100    0.000    0.000    0.001    0.000 {method 'any' of 'numpy.ndarray' objects}
      200    0.000    0.000    0.000    0.000 {built-in method __new__ of type object at 0x100186920}
      100    0.000    0.000    0.026    0.000 _methods.py:25(_amax)
      100    0.000    0.000    0.000    0.000 {method 'astype' of 'numpy.ndarray' objects}
      100    0.000    0.000    0.000    0.000 {isinstance}
      100    0.000    0.000    0.000    0.000 <string>:8(__new__)
      100    0.000    0.000    0.002    0.000 _methods.py:34(_prod)
      100    0.000    0.000    0.000    0.000 {method 'sort' of 'numpy.ndarray' objects}
      100    0.000    0.000    0.000    0.000 numeric.py:406(asarray)
      100    0.000    0.000    0.001    0.000 _methods.py:37(_any)
      100    0.000    0.000    0.000    0.000 {method 'swapaxes' of 'numpy.ndarray' objects}
      100    0.000    0.000    0.000    0.000 {callable}
        1    0.000    0.000    4.633    4.633 <string>:1(<module>)
        1    0.000    0.000    0.000    0.000 {range}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


pc91011:accumulators cpo$ `

skbeam/core/accumulators/bin_image.py
CJ-Wright: That might not necessarily be fair. Since the bin_values describes where a pixel is on your histogram while the binning itself describes which pixels go together and/or the resolution of your pixels. For instance the Q resolution of the pixels actually changes as a function of Q.
[res.pdf](https://github.com/scikit-beam/scikit-beam/files/266022/res.pdf) This is an example with a Perkin Elmer detector.

chrisvam: Ah ... i may be understanding (but maybe not).  So you're saying that it would be useful to not only bin in Q, but also delta-Q and that binned_statistic has the ability to do this?

CJ-Wright: I am more saying that being able to specify some crazy collection of bins would be useful, as the size of each bin might not be equal to other bin sizes. `binned_statistic` has the ability to specify the bin edges.

CJ-Wright: You might look at #444 for the binning algorithm

chrisvam: Do you think I should consider refactoring the binned_statistic code so that it maintains state in a class instance (hence faster) but has all the features you want?

CJ-Wright: I mean that is up to you. One could take your current class and add the functionality to it. ie. if someone hands it a statistic function or their own function then use a different chunk of code. Incidentally, I have written a prototype of binned_statistic which uses multiprocessing, and does all the statistical measures in one shot. https://gist.github.com/CJ-Wright/747476a33ada014d7d91f6731a4c77f3 I don't know if that is helpful or not. You should also fee free to ignore me, I mean it sounds like these might be different use cases and one does not invalidate the other.

tacaswell:
As #446 / #453 have been merged I think we can drop this as that is a more general implementation of this?

chrisvam:
Yes, I believe this approach is not as good as #453 so can be closed.

PR #432, opened 2016-03-28
DOC: Fixed reference

stuwilkins:
The reference was wrong, I think if fixed it in all places.


ericdill:
Great catch. Thanks @stuwilkins !

PR #434, opened 2016-04-04
Histogram

CJ-Wright:
@ericdill

### DIFF
--- a/skbeam/core/accumulators/histogram.pyx
+++ b/skbeam/core/accumulators/histogram.pyx
@@ -1,3 +1,4 @@
+from __future__ import division
 """
 Histogram
 
--- a/skbeam/core/accumulators/tests/test_histogram.py
+++ b/skbeam/core/accumulators/tests/test_histogram.py
@@ -1,5 +1,6 @@
+from __future__ import division
 import numpy as np
-from numpy.testing import assert_array_almost_equal
+from numpy.testing import assert_array_almost_equal, assert_array_equal
 from numpy.testing import assert_almost_equal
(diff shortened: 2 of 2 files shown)


ericdill:
Ok I figured out what was going on here.  There are three things to note.

1. If you are on py2 and you provide three integers as the `(nbins, low, high)` tuple to `Histogram.__init__` then py2 will *helpfully* do integer arithmetic and result in binsizes potentially being zero because of `(high-low) / nbins` when nbins > (high-low).  In any case, the bin size will not be what you expect it to be. This is solved with the addition of `from __future__ import division` at the top of the histogram.pyx code (which you already have in 9dec76b)

2. The histogram accumulator class in histogram.pyx bins on the half-open interval from `low` to `high` in steps of `(high-low)/nbins` which means that if any of the coordinates you are trying to bin happen to be *exactly* equal to the right-most edge, then they will be excluded from the histogram. This is worthy of note because it is *not* how numpy does its binning.  I think I would like a test that shows this behavior, which, coincidentally `test_simple_fail` does.  Could you add `from nose.tools import raises` and then add an `@raises(AssertionError)` as a decorator to `test_simple_fail`?

3. The only unexpected reason why this test is failing, then, is because you need to do a weighted histogram to make sure that these arrays are actually equal.  To do this, change the histogram call to be `histogram(a, h.edges[0], weights=b)` to be the same as `h.fill(a, weights=b)`.

4. Finally, and this is really just a note, this test will pass if you change the `binlowhigh` tuple to be `(5, 0, 3.1)` because none of the coordinates fall on the right-most bin edge and then our histogram class and the numpy histogram class agree.

PR #441, opened 2016-05-06
Unit change in xrf cross section

licode:
change xraylib function to output cs in barns.

### DIFF
--- a/skbeam/core/constants/tests/test_xrf.py
+++ b/skbeam/core/constants/tests/test_xrf.py
@@ -38,7 +38,8 @@
 from __future__ import absolute_import, division, print_function
 import six
 import numpy as np
-from numpy.testing import (assert_array_equal, assert_array_almost_equal, assert_raises)
+from numpy.testing import (assert_array_equal, assert_array_almost_equal,
+                           assert_raises)
 from nose.tools import assert_equal, assert_not_equal
 
 from skbeam.core.constants.xrf import (XrfElement, emission_line_search,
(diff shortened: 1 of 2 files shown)


licode:
This should be ready to go. It is very simple change on unit.

tacaswell:
I agree it is a simple code change, but it is a HUGE API change.  This is also one of the scariest sort of API breaks, nothing will obviously break (no exceptions, no syntax errors...) but the results will be wrong.


What is the motivation for this?  Are you sure no one else is using these numbers?

licode:
Right, this is like a hidden API change going above all the tests. However, I've been thinking about this for a while before making this change. The motivation comes from the elemental concentration calculation. In general, it is easy for users want to use the information, like how many atoms in the sample, to interpolate some elements not included in the reference sample. If we provide cross section in barns/atom, they only need to use total intensity counts divided by cross section (also consider reference sample), the results are shown in terms of the number of atoms. If we use cm2/g, users need to consider one more step to take care of atomic mass. And this is kind of confusing according to discussion with users.

licode:
I am thinking to add another key/value to XRAYLIB_MAP. For example, add different key and new function, while the old ones keep the same. 
```
'csb': (line_dict, xraylib.CSb_FluorLine_Kissel)
```
So nothing will break, we also have different options on output unit. This also better matches what xraylib has. @tacaswell ?

tacaswell:
:+1: to not breaking API!

licode:
New function is added without touching the old ones. This should be ready to go. @tacaswell

skbeam/core/constants/tests/test_xrf.py
```diff
@@ -118,6 +120,18 @@ def test_XrayLibWrap_Energy():
...
+def test_cs_different_units():
+    e = XrfElement('Fe')
+    # test at different energies
+    for eng in range(10, 20):
+        cs1 = np.array([v for k, v in e.cs(eng).all])   # unit in cm2/g
+        cs2 = np.array([v for k, v in e.csb(eng).all])  # unit in barns/atom
+        cs1 /= cs1[0]
+        cs2 /= cs2[0]
+        # ratio should be the same no matter which unit is used
+        assert_array_almost_equal(cs1, cs2, decimal=10)
```
tacaswell: clever test!

PR #435, opened 2016-04-05
ENH: Add edge and ring blur masks to scikit-beam

CJ-Wright:
This PR should be better.

Issue #34, opened 2014-07-31
Determining ideal bin size for Powder Diffraction


CJ-Wright:
We will need a way to calculate the size of the bins for any integration algorithm we implement.  If the bins are too large, then we loose resolution in our diffraction pattern.  If the bins are too small, then some may contain no pixels, causing the pattern to become jagged.  This will also determine our maximum Q resolution, an important part of propagating diffraction pattern error.  Most likely this should be calculated from the detector setup/calibration information.

ericdill:
This does not happen a priori, but the helper function was done in #55 .  "Automatically" determining bin sizes sounds like it requires a scientific publication to justify implementing an algorithm.  Please reopen if there is a scientific manuscript that we can obtain an algorithm from.

PR #444, opened 2016-05-13
ENH: new binning algorithm for Q

CJ-Wright:
Here is the binning algorithm which correctly bins the detector using its Q resolutions


CJ-Wright:
@ericdill ready for review?

CJ-Wright:
In retrospect my previous two commits should have read:
"flake for the flake god, tests for the test throne"

CJ-Wright:
Also addresses #34

skbeam/core/tests/test_recip.py
```diff
@@ -159,6 +159,16 @@ def test_gisaxs():
...
 
 
+def test_generate_q_bins():
+    rmax = .294
+    qmax = 38
+    pixel_size = .0002
+    distance = .23
+    wavelength = .143
+    a = recip.generate_q_bins(rmax, qmax, pixel_size, distance, wavelength)
+    assert np.any(a)
```
tacaswell: This is not a super good test, as replacing your function with `lambda *args: np.ones(5, 5)` would pass.

tacaswell:
How would this interact with the code in #435

skbeam/core/recip.py
```diff
@@ -342,3 +342,37 @@ def gisaxs(incident_beam, reflected_beam, pixel_size, detector_size,
...
+    Returns
+    -------
+    ndarray:
+        The bin edges, suitable for np.histogram or
+        scipy.stats.binned_statistic
+    """
+    base_pixels = np.arange(0, rmax, pixel_size)
+    pixel_bottom = base_pixels
+    pixel_top = base_pixels + pixel_size
+    tthb = np.arctan(pixel_bottom / distance)
```
tacaswell: can you give these verbose names?  Seems dumb, but it really helps with code maintainability long term.

skbeam/core/recip.py
```diff
@@ -342,3 +342,37 @@ def gisaxs(incident_beam, reflected_beam, pixel_size, detector_size,
...
     return gisaxs_output(alpha_i, theta_f, alpha_f, tilt_angle, qx, qy, qz, qr)
+
+
+def generate_q_bins(rmax, qmax, pixel_size, distance, wavelength):
+    """
+    Generate the Q bins at the resolution of the detector
+    Parameters
+    -----------
+    rmax: float
+        The maximum radial distance on the detector
```
tacaswell: units on this?

skbeam/core/recip.py
```diff
@@ -342,3 +342,37 @@ def gisaxs(incident_beam, reflected_beam, pixel_size, detector_size,
     qr = np.sqrt(qx**2 + qy**2)
 
     return gisaxs_output(alpha_i, theta_f, alpha_f, tilt_angle, qx, qy, qz, qr)
+
+
+def generate_q_bins(rmax, qmax, pixel_size, distance, wavelength):
```
tacaswell: is it worth taking in a `qmin` to not assume that `q==0` is on the detector?

skbeam/core/recip.py
```diff
@@ -342,3 +342,37 @@ def gisaxs(incident_beam, reflected_beam, pixel_size, detector_size,
...
+    """
+    Generate the Q bins at the resolution of the detector
```
tacaswell: This generates non uniform-width bins such that each bin size in q space maps back to a single pixel?

tacaswell:
What does the histogram of q values for a 1kx1k and 2kx2k detector look like?

CJ-Wright:
This code is required for #435. I found that while the mask code works fine on its own it performs much better with the correct Q bins. The masker has an azimuthal integrator inside of it, thus with evenly spaced bins pixels are placed together which should not be together. This causes additional pixels to be masked inappropriately.

skbeam/core/recip.py
CJ-Wright: The units don't matter so long as they are the same as the `pixel_size` and `distance`. I usually use meters, but that is only because that is what pyFAI spits out. Since we eventually take the ratio of the pixel size and the sample to detector distance we only care that they have the same units. I can include units if you want, it just might not be necessary. The function itself returns the bins in the inverse whatever units the wavelength was in.

skbeam/core/recip.py
CJ-Wright: Almost, the function does generate the non uniform-width bins. However, the mapping is to single rings. Dealing with single pixels is hard, as the square nature of the pixels conflicts with the annular nature of the diffraction. For example, a pixel on a 0 degree angle from the point of incidence would have a better resolution than on at 45 degrees, as the 45 degree pixel is on its diagonal which is longer. Technically, this correctly maps the resolution only for pixels which are straight up, down, left or right from the point of incidence. However, the deviation caused by the pixels off axis are small and thus neglected. To properly capture this off axis behavior one might have to use a split pixel method, so that each subsection of a pixel can be put with subpixels at the same Q value. However, Simon's group have shown that this introduces co-variances into the scattering data.

CJ-Wright:
@tacaswell I am not certain what you mean by "histogram of q values". Can you clarify?

tacaswell:
If you just make a histogram of the `q` values for a typical image how quickly does it get uniform?

Sorry, meant to ask about interaction with #430

skbeam/core/recip.py
tacaswell: Just not that this is in 'real' units and not pixel count.

Would it make more sense to take in pixel number given that you also take in pixel size and it is easier to get the pixel shape of your detector (my asking the image how big it is) than it is to get the physical size.

skbeam/core/recip.py
tacaswell: OK, so 'single pixel wide' rings?

tacaswell:
I would also be interested to see a 'number of pixel in ring' vs 'q' plot.

skbeam/core/recip.py
CJ-Wright: Yes

CJ-Wright:
Pixels per ring
pixels_per_ring

CJ-Wright:
@tacaswell 
Pixels per ring, including even binning
And even binning vs q binning
pixels
res

CJ-Wright:
For more info, including analysis of masking effectiveness, please see [Chapter 5](https://github.com/CJ-Wright/Masters_Thesis/blob/master/thesis.pdf)

tacaswell:
To re-phrase this into my pixel-centric-mind working in the limit of an un-tilted detector.

  -  using even Q binning means that the annuli on the array that label a given bin are getting wider in pixel space at high Q because the ΔQ per pixel is going down due to geometric effects.  This can be see in the ~quadratic shape of the number of pixels per ring (the area of a ring goes an `r Δr` and in this case `Δr ∝ r`, thus the area should go as `r*r`)
 - using bins set by the Q resolution of the pixels keeps annuli at fixed width in pixel space (which can be seen that the number of pixels is mostly linear)

The binning has to be done in Q space (not just defining even width rings in pixel space) because real detectors are not perfectly flat so Q would not be constant across all of the pixels in bins defined in pixel space.

This is better than just using a fixed ΔQ of the farthest out pixel because that could lead to under-binning at low Q.

I am thoroughly convinced that this is a good idea in principle, but need to look at the code again.

CJ-Wright:
I think that is mostly correct.
The issue here is more general than the tilt of the detector, since the Q resolution function is non-linear for even flat detectors. This is mostly the effect of the detector being flat and the scattering being, ideally, spherical. Thus the number of pixels subtended by an arbitrary scattering angle changes as a function of the angle. Thus our binning needs to reflect this so that the angles are binned by the changing angular resolution.

CJ-Wright:
@tacaswell So I ran everything in pixel space, and it seems to turn out the same way as using my math for Q space. So as long as you use either even pixel bins (at the pixel diagonal size) or non-linear Q bins things turn out to be the same.

skbeam/core/recip.py
```diff
@@ -342,3 +342,42 @@ def gisaxs(incident_beam, reflected_beam, pixel_size, detector_size,
     qr = np.sqrt(qx**2 + qy**2)
 
     return gisaxs_output(alpha_i, theta_f, alpha_f, tilt_angle, qx, qy, qz, qr)
+
+
+def generate_q_bins(rmax, pixel_size, distance, wavelength, rmin=0):
```
tacaswell: I would put rmin first and not give it a default.

skbeam/core/recip.py
```diff
@@ -342,3 +342,42 @@ def gisaxs(incident_beam, reflected_beam, pixel_size, detector_size,
...
+    """
+    Generate the Q bins at the resolution of the detector
+    Parameters
+    -----------
+    rmax: float
+        The maximum radial distance on the detector in distance units.
+        Note that this should go to the bottom edge of the pixel.
+    pixel_size: float
```
tacaswell: which dimension of pixels?  Do we need to worry about non-square pixels?  Should the resolution be comptued using the height / width or the diagonal?

skbeam/core/recip.py
```diff
@@ -342,3 +342,42 @@ def gisaxs(incident_beam, reflected_beam, pixel_size, detector_size,
...
+        The size of the pixels, in the same units as rmax
+    distance: float
+        The sample to detector distance, in the same units as rmax
+    wavelength: float
```
tacaswell: units?

skbeam/core/tests/test_recip.py
```diff
@@ -159,6 +159,32 @@ def test_gisaxs():
     assert_array_almost_equal(alpha_f_target, g_output.alpha_f[:, 1])
 
 
+def test_generate_q_bins():
+    import scipy.stats as sts
+    pyfai_geo = recip.geo.Geometry(
```
tacaswell: Why do we need to use pyfai here?

skbeam/core/tests/test_recip.py
```diff
@@ -159,6 +159,32 @@ def test_gisaxs():
...
+        detector='Perkin', pixel1=.0002, pixel2=.0002,
+        dist=.23,
+        poni1=.209, poni2=.207,
+        # rot1=.0128, rot2=-.015, rot3=-5.2e-8,
+        wavelength=1.43e-11
+    )
+    r = pyfai_geo.rArray((2048, 2048))
```
tacaswell: use small images so the tests go faster

CJ-Wright:
Actually this PR isn't required, we just need to do everything in radial dimensions and then convert to Q at the end and everything should work out the same.

tacaswell:
I do not understand.

CJ-Wright:
You were correct, if I did the binning in r space and then converted it to Q space the bins work out to be the same

tacaswell:
How does that work if the detector is significantly tilted?  At that point the resolution of the detector is very anisotropic anyway so maybe you have to do something different?

CJ-Wright:
@tacaswell Here is a script which shows that I was correct the first time! The image needs to be processed in Q to begin with.
https://gist.github.com/CJ-Wright/ca3acbb20f780c45c290b29aa5c681fd

danielballan:
@CJ-Wright In that case can you go back and address Tom's line notes above?

CJ-Wright:
@danielballan I think I need to go back to the drawing board for this functionality, I need a better way to assign the per-ring Q resolution.

PR #445, opened 2016-05-17
ENH: Updates to allow sequential processing

stuwilkins:
This is currently a WIP but this is an effort to allow partial and sequential clunking of the data. Please do not merge yet

### DIFF
--- a/skbeam/core/utils.py
+++ b/skbeam/core/utils.py
@@ -950,7 +950,7 @@ def grid3d(q, img_stack,
 
     # call the c library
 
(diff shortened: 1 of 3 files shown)


danielballan:
@stuwilkins It sounds like this could be more valuable now that we have SWMR. Have you had a chance to revisit it?

tacaswell:
@licode can you review this?

licode:
Sure, I will take care of this.

licode:
Though I am a big fan of cython these days.

licode:
I think there might be better way to release and obtain GIL. 

This PR looks good to me in overall.

PR #446, opened 2016-05-23
Binstat

chrisvam:
@CJ-Wright @tacaswell @licode @tjlane Following CJ's comments, it seemed to me that, as an alternative approach to #430, I should take a stab at refactoring scipy.stats.binned_statistic into a class that maintains state (for significant performance boost). I have the feeling it has a mostly richer interface than #430 (e.g. variable size bins, user-defined stats) with a few drawbacks (no explicit mask in the BinnedStatistic constructor: we should probably add it). At the moment I'm just maintaining the binned_statistic interface, to a good approximation.

The code is still a little messy, but perhaps we can discuss to see if this is a better general approach.


skbeam/core/accumulators/binned_statistic.py
```diff
@@ -0,0 +1,445 @@
...
+        Compute a binned statistic for a set of data.
+
+        This is a generalization of a histogram function.  A histogram divides
+        the space into bins, and returns the count of the number of points in
+        each bin.  This function allows the computation of the sum, mean,
+        median, or other statistic of the values within each bin.
+
+        Parameters
+        ----------
+        values : array_like
```
tacaswell: Should this be `x`?

chrisvam: You are right.  I hadn't re-thought the scipy documentation well enough after re-factoring.  I have tried to do that in the most recent commits.

tacaswell:
If this code is very closely extracted from scipy, we need to copy over the scipy copyright with it.

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -0,0 +1,362 @@
...
+
+        super(RadialBinnedStatistic, self).__init__(rpix.reshape(-1),
+                                                    statistic,
+                                                    bins=bins, range=rrange)
+
+    def __call__(self, values):
+        # check for what I believe could be a common error
+        if values.shape != self.expected_shape:
+            raise ValueError('"values" has incorrect shape')
+        return super(RadialBinnedStatistic, self).__call__(values.reshape(-1))
```
tacaswell: I was going to suggest `ravel` but then read the docs just to be sure and it suggests this method http://docs.scipy.org/doc/numpy-1.10.4/reference/generated/numpy.ravel.html

tacaswell:
One gotcha that worries me a bit is 

```python
binner = RadialBinnedStatistic(...)
a = binner(img1)
b = binner(img2)
assert img1 != img2
assert a == b
assert a is b
```

There either needs to be an option to return a copy or a big red `.. warning` box in the docs about this.

tacaswell:
This makes sense to me on the API side.

What sort of performance gain is there over the scipy version?

CJ-Wright:
FYI, I don't know where your bottle necks are with this code but you can parallelize the for loops at lines 178, 185, 247, and 260. I think I have code around that is similar if you want it.

chrisvam:
Tom wrote:

> What sort of performance gain is there over the scipy version?

I did some comparisons as part of the previous pull request and this approach was about 3.5 times faster for a typical LCLS case.  Details are in the thread here that CJ wright started:

https://github.com/scikit-beam/scikit-beam/pull/430

I showed some profiling results in that thread as well, and the majority of time in the scipy approach was spent in numpy.core.multiarray.digitize.

chrisvam:
CJ Wright wrote:

> FYI, I don't know where your bottle necks are with this code but you can parallelize the for loops at lines 178, 185, 247, and 260. I think I have code around that is similar if you want it.

Thanks for the thought, CJ.  At LCLS we typically parallelize using MPI over x-ray laser shots (which arrive at 120Hz).  So, to first order, none of our other cores are available on a node, since they're processing other shots.  So I would vote not to embed parallelization in this code.

chrisvam:
Tom wrote:

> There either needs to be an option to return a copy or a big red .. warning box in the docs about this.

Agreed.  I did that badly when I refactored the code.  I now allocate a new results array for every __call__ in the most recent commit.  That's impressive that you picked that up.

chrisvam:
Tom wrote:

> If this code is very closely extracted from scipy, we need to copy over the scipy copyright with it.

Yes.  I have include that in the most recent commit.

Note that the most recent commit includes some significant changes to address a request from some of our users for both R and Phi binning.  That renders it a work-in-progress still, unfortunately.

chrisvam:
@tacaswell @CJ-Wright @ericdill Hi guys.  FYI, I suggest this is worth looking at again.  I incorporated some suggestions from SLAC people, implementing an R-Phi projection, where the radial projection is a special case with the number of phi-bins set to 1.

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -0,0 +1,508 @@
...
+            bins = [np.asarray(bins, float)]
+
+        if range is not None:
+            if len(range) == 2:
+                range = [range]
+
+        super(BinnedStatistic1D, self).__init__([x], statistic=statistic,
+                                                bins=bins, range=range)
+
+    def __call__(self, values):
```
tacaswell: Do not strictly need this if all you are doing is bouncing up the call chain.

chrisvam: Agreed.  Removed.

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -0,0 +1,505 @@
...
+        """
+        return self.edges
+
+    def __call__(self, values):
+        """
+        Parameters
+        ----------
+        values : array_like
+            The values on which the statistic will be computed.  This must be
+            the same shape as `x` in the constructor.
```
CJ-Wright: What is `x`? It doesn't match any of the names in the `__init__`.

chrisvam: You are correct.  Changed this to "samples" in the most recent commit.  Thanks!

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -0,0 +1,505 @@
...
+            if True, use "cartesian" array ordering, with x corresponding
+            to matrix columns and y corresponding to matrix rows.
+            Otherwise the opposite ("matrix" ordering).
+        xc,yc: int, optional
+            location (in pixels) of origin (default: image center).
+            see "cartesian" parameter for definition of x/y.
+        rrange: (float, float), optional
+            The lower and upper radial range of the bins, in pixels.
+            If not provided, all pixel r values are included.
+        phirange: (float, float), optional
```
tacaswell: Why degrees instead of radians?

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -0,0 +1,505 @@
...
+        Parameters:
+        -----------
+        rbins: int
+            number of radial bins in returned histogram.
+        phibins: int
+            number of phi bins in returned histogram.
+        xsize,ysize: int
+            shape of image in pixels.  see "cartesian" parameter
+            for definition of x/y.
+        cartesian: bool, optional
```
tacaswell: This seems like it will cause more trouble than it is worth.  I would like to adopt the skimage convention and use (row, col) for docs and variable names.

chrisvam: I'd be happy to do that.  Do I understand correctly from your comment that the angle should be arctan(col/row)?

chrisvam: My latest commits now only support matrix-mode, and I think everything is called row/col.  To keep the code clean it's using a feature only available in numpy 1.7: a matrix-indexing version of numpy.meshgrid.

skbeam/core/accumulators/binned_statistic.py
chrisvam: I have switched to radians in the latest commit, although I have the impression that people tend to think more in degrees.

tacaswell: I really did mean this as a question

tacaswell:
Closed by #453

PR #449, opened 2016-07-11
Histogram: Support more types and lists/float/int

chrisvam:
This PR supports more types in the interface.  For user-friendliness it is also more flexible, accepting ints/lists.  The compile time is longer and the .so on my mac is 2.4MB which worries me a bit.  To reduce it to that level I had to enforce the requirement that the x/y coordinates be the same time, which I think isn't too onerous (the weights can still be a different type).


tacaswell:
Closing in favor of #450 which has the same content, but with less un-related commits.

PR #456, opened 2016-08-02
Fit2d save2

CJ-Wright:
Now we can save masks in a file format compatible with fit2d. Take 2


CJ-Wright:
@tacaswell @danielballan should be ready to go. Is 7 commits ok, or do you want them squashed.

danielballan:
Thanks, @CJ-Wright! As we say, build bridges not walls. Maybe this will someday inspire fit2D to join us in the open source community.

PR #457, opened 2016-08-05
Add mask to circular average

ordirules:
This may be redundant but I don't think circular_average supports masks? Is this true, and if so, is it possible to add this feature? I am trying to migrate to skbeam :-)

I know @yugangzhang already has code to do this and I'm sure there are numerous others. If such code exists, please point me to it and ignore this. :)

@danielballan also you might know more about this.

Test of the PR here (and a nosetest included in PR)

``` python
from matplotlib.pyplot import figure, loglog,plot, clf
import numpy as np
from skbeam.core.roi import circular_average

#try something simpler
Nx, Ny = 100, 100
x0,y0 = 50,50
nor = 100
x = np.arange(Nx)-x0
y = np.arange(Ny)-y0
X,Y = np.meshgrid(x,y)
R = np.sqrt(X**2 + Y**2)
IMG = R**2
mask = np.ones_like(IMG)
IMG2 = np.copy(IMG)
IMG2[10:20,10:20] = 1e5
mask[10:20,10:20] = 0#mask some region
r,I = circular_average(IMG,(y0,x0), nx=nor)
rmasked,Imasked = circular_average(IMG2,(y0,x0), nx=nor, mask=mask)
rbad,Ibad = circular_average(IMG2,(y0,x0), nx=nor)


figure(2);clf();
loglog(r,I,color='k')
loglog(rmasked,Imasked,color='r')

#plot percentage error just to see how masking affects circavg
figure(3);clf();
plot(np.abs(I-Imasked)/I)
```

### DIFF
--- a/skbeam/core/roi.py
+++ b/skbeam/core/roi.py
@@ -399,7 +399,7 @@ def mean_intensity(images, labeled_array, index=None):
 
 
 def circular_average(image, calibrated_center, threshold=0, nx=100,
-                     pixel_size=(1, 1),  min_x=None, max_x=None):
+                     pixel_size=(1, 1),  min_x=None, max_x=None, mask=None):
     """Circular average of the the image data
     The circular average is also known as the radial integration
     Parameters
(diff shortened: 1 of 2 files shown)


ordirules:
I just checked git blame, I think this might be more appropriate to include @sameera2004 and @tacaswell.

(Also, Travis failed, is that normal?)

CJ-Wright:
Have you looked at #429? Also auto masking is coming see #435.

ordirules:
This is what I was looking for. Thanks, I'll read later and follow discussion from there :+1:

PR #459, opened 2016-08-08
ENH: better circular average

CJ-Wright:
Now supporting masks and any statistic desired. Also uses auto binning if bins not stated.

### DIFF
--- a/skbeam/core/tests/test_utils.py
+++ b/skbeam/core/tests/test_utils.py
@@ -43,12 +43,17 @@ from numpy.testing import (assert_array_equal, assert_array_almost_equal,
                            assert_almost_equal)
 from nose.tools import assert_equal, assert_true, raises
 
+from skbeam.testing.decorators import known_fail_if
+
 import skbeam.core.utils as core
 
 import logging
+
 try:
     from pyFAI.geometry import Geometry
+    pf = True
 except ImportError:
+    pf = False
     pass
 logger = logging.getLogger(__name__)
 
@@ -153,7 +158,7 @@ def test_bin_edges():
         # 2 entries
         {'range_min': 1.234, 'step': np.pi / 10},
         # 1 entry
-        {'range_min': 1.234, },
+        {'range_min': 1.234,},
         # max < min
         {'range_max': 1.234, 'range_min': 5.678, 'step': np.pi / 10},
         # step > max - min
@@ -182,8 +187,8 @@ def test_grid3d():
     # slice tricks
     # this make a list of slices, the imaginary value in the
     # step is interpreted as meaning 'this many values'
-    slc = [slice(_min + (_max - _min)/(s * 2),
-                 _max - (_max - _min)/(s * 2),
+    slc = [slice(_min + (_max - _min) / (s * 2),
+                 _max - (_max - _min) / (s * 2),
                  1j * s)
            for _min, _max, s in zip(q_min, q_max, dqn)]
     # use the numpy slice magic to make X, Y, Z these are dense meshes with
@@ -224,8 +229,8 @@ def test_process_grid_std_err():
     # slice tricks
     # this make a list of slices, the imaginary value in the
     # step is interpreted as meaning 'this many values'
-    slc = [slice(_min + (_max - _min)/(s * 2),
-                 _max - (_max - _min)/(s * 2),
+    slc = [slice(_min + (_max - _min) / (s * 2),
+                 _max - (_max - _min) / (s * 2),
                  1j * s)
            for _min, _max, s in zip(q_min, q_max, dqn)]
     # use the numpy slice magic to make X, Y, Z these are dense meshes with
@@ -244,12 +249,12 @@ def test_process_grid_std_err():
     # check the values are as expected
     npt.assert_array_equal(mean,
                            np.ones_like(X) * np.mean(np.arange(1, 101)))
-    npt.assert_array_equal(occupancy, np.ones_like(occupancy)*100)
+    npt.assert_array_equal(occupancy, np.ones_like(occupancy) * 100)
     # need to convert std -> ste (standard error)
     # according to wikipedia ste = std/sqrt(n)
     npt.assert_array_almost_equal(std_err,
                                   (np.ones_like(occupancy) *
-                                   np.std(np.arange(1, 101))/np.sqrt(100)))
+                                   np.std(np.arange(1, 101)) / np.sqrt(100)))
 
 
 def test_bin_edge2center():
@@ -272,7 +277,7 @@ def test_small_verbosedict():
         assert_equal(eval(six.text_type(e)), expected_string)
     else:
         # did not raise a KeyError
-        assert(False)
+        assert (False)
 
 
 def test_large_verbosedict():
@@ -293,7 +298,7 @@ def test_large_verbosedict():
         assert_equal(eval(six.text_type(e)), expected_sting)
     else:
         # did not raise a KeyError
-        assert(False)
+        assert (False)
 
 
 def test_d_q_conversion():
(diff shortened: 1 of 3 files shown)


ordirules:
@CJ-Wright  @danielballan 
For nomenclature purposes, would ``bin2D`` or ``bin_grid`` be more appropriate? It's basically binning, but assuming pixels are on a grid and (potentially) masked. Sounds like it has a much broader reach.

danielballan:
I like `bin_grid`.

CJ-Wright:
ok, bin_grid it is. I am still working out some of the corner cases for the auto-binning. Currently I have the binning set up like this:
The radial value for each pixels is computed at the center of the pixel.

Thus the farthest value for the `np.arange` is `np.max(r) + pixel_resolution`/2., where `pixel_resolution` is equal to the length of the diagonal of the pixel.

The step size is the `pixel_resolution`

The starting bin value is the problem, for the tests I was running `np.min(r) - pixel_resolution/2.` worked fine, but it does not work when the beam is in the center of the pixel. It seems that the first bin is a special case, where the bin size needs to change depending on where the beam center is.
eg, if the center is at a corner then the above equation is fine, since each pixel center is half a pixel away from the center. But if the center falls exactly in the center of the pixel, then `np.min(r)` is 0.0, thus leading the above equation to give a negative radius bin which seems very unmathematical.

On the other hand this might be ok, as when we do the bin_edges to centers the bin will be 0.0, the correct r value for the pixel.
Thoughts?

ordirules:
Maybe give up the fact that the step size is the resolution? Would it matter that we match that exact value?
You could choose some number of steps that's close enough. Note that for a circular average, you can easily choose sub pixel steps and still have enough points to average (for larger radius).

I'd say either just do `np.max(np.min(r_array) - res/2.,0)` (slightly modifies step size) or choose some default step size and use `np.linspace` instead.

skbeam/core/roi.py
```diff
@@ -398,50 +398,55 @@ def mean_intensity(images, labeled_array, index=None):
...
-    return bin_centers, ring_averages
+    # If there is no mask, make one
+    if mask is None:
+        mask = np.ones(image.shape, dtype=int).astype(bool)
+
+    if bins is None:
+        res = np.sqrt(np.sum([a ** 2 for a in pixel_size]))
+        bins = np.arange(np.min(r_array) - res/2., np.max(r_array) + res/2.,
+                         res)
+    bin_edges, int_stat, bin_num = sts.binned_statistic(r_array[mask],
```
ordirules: don't forget `import scipy.stats as sts`
That does bring a point. Should scipy be used as a dependency here?
We could have our own binning routines using numpy (basically digitizing and then bincounting etc)
@danielballan you know more about the philosophy of dependencies

CJ-Wright: Ah, yes I did forget that import, I think having scipy as a dep is fine, I mean it is part of the scientific python stack and it might even be a dep that is included in the conda recipe already.

I would prefer to not have to rewrite functions that already exist unless we are expecting that we can do it much better (faster, lower memory, etc.). Although I do know that a class version of binned_statistic is in the pipeline. We could expose a hook for this in the signature if we want.

CJ-Wright:
The reason I have been so persnickety about the binning is that I have been running some trials on automated masking and integration. It seems that if you pick different binnings you get oscillation in the standard deviation of the data. The oscillation is due to some of the data points being put into a different bin than they should be in. While this doesn't cause too many issues with the data, usually, when I go to use the statistics/binning to create a mask the miss-binning of the pixels causes the masker to erroneously remove pixels.

I've found that the best binning, which minimizes the standard deviation oscillation, is at the resolution of the least resolved pixels. These least resolved pixels are on the diagonal, thus the resolution equation above. It seems however, that upon additional testing the above equations give back the correct result including non-zero first bin values (after using `bin_edges_to_centers`).

ordirules:
If I understand correctly, you're saying plotting the standard deviation along a ring is greatly affected by the bin choice even up to pixel resolution? That sounds like an interesting result worth mentioning.

Do you have sample plots of the standard deviation per ring for different bin choices? I think it would be nice to document this somewhere (or refer to some paper etc in the comments) to justify the choice so a future contributor doesn't "improve" your code in the wrong way.

@danielballan may know more about how in scikit-beam to document routines that have some subtle scientific motivation.

CJ-Wright:
I do have some truly great graphs but this margin is too small to contain them. (actually I'm away from my computer for a few days but I do have the graphs on it so when I get back to my computer I'll upload them) (with apologies to Fermat)

ordirules:
ok sounds good, if you're in the process of writing anything up about it though, the PR can wait :-)
sounds like a nice result

skbeam/core/roi.py
danielballan: I would expect `binned_statistic` to be faster (and, arguably, more readable) that a custom-rolled solution using numpy. Our philosophy is: restrict package-wide deps as much as possible, but let individual modules require whatever they need to leverage trusted code from the scipy community (like scipy itself) and the X-ray software community (like xraylib).

CJ-Wright:
@ordirules Here are some images, note that for the simulated image, each pixel value was equal to the radial distance from the center.

binning_example

skbeam/core/roi.py
```diff
@@ -398,50 +398,55 @@ def mean_intensity(images, labeled_array, index=None):
...
-
-    bin_centers = utils.bin_edges_to_centers(bin_edges)[th_mask]
-
-    return bin_centers, ring_averages
+    # If there is no mask, make one
+    if mask is None:
+        mask = np.ones(image.shape, dtype=int).astype(bool)
+
+    if bins is None:
+        res = np.sqrt(np.sum([a ** 2 for a in pixel_size]))
```
tacaswell: `np.hypot`

skbeam/core/roi.py
```diff
@@ -398,50 +398,55 @@ def mean_intensity(images, labeled_array, index=None):
     return mean_intensity, index
 
 
-def circular_average(image, calibrated_center, threshold=0, nx=100,
-                     pixel_size=(1, 1),  min_x=None, max_x=None):
+def circular_average(image, r_array, pixel_size=(1, 1), statistic='mean',
```
tacaswell: This is completely changing the API on this function.  This needs documentation + deprecation cycle.

tacaswell:
The oscillations look like they are driven by there being small numbers of pixels in the ring because you are way out at the corners of the image which is not super surprising.

CJ-Wright:
I thought of the oscillation as being caused by the beating of the bin resolution against the detector resolution. Thus every once in a while a bin will have more or less pixels than expected because pixels which would belong to a different bin were put into this bin. Since the erroneously binned ring has different values than the rest of the ring the STD is larger.

ordirules:
I'm confused with the counts plot. Shouldn't the integral of each curve yield the total number of pixels and hence be the same. Are you possibly overcounting pixels? I don't think binning should ever overcount. That might be the issue maybe?

skbeam/core/roi.py
ordirules: this will be a new function @CJ-Wright maybe change the name now to avoid confusion?

CJ-Wright:
Yes, `scipy.stats.binned_statistic` only counts the pixels once.

skbeam/core/roi.py
CJ-Wright: Yes this is getting moved to `bin_grid` as soon as I can commit.

tacaswell:
@CJ-Wright Yes, we are saying the same thing. 

I think the plan for this should be:

 1) restore the API on this function
 2) create a new function with this signature
 3) change the default `nx` to `None` and then use this logic to select the bin count (instead of 100).  That is an API break, but it is small, an improvement and I think it is justifiable

CJ-Wright:
@tacaswell Do you want this `bin_grid` function to go into `roi` or is there a better place?
What is the best way to rewind this branch back to master?

skbeam/core/tests/test_utils.py
```diff
@@ -178,8 +187,8 @@ def test_grid3d():
     # slice tricks
     # this make a list of slices, the imaginary value in the
     # step is interpreted as meaning 'this many values'
-    slc = [slice(_min + (_max - _min)/(s * 2),
```
tacaswell: can you make the `//` as long as you are touching it?

tacaswell: The same for most of the rest of these?

skbeam/core/utils.py
```diff
@@ -882,8 +884,8 @@ def grid3d(q, img_stack,
         from ..ext import ctrans
     except ImportError:
         raise NotImplementedError(
-            "ctrans is not available on your platform. See"
```
tacaswell: good catch!

CJ-Wright:
@danielballan Ready to go!

skbeam/core/tests/test_utils.py
CJ-Wright: I just touched it because the auto pep8 in pycharm, sure

skbeam/core/utils.py
CJ-Wright: Thanks, I got the error myself so I thought I'd fix it

skbeam/core/tests/test_utils.py
CJ-Wright: Actually that caused errors

CJ-Wright: I reverted the branch

danielballan: hmm -- worth opening an issue to revisit that then

CJ-Wright: Here is the [broken build](https://travis-ci.org/CJ-Wright/scikit-beam/builds/154831704)

skbeam/core/utils.py
```diff
@@ -1215,6 +1217,51 @@ def geometric_series(common_ratio, number_of_images, first_term=1):
...
     geometric_series = [first_term]
 
-    while geometric_series[-1]*common_ratio < number_of_images:
-        geometric_series.append(geometric_series[-1]*common_ratio)
+    while geometric_series[-1] * common_ratio < number_of_images:
+        geometric_series.append(geometric_series[-1] * common_ratio)
     return geometric_series
+
+
+def bin_grid(image, r_array, pixel_sizes, statistic='mean', mask=None,
```
ordirules: Sorry @CJ-Wright  I just caught this. Why is `pixel_sizes` a mandatory parameter? If `bins` is specified, it shouldn't be necessary. Perhaps you should make a change? I'm also fine to do it if you're okay with it.

Fortunately, this can be changed to 
```python
bin_grid(image, r_array, pixel_sizes=None, statistic='mean'
```
without breaking the API.

CJ-Wright: Be my guest! I'd do it myself but am very bandwidth limited at the moment. You might note that either the bins or the pixel size must be given. (This really should be refactored though, in light of the new binned statistic classes)

ordirules: yes, I was thinking actually maybe superseding this by adding a `rpix` parameter to the `RadialBinnedStatistic`. Then maybe refactor your function to just do the autobinning, which isn't done anywhere else and essential. Also a little busy, but should submit something for discussion later.

PR #461, opened 2016-08-25
ENH: move to integer division (caused nan in the wrong place error previously)

CJ-Wright:


CJ-Wright:
@tacaswell These were the changes you requested previously to the tests. I put them here so that their issues could be addressed.

tacaswell:
Ah, I gravely misunderstood what was going on here.  Sorry for the noise.

Closing because I was wrong :sheep:

PR #462, opened 2016-08-26
Masking added to circular average as an option

ordirules:
@yugang is a little busy, here is the update after discussing with him. 
I forked his branch into my repository and made some modifications.

@CJ-Wright @danielballan @tacaswell  (please include any others)

This is regarding PR #459 

Changes:
- removed bin_1D as @sameera2004 's suggestion.
- removed an else statement 
- added a test for the masking

### DIFF
--- a/skbeam/core/roi.py
+++ b/skbeam/core/roi.py
@@ -425,7 +425,7 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
         Right edge of last bin defaults to maximum value of x
     mask : mask for 2D data. Assumes 1 is non masked and 0 masked.
         None defaults to no mask.
-    
+
     Returns
     -------
     bin_centers : array
--- a/skbeam/core/roi.py
+++ b/skbeam/core/roi.py
@@ -456,7 +456,7 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
 
     bin_centers = utils.bin_edges_to_centers(bin_edges)[th_mask]
 
-    return bin_centers, ring_averages 
+    return bin_centers, ring_averages
 
 
 def kymograph(images, labels, num):
--- a/skbeam/core/roi.py
+++ b/skbeam/core/roi.py
@@ -397,6 +397,53 @@ def mean_intensity(images, labeled_array, index=None):
         mean_intensity[n] = ndim.mean(img, labeled_array, index=index)
     return mean_intensity, index
 
+
+
+def bin_1D(x, y, nx=None, min_x=None, max_x=None):
+    """
+    Bin the values in y based on their x-coordinates
+
+    Parameters
+    ----------
+    x : array
+        position
+    y : array
+        intensity
+    nx : integer, optional
+        number of bins to use defaults to default bin value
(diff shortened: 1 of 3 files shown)


skbeam/core/roi.py
```diff
@@ -423,15 +423,30 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
...
+    See Also
+    --------
+    bad_to_nan_gen : Create a mask with np.nan entries
+
+    Notes
+    -----
     """
     radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size)
 
+    if mask is not None:
```
CJ-Wright: I would go with:
```python
if not mask:
    mask = np.ones(image.shape)
image = image[mask]
radial_val = radial_val[mask]
```
This should take care of everything, and then you don't need the `np.ravel`s in the code below.

ordirules: thanks, I like the idea of using bools. Just to make sure, I ran a timing test, which is okay except for one detail. Will write in comment.

ordirules:
About changing the accessing to a boolean access (`w=np.where(mask==1);img[w].ravel()` versus `mask = np.ones(img.shape);img[mask];`, it seems that the code is sped up with masking, but slowed down 10% without masking.

In terms of timing, I think this suggests that `np.ones` < `np.where`  < `img[mask]`.
The data used is:
```python
N = 1000
r0 = N/2, N/2
img = np.random.random((N,N))
# toggle between two
mask = np.random.random((N,N)) > .5
# or else used mask=None
```
And timings are:
```python 
Time circavg1 (nomask), old indexing: 0.12648274898529052
Time circavg2 (nomask), bool indexing: 0.13155609369277954
Time circavg2 (with mask), old indexing: 0.0859883451461792
Time circavg2 (with mask), bool indexing: 0.08157441377639771
```
I'm tempted to revert back to not initializing an array of 1's in the mask and just use `np.where`, although it seems clean. Any arguments?

CJ-Wright:
Personally, a 10ms difference isn't that big of a deal to me but I'm leave it up to you.

skbeam/core/roi.py
```diff
@@ -423,15 +423,33 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
...
+    See Also
+    --------
+    bad_to_nan_gen : Create a mask with np.nan entries
+
+    Notes
+    -----
     """
     radial_val = utils.radial_grid(calibrated_center, image.shape, pixel_size)
 
+    if mask is None:
```
tacaswell: I do not think you need this as you only slice into the data if the mask is non-trivial.

tacaswell:
You appear to be failing on flake8 tests

ordirules:
@CJ-Wright it's not just 10ms, it appears to be 10%. It also appears to be non-linear. I ran for a larger array (N=4000) and it went up to a 20% difference. I think it's safe not to add extra memory declarations if not necessary (`ravel()` doesn't cost anything).

ordirules:
@tacaswell thanks

PR #463, opened 2016-08-28
Add interpolation feature for masked circular averaged images

ordirules:
(mentions: @CJ-Wright , @danielballan, @tacaswell)
It's just one extra step to fill in the image with the circular averaged values.
Would it be appropriate to add it to `circular_average`?

Here is my notebook example (are there better ways to share notebooks? I've searched but could not find appropriate ones):
https://github.com/ordirules/notebooks/blob/master/circavg_SIMGtest.ipynb

Another option could be to have SIMG be a bool and the function to return a tuple of three values (bin centers, values,  IMG_interpolated). I prefer the first.

This is a useful feature for me (and I think XPCS, which needs an image to normalize from, @yugangzhang ). Any thoughts? thanks!


ordirules:
I'll add the test a little later to satisfy codecov.

tacaswell:
I would rather a `construct_circ_avg_image(radius, average_intensity, center)` function.

In general I think it is better to have more simple, easy to understand functions than fewer functions with a large number of options.

CJ-Wright:
Do you need to do an interpolation of the data? Can you just go with `np.digitize` to decide which pixels correspond to which values?

ordirules:
@CJ-Wright I figured interp would provide finer resolution in the case that some points are masked.

@tacaswell I agree that sounds like a better idea. Will work on that soon

ordirules:
here's the new code and an example here, but it is by no means finished. I need to test a few cases (including anisotropic pixels), and what happens to values outside the boundaries. Maybe document the edge cases a little better.

notebook updated here:
https://github.com/ordirules/notebooks/blob/master/circavg_SIMGtest.ipynb

feel free to give feedback on the API and if there's a more efficient means to go about this. This code would definitely be useful for me.

thanks for the previous feedback and input!

skbeam/core/roi.py
```diff
@@ -458,6 +459,67 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
...
+        Normally, data will be taken from circular_average and used to
+        re-interpolate into an image.
+
+    Parameters
+    ----------
+    radii : 1D array of floats
+        the radii (must be in pixels)
+    intensities : 1D array of floats
+        the intensities for the radii
+    centerdims : 4 tuple of floats, optional
```
tacaswell: I would split up the center and the shape as inputs.

skbeam/core/tests/test_roi.py
```diff
@@ -318,22 +318,30 @@ def test_circular_average():
...
+    simg = np.zeros_like(image)
+
     bin_cen_masked, ring_avg_masked = roi.circular_average(image,
                                                            calib_center,
                                                            min_x=0,
                                                            max_x=10,
                                                            nx=6,
-                                                           mask=mask)
+                                                           mask=mask,
+                                                           SIMG=simg)
```
tacaswell: This is probably what is failing the tests?

ordirules: sorry, forgot to update the tests. I did but I still need to check this works with anisotropic pixels later, tonight.. thanks!

skbeam/core/roi.py
ordirules: yeah it's more sensible, thanks. both are still required, if one is missing the defaults are used for both

skbeam/core/roi.py
```diff
@@ -458,6 +459,67 @@ def circular_average(image, calibrated_center, threshold=0, nx=100,
...
+
+    Notes
+    -----
+    Some pixels may not be filled if the dimensions chosen are too large.
+        Run this code again on a list of values equal to 1 to obtain a mask.
+
+    Example
+    -------
+    """
+    if center is None or dims is None:
```
tacaswell: I think you need to handle these cases differently.  If I specify `dims` and not `center` my `dims` input should not be discarded.

ordirules: that makes sense. `center` but no `dims` does not however. I'll raise an exception there just in case.

ordirules:
ok done and tested for anisotropic pixels, added a test (asserting two values I previously numerically computed and visually verified are equal).

I fooled around a little to try to be sure it's working properly, here:
https://github.com/ordirules/notebooks/blob/master/circavg_SIMGtest.ipynb

Should something like this go to scikit-beam examples?

ordirules:
hmm one more thought. there remains another question. What should values out of bounds be interpolated to? In my opinion, I think the higher bound should be set to zero whereas the lower bound should be set to be equal to the value at the smallest radius. This is currently how it's set with the line:
```CIMG = np.interp(radial_val, radii, intensities, right=0)```

and that is how the code is run in note book

(I'll satisfy codecov later tomorrow, I thought I was covering all the code, does it require trying all permutations of parameters? that can be quite lengthy)

**Comment by ordirules** [comment]
one more thought: this might be more appropriate for image.py no?

**Comment by ordirules** [comment]
made two changes: interpolation of edges can be customized by user (and now set to 0 by default)
also, moved `construct_circ_avg_image` to image.py. This is suggested and I just have to rollback one commit to go back to original idea. This comment supercedes previous comments

File: skbeam/core/image.py [hunk 4]
```diff
@@ -96,3 +97,77 @@ def _corr_ax1(input_image):
...
+
+    Notes
+    -----
+    Some pixels may not be filled if the dimensions chosen are too large.
+        Run this code again on a list of values equal to 1 to obtain a mask
+        (and set left=0 and right=0).
+    """
+    if dims is None:
+        if center is not None:
+            print("Error: Specifying a dims but not a center does not make"
```
danielballan: Why not:

```python
raise ValueError("Specifying a dims but not a center does not make "
                 "sense and may lead to unexpected results.")

danielballan:
Yeah, `image.py` seems like a good home for it. I left a minor line note -- otherwise this looks good to me.

danielballan:
@tacaswell is +1 on merging as well. Address my line note (or not) as your discretion and self-merge this PR at will.

ordirules:
thanks, made change (and verified with my notebook and nosetest), waiting for Travis etc

I will keep the verbose line of commits as is (including the move from roi to image), if you prefer them squashed, let me know.

skbeam/core/image.py
ordirules: that looks much neater thanks!

ordirules:
@danielballan @tacaswell Travis failed on histogram.
Is there a way to restart Travis?

This seems like an ongoing issue but as far as we've seen histogram and numpy's histogram give the same results. Maybe the test in core/accumulators/tests/test_histogram.py should have the random values seeded to be safe?

ordirules:
@danielballan going to merge. removed some edits in roi.py and test_roi.py since they were out of context (and didnt affect flake 8). in consequence, Travis ran again and succeeded

danielballan:
Great. No need to squash commits. To answer your question, you can re-run Travis. There is a button in the upper right of the build log page. Merge at will.

PR #464, opened 2016-08-28
Image filling using Gaussian smoothing

ordirules:
(Mentions: @danielballan @tacaswell)
This is more an aesthetic feature. Fill an image using Gaussian smoothing (see notebook reference below for quick transfer of information through an image).

Will later add savitzky golay and circular average filling (basically, reinterpolating circular average and sampling from Poisson distribution to make it blend in) a little later. I figure it might be best to add in things step by step. (If this is something useful)

This does two main things:
1. Fills in the image by 'bleeding' in values from neighboring pixels using Gaussian smoothing.
2. Uses (if desired) the intensity values that 'bled' in as a sample to the Poisson distribution. This makes the 'bled' in pixels blend better with the image. Allows for Poisson distribution and quasi-Poisson where the images are actually the average of a sequence of images averaged (of which the statistics came from a Poisson distribution).

Future thoughts: could possibly include multinomial distribution for undersampled speckle (when the pixels obey speckle statistics but they are smaller than a pixel).

Examples (see my notebook):
https://github.com/ordirules/notebooks/blob/master/gaussianfill-example.ipynb
(note the smaller gaps fill well whereas the larger ones do not, for obvious reasons. Both cases are meant to sort of display the limits of this)

Future method plans: circular average filling, Savitzky-Golay filling.

Will also add a test to this later.

### DIFF
--- a/skbeam/core/image.py
+++ b/skbeam/core/image.py
@@ -44,7 +44,7 @@ import logging
 from scipy.ndimage.filters import gaussian_filter
 logger = logging.getLogger(__name__)
 
-from core.stats import poissonize
+from skbeam.core.stats import poissonize
 
 
 def find_ring_center_acorr_1D(input_image):
--- a/skbeam/core/image.py
+++ b/skbeam/core/image.py
@@ -41,6 +41,7 @@ from __future__ import absolute_import, division, print_function
 import numpy as np
 from . import utils
 import logging
+from scipy.ndimage.filters import gaussian_filter
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


ordirules:
codecov wasn't happy. A piece of my function generates random values. I added a test for this but seeded numpy's random generator. Is this the common practice?

tacaswell:
Yes, seeding the RNG is standard practice.

tacaswell:
How does this relate to http://scikit-image.org/docs/dev/auto_examples/filters/plot_inpaint.html ?

ordirules:
Thanks, I hadn't heard of this. Just tried it. Their code is more complex and performs better (see notebook linked below comparing the two). I chose the sigma that gave the best results from my code and it's nowhere near their result.
However, I'm trying to run their code on our images (4 megapixels). I left it running for 15 min and still hasn't completed... So I'm not sure how it compares with conventional xray scattering data just yet...

In the meantime, I think this code could be a practical compromise. Perhaps the first piece could be proposed into scikit image and the second piece (adding Poisson noise) be put into scikit beam?


notebook: https://github.com/ordirules/notebooks/blob/master/skimage_inpaint_gaussfill_compare.ipynb

tacaswell:
I do not quite understand the use-case for this function.

ordirules:
The earlier notebook I linked shows the motivation:
https://github.com/ordirules/notebooks/blob/master/gaussianfill-example.ipynb

However, this remains an aesthetic technique with no quantitative use. What makes it different from the skimage routine is it's much faster (though less precise) and allows to simulate shot noise.

I think this is a question of whether it's compatible with the skbeam philosophy. Maybe given it's an aesthetic technique it should be something rather for skimage. What do you think?

skbeam/core/image.py
```diff
@@ -96,3 +98,126 @@ def _corr_ax1(input_image):
...
+        To obtain new mask (not all pixels may be filled), it is
+        recommended to run the mask through this.
+    """
+    imgf = np.copy(img)*mask
+    imgg = gaussian_filter(img*mask, sigma)
+    imgmsk = gaussian_filter(mask*1., sigma)
+    imgmskrat = gaussian_filter(mask*0.+1, sigma)
+    w = np.where((mask < .1)*(imgmsk > 0))
+    imgf[w] = imgg[w]/imgmsk[w]*imgmskrat[w]
+    if poisson:
```
danielballan: Let's extract this as a separate function.

danielballan:
Conclusion from in-person meeting:

We'll merge this after the Poisson stuff is moved into a separate function and then investigate overlap with scikit-image and possible push this upstream to them.

ordirules:
Separated functions and rebased.

`poissonize` is just a wrapper function that makes sure `np.random.poisson` does not crash. `np.random.poisson` raises a ValueError if integer values supplied are too high. Honestly, I'd vote not to include this function into scikit beam since it's to first order a wrapper function of numpy (minus the ValueError circumvention). (Also, adding `poissonize` would suggesting adding `random_normal` et al )

For Gaussian filling, I would suggest it should come with an example. I cannot honestly find an application for it myself. (All my data is isotropic, so filling with some circular average to estimate values is superior over smoothing)
After this, it becomes simply an aesthetic technique, which may not be useful for scikit-beam.

So, I'm fine to close this, and maybe PR in the future if there is an application.
I'll keep it open for suggestions before doing so.

thanks @danielballan !

PR #466, opened 2016-09-11
Generalize binning

chrisvam:
As I discussed previously with Tom, I realized that the binning specification in the RPhiBinnedStatistic and RadialBinnedStatistic was not as flexible as what was underneath in the scipy binned_statistic code.  This PR tries to address that.  Over the course of doing this, it became clear that my inheritance wasn't structured correctly, so I tried to fix that.

There are three things that seem non-ideal to me:

- the way I implement masking (which is not available in scipy.binned_statistic) is imperfect, but it's the best I can see at the moment.  see comments in code
- there are about 10 lines of code that are duplicated in both RPhiBinnedStatistic and RadialBinnedStatistic.  I might be able to fix this with some inheritance or fancy python tricks
- I should test RPhiBinnedStatistic using scipy.binned_statistic_2d

But this PR improves the interface, which is important.

p.s. this time I started out with a fresh fork/clone and did the "magic 4 commands" that Tom sent me to create the feature-branch.  So hopefully there will not be git issues this time.  Learning.

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -397,9 +397,15 @@ class BinnedStatistic2D(BinnedStatisticDD):
         return super(BinnedStatistic2D, self).__call__(values)
 
 
-def get_r_phi(rowsize, colsize, rowc, colc):
-    rowc = rowsize//2 if rowc is None else rowc
-    colc = colsize//2 if colc is None else colc
+def get_r_phi(shape, origin):
+    rowsize = shape[0]
+    colsize = shape[1]
+    if origin is None:
+        rowc = rowsize//2
+        colc = colsize//2
+    else:
+        rowc = origin[0]
+        colc = origin[1]
     row = np.arange(rowsize)-rowc
     col = np.arange(colsize)-colc
     # meshgrid indexing='ij' option requires numpy 1.7 or later
@@ -415,13 +421,13 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
     image in both radius and phi.
     """
 
-    def __init__(self, rowsize, colsize, bins=10, range=None,
-                 rowc=None, colc=None, mask=None, statistic='mean'):
+    def __init__(self, shape, bins=10, range=None,
+                 origin=None, mask=None, statistic='mean'):
         """
         Parameters:
         -----------
-        rowsize,colsize: int
-            shape of image in pixels.
+        shape: tuple of ints of length 2.
+            shape of image.
         bins : int or [int, int] or array_like or [array, array], optional
             The bin specification:
             * number of bins for the two dimensions (nr=nphi=bins),
@@ -437,9 +443,9 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
             [[rmin, rmax], [phimin, phimax]]. All values outside of this range
             will be considered outliers and not tallied in the histogram.
             See "bins" parameter for definition of phi.
-        rowc,colc: int, optional
+        origin: tuple of ints with length 2, optional
             location (in pixels) of origin (default: image center).
-        mask: 2-dimensional np.ndarray, optional
+        mask: 2-dimensional np.ndarray of ints, optional
             array of zero/non-zero values, same shape as image used
             in __call__.  zero values will be ignored.
         statistic : string or callable, optional
@@ -460,7 +466,7 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
                 will be called on the values in each bin.  Empty bins will be
                 represented by function([]), or NaN if this returns an error.
         """
-        rpix, phipix = get_r_phi(rowsize, colsize, rowc, colc)
+        rpix, phipix = get_r_phi(rowsize, colsize, origin)
 
         self.expected_shape = rpix.shape
         if mask is not None:
@@ -491,13 +497,13 @@ class RadialBinnedStatistic(BinnedStatistic1D):
     image in both radius and phi.
     """
 
-    def __init__(self, rowsize, colsize, bins=10, range=None,
-                 rowc=None, colc=None, mask=None, statistic='mean'):
+    def __init__(self, shape, bins=10, range=None,
+                 origin=None, mask=None, statistic='mean'):
         """
         Parameters:
         -----------
-        rowsize,colsize: int
-            shape of image in pixels.
+        shape: tuple of ints of length 2.
+            shape of image.
         bins : int or sequence of scalars, optional
             If `bins` is an int, it defines the number of equal-width bins in
             the given range (10 by default).  If `bins` is a sequence, it
@@ -509,9 +515,9 @@ class RadialBinnedStatistic(BinnedStatistic1D):
             The lower and upper range of the bins.  If not provided, range
             is simply ``(x.min(), x.max())``.  Values outside the range are
             ignored.
-        rowc,colc: int, optional
+        origin: tuple of ints with length 2, optional
             location (in pixels) of origin (default: image center).
-        mask: 2-dimensional np.ndarray, optional
+        mask: 2-dimensional np.ndarray of ints, optional
             array of zero/non-zero values, same shape as image used
             in __call__.  zero values will be ignored.
         statistic : string or callable, optional
@@ -532,7 +538,7 @@ class RadialBinnedStatistic(BinnedStatistic1D):
                 will be called on the values in each bin.  Empty bins will be
                 represented by function([]), or NaN if this returns an error.
         """
-        rpix, _ = get_r_phi(rowsize, colsize, rowc, colc)
+        rpix, _ = get_r_phi(shape, origin)
         self.expected_shape = rpix.shape
 
         if mask is not None:
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -385,6 +385,18 @@ class BinnedStatistic2D(BinnedStatisticDD):
         return super(BinnedStatistic2D, self).__call__(values)
 
 
+def get_r_phi(rowsize, colsize, rowc, colc):
+    rowc = rowsize//2 if rowc is None else rowc
+    colc = colsize//2 if colc is None else colc
+    row = np.arange(rowsize)-rowc
+    col = np.arange(colsize)-colc
+    # meshgrid indexing='ij' option requires numpy 1.7 or later
+    rowgrid, colgrid = np.meshgrid(row, col, indexing='ij')
+    rpix = np.sqrt(rowgrid**2 + colgrid**2)
+    phipix = np.arctan2(colgrid, rowgrid)
+    return rpix, phipix
+
+
 class RPhiBinnedStatistic(BinnedStatistic2D):
     """
     Create a 2-dimensional histogram by binning a 2-dimensional
@@ -436,27 +448,14 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
                 will be called on the values in each bin.  Empty bins will be
                 represented by function([]), or NaN if this returns an error.
         """
+        rpix, phipix = get_r_phi(rowsize, colsize, rowc, colc)
 
-        rowc = rowsize//2 if rowc is None else rowc
-        colc = colsize//2 if colc is None else colc
-        row = np.arange(rowsize)-rowc
-        col = np.arange(colsize)-colc
-        # meshgrid indexing='ij' option requires numpy 1.7 or later
-        rowgrid, colgrid = np.meshgrid(row, col, indexing='ij')
-        self.expected_shape = rowgrid.shape
-
-        rpix = np.sqrt(rowgrid**2 + colgrid**2)
-
-        phipix = np.arctan2(colgrid, rowgrid)
-
+        self.expected_shape = rpix.shape
         if mask is not None:
             if mask.shape != self.expected_shape:
                 raise ValueError('"mask" has incorrect shape. '
                                  ' Expected: ' + str(self.expected_shape) +
                                  ' Received: ' + str(mask.shape))
-            # a somewhat ugly way to mask pixels
-            # need to bury this in lower level class
-            # rpix[mask == 0] = rrange[0]-1
 
         super(RPhiBinnedStatistic, self).__init__(rpix.reshape(-1),
                                                   phipix.reshape(-1),
@@ -473,23 +472,71 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
         return super(RPhiBinnedStatistic, self).__call__(values.reshape(-1))
 
 
-class RadialBinnedStatistic(RPhiBinnedStatistic):
+class RadialBinnedStatistic(BinnedStatistic1D):
     """
-    Create a 1-dimensional histogram by binning a 2-dimensional
-    image in radius.
+    Create a 2-dimensional histogram by binning a 2-dimensional
+    image in both radius and phi.
     """
 
     def __init__(self, rowsize, colsize, bins=10, range=None,
-                 rowc=None, colc=None, mask=None,
-                 statistic='mean'):
+                 rowc=None, colc=None, mask=None, statistic='mean'):
         """
-        See RPhiBinnedStatistic documentation.
-        """
+        Parameters:
+        -----------
+        rowsize,colsize: int
+            shape of image in pixels.
+        bins : int or sequence of scalars, optional
+            If `bins` is an int, it defines the number of equal-width bins in
+            the given range (10 by default).  If `bins` is a sequence, it
+            defines the bin edges, including the rightmost edge, allowing for
+            non-uniform bin widths.  Values in `x` that are smaller than lowest
+            bin edge are assigned to bin number 0, values beyond the highest
+            bin are assigned to ``bins[-1]``.
+        range : (float, float) or [(float, float)], optional
+            The lower and upper range of the bins.  If not provided, range
+            is simply ``(x.min(), x.max())``.  Values outside the range are
+            ignored.
+        rowc,colc: int, optional
+            location (in pixels) of origin (default: image center).
+        mask: 2-dimensional np.ndarray, optional
+            array of zero/non-zero values, same shape as image used
+            in __call__.  zero values will be ignored.
+        statistic : string or callable, optional
+            The statistic to compute (default is 'mean').
+            The following statistics are available:
 
-        super(RadialBinnedStatistic, self).__init__(rowsize, colsize,
-                                                    (bins, 1), range,
-                                                    rowc, colc,
-                                                    mask, statistic)
+              * 'mean' : compute the mean of values for points within each bin.
+                Empty bins will be represented by NaN.
+              * 'median' : compute the median of values for points within each
+                bin. Empty bins will be represented by NaN.
+              * 'count' : compute the count of points within each bin.  This is
+                identical to an unweighted histogram.  `values` array is not
+                referenced.
+              * 'sum' : compute the sum of values for points within each bin.
+                This is identical to a weighted histogram.
+              * function : a user-defined function which takes a 1D array of
+                values, and outputs a single numerical statistic. This function
+                will be called on the values in each bin.  Empty bins will be
+                represented by function([]), or NaN if this returns an error.
+        """
+        rpix, _ = get_r_phi(rowsize, colsize, rowc, colc)
+        self.expected_shape = rpix.shape
+
+        if mask is not None:
+            if mask.shape != self.expected_shape:
+                raise ValueError('"mask" has incorrect shape. '
+                                 ' Expected: ' + str(self.expected_shape) +
+                                 ' Received: ' + str(mask.shape))
+
+        super(RadialBinnedStatistic, self).__init__(rpix.reshape(-1),
+                                                    statistic,
+                                                    bins=bins,
+                                                    range=range)
 
     def __call__(self, values):
-        return np.squeeze(super(RadialBinnedStatistic, self).__call__(values))
+        # check for what I believe could be a common error
+        if values.shape != self.expected_shape:
+            raise ValueError('"values" has incorrect shape.'
+                             ' Expected: ' + str(self.expected_shape) +
+                             ' Received: ' + str(values.shape))
+        return super(RadialBinnedStatistic, self).__call__(values.reshape(-1))
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -426,7 +426,7 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
         """
         Parameters:
         -----------
-        shape: tuple of ints of length 2.
+        shape : tuple of ints of length 2.
             shape of image.
         bins : int or [int, int] or array_like or [array, array], optional
             The bin specification:
@@ -443,9 +443,9 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
             [[rmin, rmax], [phimin, phimax]]. All values outside of this range
             will be considered outliers and not tallied in the histogram.
             See "bins" parameter for definition of phi.
-        origin: tuple of ints with length 2, optional
+        origin : tuple of ints with length 2, optional
             location (in pixels) of origin (default: image center).
-        mask: 2-dimensional np.ndarray of ints, optional
+        mask : 2-dimensional np.ndarray of ints, optional
             array of zero/non-zero values, same shape as image used
             in __call__.  zero values will be ignored.
         statistic : string or callable, optional
@@ -493,8 +493,8 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
 
 class RadialBinnedStatistic(BinnedStatistic1D):
     """
-    Create a 2-dimensional histogram by binning a 2-dimensional
-    image in both radius and phi.
+    Create a 1-dimensional histogram by binning a 2-dimensional
+    image in radius.
     """
 
     def __init__(self, shape, bins=10, range=None,
@@ -511,10 +511,14 @@ class RadialBinnedStatistic(BinnedStatistic1D):
             non-uniform bin widths.  Values in `x` that are smaller than lowest
             bin edge are assigned to bin number 0, values beyond the highest
             bin are assigned to ``bins[-1]``.
+            Phi has a range of -pi to pi and is defined as arctan(col/row)
+            (i.e. y is column and x is row, or "matrix" format,
+            not "cartesian")
         range : (float, float) or [(float, float)], optional
             The lower and upper range of the bins.  If not provided, range
             is simply ``(x.min(), x.max())``.  Values outside the range are
             ignored.
+            See "bins" parameter for definition of phi.
         origin: tuple of ints with length 2, optional
             location (in pixels) of origin (default: image center).
         mask: 2-dimensional np.ndarray of ints, optional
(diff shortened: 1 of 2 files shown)


ordirules:
I'm just aware of this code and this is definitely useful for us too :-). (I might try to also carefully review and give feedback tonight if I can)

Maybe mention Tom and Dan using '@' so they see it?
@tacaswell , @danielballan

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -72,6 +72,9 @@ def __init__(self, sample, statistic='mean',
             minimum and maximum values along each dimension.
 
         """
+        if mask is None:
+            mask = np.ones_like(sample)
```
tacaswell: We need to document the convention of is the mask '1-for-good-pixels' or '1-for-bad-pixels' someplace and make sure we are consistent about it.

attn @CJ-Wright

CJ-Wright: This is already documented in the threshold mask you wrote, [here](https://github.com/scikit-beam/scikit-beam/blob/master/skbeam/core/mask.py#L83)

chrisvam: I have updated the documentation for mask specifying that 0-values are ignored.  Interesting that pyFAI does it the other way.  To add to what CJ wrote: our convention also allows img*mask to do something natural.  I'd be interested to hear what the pyFAI people have to say about the other convention.  Maybe we can ask them on our next phone call.

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -379,43 +398,64 @@ def __init__(self, x, y, statistic='mean',
...
         super(BinnedStatistic2D, self).__init__([x, y], statistic=statistic,
-                                                bins=bins, range=range)
+                                                bins=bins, range=range,
+                                                mask=mask)
 
     def __call__(self, values):
         return super(BinnedStatistic2D, self).__call__(values)
 
 
+def get_r_phi(shape, origin):
```
ordirules: I would use utils.radial_grid and utils.angle_grid rather than add a new function. Drawback is you need two function calls. 
Another option is integrate this into utils somehow. Perhaps create a angle_radial_grid? As a placeholder, it could just call the two functions for now and eventually be changed into something more optimized.

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -434,36 +474,21 @@ def __init__(self, rbins, phibins, rowsize, colsize,
                 will be called on the values in each bin.  Empty bins will be
                 represented by function([]), or NaN if this returns an error.
         """
+        rpix, phipix = get_r_phi(shape, origin)
```
ordirules: here's the suggestion
```python 
from utils import radial_grid, angle_grid
rpix = radial_grid(origin, shape)
phipix = angle_grid(origin, shape)
```

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -130,7 +136,13 @@ def __init__(self, sample, statistic='mean',
...
         Ncount = {}
         for i in np.arange(self.D):
-            Ncount[i] = np.digitize(sample[:, i], self.edges[i])
+            # Apply mask in a non-ideal way by setting value outside range.
+            # Would be better to do this using bincount "weights", perhaps.
+            thissample = sample[:, i]
+            thismask = mask[:, i]
+            thissample[thismask == 0] = (self.edges[i][0] -
+                                         0.01 * (1+np.fabs(self.edges[i][0])))
+            Ncount[i] = np.digitize(thissample, self.edges[i])
```
ordirules: before version 1.10.0 (just last year), numpy had an excruciatingly slow version of np.digitize
(was fixed here https://github.com/numpy/numpy/pull/5101). I don't think any python user will encounter this again, but could we add some note regarding this somewhere just in case?
See release notes : https://github.com/numpy/numpy/blob/master/doc/release/1.10.0-notes.rst

I suggest in the function doc:
Notes
--------
If using numpy versions < 1.10.0, you may notice significantly slow behavior. This has to do with digitize, which was optimized 1.10.0 onwards.

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -70,8 +70,13 @@ def __init__(self, sample, statistic='mean',
             A sequence of lower and upper bin edges to be used if the
             edges are not given explicitely in `bins`. Defaults to the
             minimum and maximum values along each dimension.
-
+        mask : array_like
```
ordirules: for line 60 and 61 above, shouldn't it be a (D,N) array?

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -130,7 +136,13 @@ def __init__(self, sample, statistic='mean',
         # Compute the bin number each sample falls into.
         Ncount = {}
         for i in np.arange(self.D):
-            Ncount[i] = np.digitize(sample[:, i], self.edges[i])
+            # Apply mask in a non-ideal way by setting value outside range.
+            # Would be better to do this using bincount "weights", perhaps.
+            thissample = sample[:, i]
```
ordirules: why are you transposing and then accessing this way? Is it not better to not transpose and just call 
```thissample = sample[i]```
@tacaswell maybe I'm missing something?

Comment (tacaswell) [inline skbeam/core/accumulators/binned_statistic.py]: This code is adapted from source in scipy, I would trust them.

I think transposing _may_ induce a copy if numpy can not sort out a way to compute the required strides for the transposed view.

File: skbeam/core/accumulators/binned_statistic.py [hunk 7]
```diff
@@ -70,8 +70,13 @@ def __init__(self, sample, statistic='mean',
             A sequence of lower and upper bin edges to be used if the
             edges are not given explicitely in `bins`. Defaults to the
             minimum and maximum values along each dimension.
-
+        mask : array_like
+            array of ones and zeros with the same shape as `sample`.
+            Values with mask==0 will be ignored.
         """
+        if mask is None:
+            mask = np.ones_like(sample)
```
ordirules: I don't understand (but I may be wrong! :-) ). `sample` are the bins. How can `mask` be the same shape as `sample` (i.e. for 2D bins, then the shape of `sample` will be (2, dimx, dimy) where dimx, dimy are dimensions of sample)?
How are you testing the code? Are you using real data? I think for something like this, we should have a working example visually tested and posted before accepting this, regardless of the tests.

I can also write a notebook later using your code, see if it works, if it helps :-).

ordirules: I also think that mask is data dependent. I would vote for keeping the mask at a higher level (i.e. binnedstatistic2D)

tacaswell:
> there are about 10 lines of code that are duplicated in both RPhiBinnedStatistic and RadialBinnedStatistic. I might be able to fix this with some inheritance or fancy python tricks

I would take code repetition over getting too fancy.

skbeam/core/accumulators/binned_statistic.py
chrisvam: Thanks very much for the great suggestion, Julien, and my apologies for taking some days to reply (multitasking, as usual).  I tried the above, and encountered two things that I needed to adapt:  I had to add an "indexing" option to radial_grid because I need to use 'ij' indexing of meshgrid, and then I had to swap the rows/columns, so my code ended up looking like this:

        rpix = radial_grid((colc,rowc), (shape[1],shape[0]), indexing='ij')

I'm happy to do it this way, although I would have to add "indexing='xy'" option as the default to utils.py.  Or I could do it the way I currently am.  Anyone have any preferences?

skbeam/core/accumulators/binned_statistic.py
chrisvam: Hi Julien,

Sure, I have added this to the constructor documentation.  One thing with this code is that (at least for LCLS usage patterns) the np.digitize only happens once when the RadialBinnedStatistic is setup.  Then the many images are projected (at 120Hz) without ever calling np.digitize again.  So for LCLS usage patterns the CPU time of the setup in the constructor is negligible compared to the 120Hz analysis of images.

skbeam/core/accumulators/binned_statistic.py
chrisvam: Hi Julien,

As Tom also mentioned, I inherited this code/documentation from scipy's binned_statistic.  To me the order appears correct since in the code it says:

        # Sample is an ND-array.                                                                                              
        N, D = sample.shape

skbeam/core/accumulators/binned_statistic.py
chrisvam: In the time I've looked at the binned_statistic code I inherited from scipy I haven't understood why they transpose the data.  As Tom said, I'm trusting them :-).

skbeam/core/accumulators/binned_statistic.py
chrisvam: Hi Julien,

I think you're correct that my mask shape was wrong.  If sample has shape (N,D) then I think mask should have shape (N).  e.g. if one is producing a projection onto 2D r-phi, the "bad pixel" mask should be the same for both r and phi.  I was testing the 1D mask (with help from Tom) by comparing with the output with scipy's binned_statistic, but I wasn't testing the 2D case.  I will try to add that, comparing with binned_statistic_2d.

Thanks for all the ideas.  Appreciated...

chrisvam: I also meant to add:  I started out trying to keep the mask at a higher level, but when I generalized the binning to behave more like the underlying scipy code, it became more difficult.  Even now the way I'm doing the mask is not ideal, but at least the interface is better.

chrisvam: I think the latest commit tests the 2D RPhiBinnedStatistic case.

skbeam/core/accumulators/binned_statistic.py
ordirules: maybe I misunderstand, why can't you just flip to (colc, rowc), (shape[0], shape[1]), doesn't that achieve the same thing? 
I might be missing something. The only case I'd see it mattering is when the fastest varying index by convention is not the rightermost index (assuming contiguous 2D array in memory)? I could see that not always being true in numpy given you can access the same memory using either `a` or `a.T`. I'm interested to learn why this matters here.
In either case, if the striding/fastest index matters, could you also do this maybe?:
```
rpix = radial_grid((rowc,colc), (shape[0],shape[1])).T
```

skbeam/core/accumulators/binned_statistic.py
ordirules: yeah I agree with you and thought about it later. the masking probably makes most sense at the lower level, since it could be generalized to anything (and binnedstatistic expects the incoming data to be of matching shape as bins, so masking then follows naturally there).

I want to also test your mask with my code later (it might help me better understand the code), when I have access to my machine, maybe I might have some advice to add (but can't now). Yeah, we're all busy multitasking :-).

skbeam/core/accumulators/binned_statistic.py
chrisvam: I tried your two suggestions by adding these two lines at the bottom of get_r_phi:

    rpix = radial_grid((rowc,colc), (shape[0],shape[1])).T
    phipix = angle_grid((rowc,colc), (shape[0],shape[1])).T

The self-tests then failed, I think because the shapes are flipped (vs. the original) in the above call.

I then tried your other suggestion by adding two lines instead:

    rpix = radial_grid((colc, rowc), (shape[0], shape[1]))
    phipix = angle_grid((colc, rowc), (shape[0], shape[1]))

But the self-tests also failed, although it's not clear to me why.  If I left out the second "phipix" line, the self-tests would pass.

When I first implemented this I remember playing around with mesh grid for a couple of hours to see if I could avoid the indexing='ij' option, but failed.  But perhaps someone smarter than me knows of a way to do it.

skbeam/core/accumulators/tests/test_binned_statistic.py
```diff
@@ -23,51 +24,112 @@ def setup(self):
         colarr = np.arange(colsize)
         rowgrid, colgrid = np.meshgrid(rowarr, colarr, indexing='ij')
         self.rgrid = np.sqrt(rowgrid**2 + colgrid**2)
+        self.phigrid = np.arctan2(colgrid, rowgrid)
```
ordirules: Actually, I think the test should also use `radial_grid` and `angle_grid`. Change this to:
```python
    from skbeam.core.utils import angle_grid, radial_grid
    self.rgrid = radial_grid((0,0),(rowsize, colsize))
    self.phigrid = angle_grid((0,0),(rowsize, colsize))
```

I think once you do this the the previous suggestion of using `angle_grid` and `radial_grid` should work. I fetched your code and made those two changes and it worked for me :-). I think this will help us keep the internal functions self-consistent. What do you think?

skbeam/core/accumulators/binned_statistic.py
ordirules: hmm, seems like it should have been this for it to work:
```python
        rpix = radial_grid((origin[1], origin[0]), (shape[1], shape[0]))
        phipix = angle_grid((origin[1], origin[0]), (shape[1], shape[0]))
        phipix = phipix.T
        rpix = rpix.T
```
But you may ignore this and please read the suggestion for using `angle_grid` and `radial_grid` in the test.
I think the trick is that this equivalence is true:
```python
y = np.arange(10)
x = np.arange(5)

Y_xy, X_xy = np.meshgrid(y,x,indexing='xy')
X_ij, Y_ij = np.meshgrid(x,y,indexing='ij')
```
output of `X_ij`, `X_xy`:
```python
In [22]: X_ij
Out[22]: 
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
       [2, 2, 2, 2, 2, 2, 2, 2, 2, 2],
       [3, 3, 3, 3, 3, 3, 3, 3, 3, 3],
       [4, 4, 4, 4, 4, 4, 4, 4, 4, 4]])

In [23]: X_xy
Out[23]: 
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
       [2, 2, 2, 2, 2, 2, 2, 2, 2, 2],
       [3, 3, 3, 3, 3, 3, 3, 3, 3, 3],
       [4, 4, 4, 4, 4, 4, 4, 4, 4, 4]])
```

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -434,36 +477,21 @@ def __init__(self, rbins, phibins, rowsize, colsize,
...
-        if rrange is None:
-            rrange = (rpix.min(), rpix.max())
-        if phirange is None:
-            phirange = (-np.pi, np.pi)
+        self.expected_shape = shape
         if mask is not None:
             if mask.shape != self.expected_shape:
                 raise ValueError('"mask" has incorrect shape. '
                                  ' Expected: ' + str(self.expected_shape) +
                                  ' Received: ' + str(mask.shape))
```
ordirules: in the if statement put:
```
mask = mask.reshape(-1)
```

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -434,36 +477,21 @@ def __init__(self, rbins, phibins, rowsize, colsize,
...
-            # a somewhat ugly way to mask pixels
-            rpix[mask == 0] = rrange[0]-1
 
         super(RPhiBinnedStatistic, self).__init__(rpix.reshape(-1),
                                                   phipix.reshape(-1),
                                                   statistic,
-                                                  bins=(rbins, phibins),
-                                                  range=(rrange, phirange))
+                                                  bins=bins,
+                                                  mask=mask.reshape(-1),
```
ordirules: change this to
`mask=mask`

ordirules:
Ok, I ran a quick test on the 2D binning using scipy and the new routine here. I fetched @chrisvam 's code first. I had to make a slight modification on a line mentioned for it to work with when `mask= None`. (There may be better ways to do this)

The results with scipy agree without mask but not with mask. See two notebooks I tried (we will be using this code so we're testing already :-) ):
no mask:
https://github.com/ordirules/notebooks/blob/master/test_BinnedStatistic2D_nomask.ipynb

with mask:
https://github.com/ordirules/notebooks/blob/master/test_BinnedStatistic2D_withmask.ipynb

You won't be able to run the notebook but can you see the results?
I'm not sure what the best way is of testing/presenting results but uploading a notebook to my own git repo then sharing the link seems to work.

Going through now, and might have some other suggestions for masking @chrisvam , I could possibly send you a pull request with suggested changes and you can review them and decide to merge, would that be okay? This is great code and a very useful and needed feature for scikit-beam, thanks for putting this in!

chrisvam:
Hi Julien,

Yeah, that would be great if you had a PR that I could incorporate.

I think it makes sense that 'mean' doesn't agree with scipy, because the way I'm implementing the mask currently is to "move" the masked pixel into an overflow bin (this is not ideal, using the bincount weights argument would probably be better).  But no matter how the mask is implemented, it will change the result of 'mean' vs. scipy, because some pixels have been moved: we are doing something fundamentally different.  'sum' however should work, since for that operation setting a pixel to zero is the same as removing it entirely.  That's why in my test code I only compare 'sum' to scipy when using the mask.  Would that explain the difference you see?

skbeam/core/accumulators/tests/test_binned_statistic.py
chrisvam: Hmm.  I'll have to think about this.  That suggests that using the radial_grid/phi_grid is fundamentally changing the grid compared to what I had previously.  Ideally I would like to use your calls, but keep the grid identical to what it was previously so that the definition of phi/r don't change, for example.  But perhaps I'm missing something, so feel free to disagree...

ordirules:
@chrisvam nevermind I got them to agree (the way I did it, `mean` should have worked). The issue had to do with something silly. I no longer specified all the r,phi coordinates but the parts that were not masked, which effectively changed the bin_edges that scipy binned_stats computed. To fix that, rather than give bin numbers, I supplied the same bin edges calculated by BinnedStatistic to scipy `binned_stats`. See here:
https://github.com/ordirules/notebooks/blob/master/test_BinnedStatistic2D_withmask.ipynb

Anyway, as far as I can tell, they agree exactly so your trick works just fine, and makes the code look intuitive. Personally, I feel that it might be better to save the mask indexes (just to reduce computation time) and only compute binned_statistic on these selected values. But given the complexity of the code perhaps we could take it one step at a time and first work on merging this and running timing/performance tests on different versions of mask indexing later?

I will send a suggested PR a little later and feel free to accept/reject the changes (it may be better to incorporate them manually one by one than just clicking merge). I look forward to using this code, it's much needed in the skbeam community!

skbeam/core/accumulators/tests/test_binned_statistic.py
ordirules: as far as I can tell, it seems to be just a numerical precision error (I checked the arrays that resulted during the test), and it seems to be a difference in the 16th decimal digit or so. For example, replacing the test with:
```assert_array_almost_equal(ref, binned,decimal=10)```
will pass.

So under the hood, I would be tempted to say that the calculation is not exactly the same in terms of numerical operations, but the general API is not broken and what we expect it to do has not changed.
So I think one option could be to  leave the code as is and use `assert_array_almost_equal` in the test. What do you think?

**Comment by chrisvam** [comment]
That all sounds good, Julien.  If you send me a PR with your proposed changes, that would be awesome.  Thank you very much for looking at this ... the code is a little bit fancy, so great to have a second person looking at it.

**Comment by ordirules** [comment]
sounds good started looking at it again now, just one question before I submit a PR to you. I found the problem why `get_r_phi` didn't agree with `radial_grid` and `angle_grid`. The problem lies in `angle_grid` (and not `radial_grid`). You can test by using `radial_grid` but not `angle_grid` (and use `get_r_phi` for that).

We make two different fundamental assumptions here:
```     self.phigrid = np.arctan2(colgrid, rowgrid)
```
I would argue that this should be the other way around, the row grid (going up and down in the array) should be the 'y' and column grid should be 'x'. Would this be a suitable change of assumptions? I think accepting the change to:
```     self.phigrid = np.arctan2(rowgrid, colgrid)
```
would make the code compatible with `angle_grid` in a nice way. What do you think?
(agreed, this code is involved, nice to have multiple people stabbing at it!)

chrisvam:
I would have one request, Julien, if it's not too much trouble:  I'm trying to get this PR in so that I can start to try to really "sell" skbeam to LCLS users, so they start getting used to the idea and hopefully excited about it (along with the histogramming code we worked on).  So if you get me your PR soon-ish (some number of days?) I would be grateful.  Thanks!

ordirules:
Done. Take a look and let us know when done and what you agree or disagree with. I've tested what I sent you, results here:
https://github.com/ordirules/notebooks/blob/master/test_BinnedStatistic2D_withmask.ipynb

then we can summarize for everyone else. Anyway, it works perfect and the API is great, so making the code finer/faster (if at all possible) could be left to another PR.

thanks for submitting this, and I hope it sells well!

chrisvam:
Hi Julien,

Tom/Dan/Eric and I discussed the x/y row/col convention previously.  As I understand it, the advantage of having x/y=col/row is that x/y appear natural in a typical imshow plot.  The advantage of having x/y=row/col is that coding is more natural with arr[x][y] (in my original PR I gave the user the ability to choose between the two, but Tom argued that was too complex, and I agree).

I think the argument at the time was that imshow behavior could be changed with the "origin" keyword, so at the time we opted for the second convention.  I have some preference for that second convention since arr[y][x] feels unintuitive, but don't feel super strongly about it.  Whatever we choose, however, we should be consistent across all skbeam software.  Perhaps @danielballan or @tacaswell wants to chime in with a convention decision?

ordirules:
I apologize for running in circles on this topic and thanks for bringing it to focus. I also came from a [x,y] convention in the past (yorick) [http://yorick.sourceforge.net/]. However, in my case, the fastest varying index in the array was still x (i.e. a[x,y] in yorick was equivalent to a[y,x] in numpy for the same physical structure in memory. So both are still accessed as [x + width*y] when the array is collapsed to 1D).
So I would vote for having x being the fastest varying index, hence use the notation [y,x] here. The notation changes when moving to numpy but in my opinion I feel like it makes more intuitive sense). I think thinking it in terms of fastest varying index (and maybe documenting so) could help with sharing the data across platforms (C).

Any other thoughts? I could be wrong about this and I also would like some input as to what is a good way to think about this. Thanks!

ordirules:
I apologize for running in circles on this topic and thanks for bringing it to focus. I also came from a [x,y] convention in the past (yorick) [http://yorick.sourceforge.net/]. However, in my case, the fastest varying index in the array was still x (i.e. a[x,y] in yorick was equivalent to a[y,x] in numpy for the same physical structure in memory. So both are still accessed as [x + width*y] when the array is collapsed to 1D).
So I would vote for having x being the fastest varying index, hence use the notation [y,x] here. The notation changes when moving to numpy but in my opinion I feel like it makes more intuitive sense). I think thinking it in terms of fastest varying index (and maybe documenting so) could help with sharing the data across platforms (C).

Any other thoughts? I could be wrong about this and I also would like some input as to what is a good way to think about this. Thanks!

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -130,7 +140,12 @@ def __init__(self, sample, statistic='mean',
         # Compute the bin number each sample falls into.
         Ncount = {}
         for i in np.arange(self.D):
-            Ncount[i] = np.digitize(sample[:, i], self.edges[i])
+            # Apply mask in a non-ideal way by setting value outside range.
+            # Would be better to do this using bincount "weights", perhaps.
+            thissample = sample[:, i]
+            thissample[mask == 0] = (self.edges[i][0] -
```
tacaswell: suggest putting the `mask` test here.  If there is no mask you will do a whole bunch of work to not do any work.  `if mask is not None` is cheaper

tacaswell:
@ordirules I do not follow you about C vs numpy.  Both are row-major and have the fastest axes last in the `[]` notation. 

These sorts of x/y vs i/j vs row/col issues have been a constant headache for me :disappointed: 

It is entirely possible that angle_grid is wrong, I will need to look at plots of them to check.

skbeam/core/accumulators/binned_statistic.py
chrisvam: Agreed.  I have added this in the most recent commit.  Also included @ordirules  suggestion of using radial_grid/angle_grid and updated phi-definition documentation accordingly.

ordirules:
@tacaswell I'm sorry my mistake, too many languages. Indeed C and numpy share the same convention (and matlab as far as I can see online). However, yorick has yet another convention. Its fastest varying index is actually the leftermost index! Yet more evidence to support the headache :-( (see (here)[http://yorick.sourceforge.net/refcard/qrlang09.php] , second line)

Sounds good, `angle_grid` looks fine to me, but def worth scrutiny. In the meantime, I think I know what @danielballan 's vote might be, `git blame` says he's the author of this ;-) (in `angle_grid` definition)
```python
    # row is y, column is x. "so say we all. amen."
    x, y = np.meshgrid(pixel_size[1] * (np.arange(shape[1]) -
                                        center[1]),
                       pixel_size[0] * (np.arange(shape[0]) -
                                        center[0]))
    return np.arctan2(y, x)
```

danielballan:
Thanks, @chrisvam. @tacaswell and @ordirules and I are meeting in person and are happy with this. Thanks for going through the revisions.

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -434,36 +458,26 @@ def __init__(self, rbins, phibins, rowsize, colsize,
                 will be called on the values in each bin.  Empty bins will be
                 represented by function([]), or NaN if this returns an error.
         """
+        if origin is None:
+            origin = shape[0]//2, shape[1]//2
```
ordirules: @chrisvam @danielballan @tacaswell 

sorry to reference a merged PR, but I forgot to mention this. I think this should rather be:
```python
origin = shape[0]/2., shape[1]/2.
```
the reason being that origin may be a `float`.

Would there be objections to changing this default?

ericdill: Do we have detectors with odd numbers of rows or columns?

ordirules: I was thinking more for simulations. I generate lots of simulations to predict/confirm certain results and I do play with the parity of pixel number just to test for bugs.

In either case, I can just manually supply the center. I figured I'd ask in case others would agree with this convention.

ericdill: I mean in general I don't have any issue with it.  I was mostly curious as to how frequently that issue would be hit.  

nit: you don't need the `.` after the `2` since we are importing `division` from `__future__` which makes division with `/` always do float division

chrisvam: I think it can't hurt, so it would be fine with me.  People in principle could run this algorithm on a region-of-interest which had an odd number of rows/cols, for example.  I'm happy to do a PR or you can.  Totally up to you...

ordirules: sure then, let's do it before it becomes used routinely. If you could submit that would be great!
correction, I also meant (if index refers to pixel center for 2 based indexing, then true center should be this):
```python
origin = (shape[0]-1)/2, (shape[1]-1)/2
```

PR #469, opened 2016-09-22
don't use scipy._lib.six callable

chrisvam:
As discussed with Tom, this one line was brought in from scipy binned_statistic to support the fact that "callable" is not available in python 3.1.  That's an old version which we don't need to support.  Removing this line enables more people to use skbeam's binned_statistic because some versions of scipy don't have scipy._lib.six.

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -39,7 +39,6 @@ from __future__ import division, print_function, absolute_import
 import warnings
 
 import numpy as np
-from scipy._lib.six import callable
 from ..utils import radial_grid, angle_grid
 
 


ericdill:
+1 from me on not supporting py3.1.  do we explicitly enumerate the versions if Python that we do support anywhere?

ericdill:
Why were we importing from the six that gets vendored by scipy instead of just importing from six that is a dep of this library?

tacaswell:
The history of this code is that it is a copy-paste-modify job to make a callable class that caches the setup for `binned_statistics`.  That it was using their vendored six slipped by in review as (at least I) was trying to not review the scipy code guts, just the the splitting it up logic.

ericdill:
Ah I see thanks. That makes sense

PR #470, opened 2016-09-22
added bin_centers property

ordirules:
I could use this for binnedstat `bin_centers` property to `BinnedStatistic`.
@chrisvam @tacaswell @danielballan 

Saves typing time.
Would this be okay? Overhead should be slight since the calculation is solely in the initial declaration.

thanks!


chrisvam:
For what it's worth:  it looks useful to me...

ericdill:
LGTM

danielballan:
👍

ordirules:
:ok_hand:

PR #473, opened 2016-09-23
Fix 1d

chrisvam:
Hey everyone,

I believe I made an error in the BinnedStatistic1D interface for edges/centers:  I believe for the one-dimensional case it is more intuitive for users if they do not use bin_edges[0] but instead use the simpler bin_edges.  You can see it makes the testing code cleaner, for example.

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -567,6 +567,22 @@ class RadialBinnedStatistic(BinnedStatistic1D):
                                                     mask=mask,
                                                     range=range)
 
+    @property
+    def bin_edges(self):
+        """
+        bin_edges : 1D array of dtype float
+        Return the bin edges.
+        """
+        return super(RadialBinnedStatistic, self).bin_edges[0]
+
+    @property
+    def bin_centers(self):
+        """
+        bin_centers : 1D array of dtype float
+        Return the bin centers.
+        """
+        return super(RadialBinnedStatistic, self).bin_centers[0]
+
     def __call__(self, values):
         # check for what I believe could be a common error
         if values.shape != self.expected_shape:
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -336,6 +336,22 @@ class BinnedStatistic1D(BinnedStatisticDD):
                                                 bins=bins, range=range,
                                                 mask=mask)
 
+    @property
+    def bin_edges(self):
+        """
+        bin_edges : 1D array of dtype float
+        Return the bin edges.
+        """
+        return super(BinnedStatistic1D, self).bin_edges[0]
+
+    @property
+    def bin_centers(self):
+        """
+        bin_centers : 1D array of dtype float
+        Return the bin centers.
+        """
+        return super(BinnedStatistic1D, self).bin_centers[0]
+
 
 class BinnedStatistic2D(BinnedStatisticDD):
     """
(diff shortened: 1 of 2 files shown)


ordirules:
@chrisvam I agree and like this. I vote +1 :-)
(I'm just a contributor like you, not a maintainer, so we need to wait for other votes)

@tacaswell seems to be build error except py3.5:
```python
ERROR: Failure: RuntimeError (the sip module implements API v11.0 to v11.2 but the PyQt4.QtCore module requires API v11.3)
```

chrisvam:
Hmmm.  My quick impression is that the build error is not because of the lines I changed, but I could be wrong.

ordirules:
I agree, I think it has something to do with a dependency. Also @danielballan will know about this, maybe @ericdill

CJ-Wright:
It is in an error in conda https://github.com/ContinuumIO/anaconda-issues/issues/1068

ericdill:
@chrisvam says
> Hmmm. My quick impression is that the build error is not because of the lines I changed, but I could be wrong.

I agree this is not your fault. Something changed underneath us

@CJ-Wright I'm not convinced that's the right issue. IIUC that is a pyqt5 vs pyqt4 issue.  Travis appears to be installing pyqt4 [here] (https://travis-ci.org/scikit-beam/scikit-beam/jobs/162109015#L341).  I am rather lost as to what this issue could be.  I'm working on updating the travis build so that it pulls from conda-forge and shows the channel urls where everything comes from. I'm hoping that pulling most packages from conda-forge will fix this issue.

CJ-Wright:
@ericdill If you could fix this that would be awesome, since I am having the same problem.
The matplotlib issue is here https://github.com/matplotlib/matplotlib/issues/7165 so idk, I have tried pinning things to pyqt4 as well with no help.

ordirules:
travis worked on my PR so I think the issue is fixed (I clicked "restart build" in Travis for this one here, you need to login with github first to see this option).
thanks @ericdill

ordirules:
quick small after thought:
Since we're returning bin_edges as a list, to be consistent with the API, should we also perhaps change the API to work from:
```python
rbinstat = RadialBinnedStatistic(img.shape, mask=mask,origin=origin,bins=[bins])
```
to

```python
rbinstat = RadialBinnedStatistic(img.shape, mask=mask,origin=origin,bins=bins)
```

might not be too late to change. @chrisvam 
(and anyone else who may have started using/testing it @CJ-Wright @tacaswell @danielballan )

chrisvam:
Hi Julien,

I'm probably being slow here, but I'm afraid I don't understand yet.  We picked up scipy's binning interface, which feels quite powerful to me.  Maybe you can provide a concrete example?

One heads-up:  I need to focus on other LCLS projects for a while, so won't have much time to work on this, unfortunately.

ordirules:
no problem. It's  a minor detail. The difference is just `[bins]` versus `bins`.
Let's say you want to use bins used from one `rbinstat` to another `rbinstat`:
```python
rbinstat = RadialBinnedStatistic(img.shape, mask=mask,origin=origin,bins=bins)
# after some code later where you may have changed the mask, or maybe have a different image shape
rbinstat2 = RadialBinnedStatistic(img2.shape, mask=mask2,origin=origin,bins=[rbinstat1.bin_edges])
```
It would be nicer not to have to use those square brackets and rather do:
```python 
rbinstat2 = RadialBinnedStatistic(img2.shape, mask=mask2,origin=origin,bins=rbinstat1.bin_edges)
```
note that in the 2D case, the call would be without brackets since the returned `bin_edges

PR #474, opened 2016-09-23
WIP: Use explicit condarc for travis

ericdill:
Looking at the logs on travis it is not necessarily clear where packages are
coming from. Additionally, we should be using conda-forge to pull packages.

Let's see how this goes...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_install:
   - chmod +x miniconda.sh
   - ./miniconda.sh -b -p ~/mc
   - export PATH=~/mc/bin:$PATH
-  - conda update conda
+  - conda update conda --yes
   - export CONDARC=ci/condarc
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,15 +20,14 @@ python:
   - 3.5
 
 before_install:
-  - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then wget http://repo.continuum.io/miniconda/Miniconda-3.5.5-Linux-x86_64.sh -O miniconda.sh; else wget http://repo.continuum.io/miniconda/Miniconda3-3.5.5-Linux-x86_64.sh -O miniconda.sh; fi
-  - chmod +x miniconda.sh
-  - ./miniconda.sh -b -p /home/travis/mc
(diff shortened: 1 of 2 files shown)


ericdill:
There are two big changes and one small change.

### Small change

Just install Miniconda3-latest (python 3).  This costs us a few extra seconds on the python 2 builds when it has to download python 2.7 from conda-forge, but saves us some complex one-liner if/then bash syntax in the travis.yaml

### Two big changes:

1. Set the channel order to be conda forge and then defaults, instead of lightsource2.  Since we are releasing conda-packages on conda-forge for scikit-beam, it makes sense to run CI on the conda packages from conda-forge.

2. Inject numpy 1.10 and 1.11 into the build matrix so that we ensure that scikit-beam works with both versions.  Again this is due to the release mechanism on conda-forge, where we are producing numpy 1.10 and 1.11 packages for each of the python versions (2.7, 3.4, and 3.5).  Testing against both numpy versions makes more sense.

.travis.yml
```diff
@@ -13,24 +13,28 @@ env:
...
   - 3.5
 
+env:
+  - NUMPY=1.10
+  - NUMPY=1.11
+
+
 before_install:
-  - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then wget http://repo.continuum.io/miniconda/Miniconda-3.5.5-Linux-x86_64.sh -O miniconda.sh; else wget http://repo.continuum.io/miniconda/Miniconda3-3.5.5-Linux-x86_64.sh -O miniconda.sh; fi
+  - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
```
licode: 👍 agree on using python 3.

licode:
Thanks on the update. 
I am fine with the change. 
Any more suggestion on defaults and conda-forge channels?

ericdill:
@licode I'm not sure how to respond to your question. Could you elaborate?

licode:
Never mind. I just want to check if others still have comments.

ericdill:
Ah I see. Sorry that I did not understand!

ericdill:
Thanks @licode

licode:
good work!

PR #475, opened 2016-09-24
modified origin default

ordirules:
modified the origin of detector, as mentioned in #466 (most recent comment), discussion with @chrisvam, @ericdill 
thanks!

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -444,7 +444,7 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
             [[rmin, rmax], [phimin, phimax]]. All values outside of this range
             will be considered outliers and not tallied in the histogram.
             See "bins" parameter for definition of phi.
-        origin : tuple of ints with length 2, optional
+        origin : tuple of floats with length 2, optional
             location (in pixels) of origin (default: image center).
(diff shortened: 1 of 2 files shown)


chrisvam:
Looks good to me.  Thanks, Julien...

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -468,7 +468,7 @@ def __init__(self, shape, bins=10, range=None,
                 represented by function([]), or NaN if this returns an error.
         """
         if origin is None:
-            origin = shape[0]//2, shape[1]//2
+            origin = shape[0]/2., shape[1]/2.
```
tacaswell: The doc string says that these should be ints.  Does anything this gets passed into require integers?

chrisvam: I believe it's only used to setup the grid, which I believe doesn't require integers.  But you're probably right:  the doc string should be updated.

ordirules: updated, @tacaswell thanks for catching that, @chrisvam thanks for input

ordirules:
thanks. test has been updated to test for `origin=None` option as well. Also had to make one of the dimensions odd to properly test, and recompute the `rgrid` and `phigrid` in the loop since origin is modified in the kwargs.

PR #477, opened 2016-10-21
ENH: use fabio to perform fit2d mask saving and loading, with boolian…

CJ-Wright:
… switching to make certain masks end up in skbeam convention


CJ-Wright:
@tacaswell @danielballan ready for review

tacaswell:
Please put shims to not break API.

tacaswell:
Is the existing writer broken?

CJ-Wright:
The existing writer will only work with 2048x2048 images. What do you mean shims?

skbeam/io/fit2d_save.py
```diff
@@ -1,120 +1 @@
-from __future__ import print_function, division
```
tacaswell: can you add a deprecation warning here directing people to the new location?

CJ-Wright: What is the syntax for that?

ericdill:
Consider renaming `skbeam/io/fit2d_io.py` to `skbeam/io/fit2d.py`.  We don't need `io` in the import twice

ericdill:
Otherwise :+1: for simplified code. Any idea how good the fabio conda package is working?

skbeam/io/fit2d_save.py
CJ-Wright: I think that I have added the warning correctly.

CJ-Wright:
Thanks @tacaswell !

PR #482, opened 2017-01-09
ENH: allow for changing statistic via call

CJ-Wright:
Closes #481

PR #484, opened 2017-01-11
Critical bug in accumulators

danielballan:
Unfortunately, I think the accumulators are broken in the recently-tagged version of scikit-beam.

```python
binner = BinnedStatisticDD(np.ones((5, 5)))
```

hits a circular reference introduced in #482 and leads to a maximum recursion depth error. This clearly indicates a need for more test coverage, which I will add to this PR before it is merged.

This PR include a fix for that bug and a more minor related bug. It also includes an opinionated API change that should be discussed. See commit messages.

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -81,8 +81,6 @@ class BinnedStatisticDD(object):
         from 1.10.0 onwards.
         """
 
-        self.statistic = statistic
-
         # This code is based on np.histogramdd
         try:
             # Sample is an ND-array.
@@ -167,8 +165,15 @@ class BinnedStatisticDD(object):
         for i in np.arange(0, self.D - 1):
             self.xy += Ncount[self.ni[i]] * self.nbin[self.ni[i + 1:]].prod()
         self.xy += Ncount[self.ni[-1]]
-        if self.statistic in ['mean', 'std', 'count']:
-            self.flatcount = np.bincount(self.xy, None)
+        self._flatcount = None  # will be computed if needed
+
+    @property
+    def flatcount(self):
+        # Compute flatcount the first time it is accessed. Some statistics
+        # never access it.
+        if self._flatcount is None:
+            self._flatcount = np.bincount(self.xy, None)
+        return self._flatcount
 
     @property
     def bin_edges(self):
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -188,7 +188,7 @@ class BinnedStatisticDD(object):
 
     @property
     def statistic(self):
-        return self.statistic
+        return self._statistic
 
     @statistic.setter
     def statistic(self, new_statistic):
@@ -196,7 +196,7 @@ class BinnedStatisticDD(object):
                 self.statistic) and self.statistic not in std_:
             raise ValueError('invalid statistic %r' % (self.statistic,))
         else:
-            self.statistic = new_statistic
+            self._statistic = new_statistic
 
     def __call__(self, values, statistic=None):
         """
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -61,6 +61,24 @@ class BinnedStatisticDD(object):
         sample : array_like
             Data to histogram passed as a sequence of D arrays of length N, or
             as an (N,D) array.
+        statistic : string or callable, optional
+            The statistic to compute (default is 'mean'). To compute multiple
+            statistics efficiently, override this at __call__ time.
+            The following statistics are available:
+
+              * 'mean' : compute the mean of values for points within each bin.
+                Empty bins will be represented by NaN.
+              * 'median' : compute the median of values for points within each
+                bin. Empty bins will be represented by NaN.
+              * 'count' : compute the count of points within each bin.  This is
+                identical to an unweighted histogram.  `values` array is not
+                referenced.
+              * 'sum' : compute the sum of values for points within each bin.
+                This is identical to a weighted histogram.
+              * function : a user-defined function which takes a 1D array of
+                values, and outputs a single numerical statistic. This function
+                will be called on the values in each bin.  Empty bins will be
+                represented by function([]), or NaN if this returns an error.
         bins : sequence or int, optional
             The bin specification:
 
@@ -166,6 +184,7 @@ class BinnedStatisticDD(object):
             self.xy += Ncount[self.ni[i]] * self.nbin[self.ni[i + 1:]].prod()
         self.xy += Ncount[self.ni[-1]]
         self._flatcount = None  # will be computed if needed
+        self._statistic = statistic
 
     @property
     def flatcount(self):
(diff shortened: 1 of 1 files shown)


skbeam/core/accumulators/binned_statistic.py
```diff
@@ -188,17 +188,16 @@ def bin_centers(self):
...
-                self.statistic) and self.statistic not in std_:
-            raise ValueError('invalid statistic %r' % (self.statistic,))
+        if not callable(new_statistic) and new_statistic not in std_:
+            raise ValueError('invalid statistic %r' % (new_statistic,))
         else:
-            self.statistic = new_statistic
+            self._statistic = new_statistic
 
-    def __call__(self, values, statistic=None):
+    def __call__(self, values):
```
CJ-Wright: Do you want a local statistic that temporarily overrides the init statistic?

danielballan:
No worries, @CJ-Wright. We've all introduced some bugs from time to time. It just serves to reinforce our healthy paranoia about adding tests next time. :- )

About the API: #484 allowed the user to decide the statistic at `__call__` time, which makes sense. Previously, to compute two different statistics (like mean and std, to pick a common case), one would needlessly have to make two different instances and do the work of `__init__` twice. In fact, as shown in the new commits I added, we don't even need to know that statistic at `__init__` time *at all*. We can move the 'statistic' parameter from `__init__` to `__call__`. As before, it defaults to mean, but is configurable by the user. There is no need to stash state about the statistic. I think this addresses #484's goal in a more complete and clear way.

How does everyone feel about this API? attn @chrisvam @ordirules

tacaswell:
I am :+1: on this API change.  If you want an object bound to not-mean either use partial or sub-class.

ordirules:
I agree, just one small comment. I am wondering if there is something more 'pipeline friendly' we can think of (while still keeping the ability to define the stat on the fly, as you show).

I think moving these parameters to a `set_params` (and `get_params`) function may be cleaner. We could then allow the statistic to either be defined in the constructor or the call. This would amount to setting the default for statistic to be `None` in the call. This would allow the option where, if statistic is not defined in the call (`None`), then check the internal settings.

The reason why I'm thinking this is eventually, I think we may want to go from:
``` python
binner = BinnedStatisticDD(np.ones((5, 5)), **kwargs)
result_img = binner(img) 
```

To something pipelined like: (ex: using the mean statistic then rebin again using the std statistic):
``` python
binner1 = BinnedStatisticDD(np.ones((100, 100)), **kwargs)
binner2 = BinnedStatisticDD(np.ones((5, 5)),statistic='std', **kwargs2)
result_img2 = binner2(binner1(img))
```

We likely could automate this with some wrapper (say, `Pipeline` as in skimage). This would require some form of a list of classes and and parameters to try. The problem with having some parameters in the initializer and some in the call will make handling the two cases a little trickier I think, and something simple like `set_params` would do the trick (and defaulting to checking for internal parameters instead of having a hard coded default).

I might have overthought this, and I defer to your experience. What do you think? If you agree, we can better precise this later. If this is still subject of contention, perhaps the statistic should be moved back to the definition for now to fix the bug and a separate PR made?

(PS : I like the flatcount function trick, nice and clean solution to that problem!)

tacaswell:
I think thinking about pipelining is good, but maybe premature.  In analogy to sklearn, they have (to their great credit) reduced the API to something very simple:  http://scikit-learn.org/stable/developers/contributing.html#apis-of-scikit-learn-objects .  All data is 1D or 2D arrays, rows are an instance of a measurement and columns are the features measured.  My understanding is that a fair amount of a machine learning pipeline is transforming your data so that they are all in the same units / scale.  I do not think we understand all of our data needs yet to be able to do something similar.

I am also concerned about baking the pipe-lining logic in at this layer, given the number of existing projects who's only goal is pipelining we want to be sure that we do not bake design decisions in at the bottom that are going to inter-operating with many down-stream frameworks difficult.

To slightly re-write your example

```python
from functools import partial
binner1 = partial(BinnedStatisticDD(np.ones((100, 100))), **kwargs)
binner2 = partial(BinnedStatisticDD(np.ones((5, 5))) ,statistic='std', **kwargs2)
result_img2 = binner2(binner1(img))
```

One could also image a pipelining framework which adopts something like the `Msg` as the core data structure and does

```python
def run(pipeline, inp):
    ret = inp
    for func, args, kwargs in pipeline:
        ret = func(ret, *args, **kwargs)
    return ret
```

which allows the outer layer to own sorting out the extra args/kwargs.

I would also point out that we did not have to use classes for this functionality.  Things of the form

```python
def _body_of_call(inp, lots, of, args):
    # body of the base __call__ method

def specialized_binner_maker(a, b, c):
    # body of __init__
    computed_kwargs = make_kwargs(a, b, c)
    def specialized_binner(inp, statistic='mean'):
         return _body_of_call(inp, statistic=statistic, **computed_kwargs)

    # specialized_binner.bins = computed_bins
    return specialized_binner
```

which would have almost the same api and be purely functional (the one thing we would loose is the bins accessor, which we could still put back on).

ericdill:
Should we get this in since max recursion bugs are bad and move the pipelineing issue to a separate issue?

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -167,95 +164,101 @@ def __init__(self, sample, statistic='mean',
...
-    @statistic.setter
-    def statistic(self, new_statistic):
-        if not callable(
-                self.statistic) and self.statistic not in std_:
-            raise ValueError('invalid statistic %r' % (self.statistic,))
-        else:
-            self.statistic = new_statistic
-
-    def __call__(self, values, statistic=None):
+    def __call__(self, values, statistic='mean'):
```
ordirules: suggestion : `statistic = 'None' `

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -167,95 +164,101 @@ def __init__(self, sample, statistic='mean',
...
         statistic_values : array
             The values of the selected statistic in each bin.
         """
-
-        if statistic is not None:
-            self.statistic = statistic
-            if self.statistic in ['mean', 'std', 'count']:
-                self.flatcount = np.bincount(self.xy, None)
-
         self.result = np.empty(self.nbin.prod(), float)
```
ordirules: suggestion: 
``` python
if statistic is None:
    statistic = self.statistic
````

File: skbeam/core/accumulators/binned_statistic.py [hunk 4]
```diff
@@ -45,8 +45,7 @@
 class BinnedStatisticDD(object):
     std_ = ('mean', 'median', 'count', 'sum', 'std')
 
-    def __init__(self, sample, statistic='mean',
-                 bins=10, range=None, mask=None):
+    def __init__(self, sample, bins=10, range=None, mask=None):
```
ordirules: suggestion: do not change this line (keep old)

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -81,8 +80,6 @@ def __init__(self, sample, statistic='mean',
         from 1.10.0 onwards.
         """
 
-        self.statistic = statistic
```
ordirules: suggestion: keep this line

ordirules:
I agree. For this PR (irrespective of pipeline), I suggest the changes written as line edits (in a review). @danielballan, please review them and let me know if you agree. If not, this is fine with me as well.
Thank you for submitting the changes and maintaining this useful code, and @CJ-Wright for the enhancement. :-)

ordirules:
@tacaswell thanks for sharing this information and these different alternatives to developing a pipeline.
Sorry to bring up this topic here, and I agree with you (and disagree with my past self) that decisions shouldn't be made here.
I would be interested to follow the discussion if public (I'd appreciate any refs/links).

licode:
I think this is ready to go. More comments?

ordirules:
Agreed. I suggested changes (here) [https://github.com/danielballan/scikit-beam/pull/8/files], although I may still have missed something. (This was a minor change to keep kwargs consistent with previous declarations). @danielballan what do you think?

If there is disagreement on that, then I suggest merge now and we could discuss an API change can be for later?

licode:
fair enough. Self merged when you are happy.

licode:
ready to go? @danielballan

danielballan:
Not yet, but I will definitely get to this today.

danielballan:
Here's where we landed:

* `__init__` accepts a statistic (default 'mean') which serves as the default statistic for the instance.
* `__call__` accepts a statistic (default None) which allows the user to override the default statistic chosen in `__init__`. This is more than a trivial convenience: it is necessary flexibility for efficiently computing multiple statistics on the same bins and data.

No changes to the tests were necessary. Additional tests should be added and I should reread this with clear eyes tomorrow, but I think we are finally close to merging.

skbeam/core/accumulators/binned_statistic.py
danielballan: Just a note: in the end this was kept, just moved.

danielballan:
OK, this still looks right to me. Can someone review/merge?

ordirules:
Looks good to me. I ran your branch on my code with `RadialBinnedStatistic` and `RPhiBinnedStatistic` and works. Good to merge. @licode can you merge? (just to get one more confirmation)

ordirules:
PS: I agree @danielballan , we should write better tests for this in the future.

licode:
You guys can go ahead to do this, as we got two people involved already. But I will do it this time. thanks!

PR #486, opened 2017-01-13
ENH: fix test for fit2d

CJ-Wright:
Fix the test for fit2d.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,6 @@ install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
   - conda create -n testenv pip nose pytest python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY scipy scikit-image six coverage cython xraylib lmfit=0.8.3 netcdf4 flake8 pyfai
   - source activate testenv
-  - if [$TRAVIS_PYTHON_VERION == '2.7']; then pip install backports.tempfile; fi;
   # # need to build_ext -i for the tests so that the .so is local to the source
   # # code.  We could also setup.py develop, but I'm not sure if that is any
   # # better
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,10 @@ env:
   - NUMPY=1.12
   - NUMPY=1.13
 
+exclude:
+  - env: NUMPY=1.10
+    python: 3.6
+
 
 before_install:
   - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
   - conda create -n testenv pip nose pytest python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY scipy scikit-image six coverage cython xraylib lmfit=0.8.3 netcdf4 flake8 pyfai
   - source activate testenv
+  - if [$TRAVIS_PYTHON_VERION == 2.7]; then pip installbackports.tempfile; fi;
   # # need to build_ext -i for the tests so that the .so is local to the source
   # # code.  We could also setup.py develop, but I'm not sure if that is any
   # # better
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,9 +38,8 @@ before_install:
 
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
-  - conda create -n testenv pip nose pytest python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY scipy scikit-image six coverage cython xraylib lmfit=0.8.3 netcdf4 flake8
+  - conda create -n testenv pip nose pytest python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY scipy scikit-image six coverage cython xraylib lmfit=0.8.3 netcdf4 flake8 pyfai
   - source activate testenv
-  - pip install pyFAI
   # # need to build_ext -i for the tests so that the .so is local to the source
   # # code.  We could also setup.py develop, but I'm not sure if that is any
   # # better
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@danielballan ready for review

skbeam/io/tests/test_fit2d_save.py
```diff
@@ -44,16 +44,18 @@
 import numpy as np
 from skbeam.io.fit2d import fit2d_save, read_fit2d_msk
 from numpy.testing import assert_array_equal
+from tempfile import tempdir
 
 
 def test_save_output_fit2d():
+    tdir = tempdir()
```
danielballan: I think you want `TemporaryDirectory`. `tempdir` is the cached result of the most recent call to `tempfile.gettempdir()`, so it is typically a value like `None` or `/tmp`, not a callable. Given that this is somehow passing Travis, I worry that the tests are still not being run. Please test the code locally with print statements to be sure it is actually executed.

skbeam/io/tests/test_fit2d_save.py
```diff
@@ -44,16 +44,18 @@
...
                np.random.random_integers(0, 200))).astype(bool)
 
-    fit2d_save(msk, filename, dir_path=None)
-    msk2 = read_fit2d_msk(filename)
+    fit2d_save(msk, filename, dir_path=tdir)
+    msk2 = read_fit2d_msk(os.path.join(tdir, filename))
     assert_array_equal(msk2, msk)
 
-    os.remove("function_values.msk")
+    os.remove(tdir)
```
danielballan: This does not work on directories, only on files. Once you switch to `TemporaryDirectory` you can use its `cleanup()` method.

licode:
Since this is related to tests, please rebase.

CJ-Wright:
I think the errors may be fixed by #510

skbeam/io/tests/test_fit2d_save.py
CJ-Wright: TemporaryDirectory is not python 2 safe. :(
Do you want backports?

danielballan: That's neat, but I don't think it's worth introducing a new dependency to solve this problem. Python 2 does provide [this tool for creating a temporary directory](https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp); it's just not as nice as the one in Python 3 because you have to do the cleanup manually. (I suggest `shutil.rmtree`.)

skbeam/io/tests/test_fit2d_save.py
```diff
@@ -44,17 +44,24 @@
...
+import sys
+
+if sys.version_info >= (3, 0):
+    from tempfile import TemporaryDirectory
+else:
+    from backports.tempfile import TemporaryDirectory
+
 
-@pytest.mark.xfail(raises=IOError)
 def test_save_output_fit2d():
```
ericdill: There's a built-in fixture in pytest that handles temporary directory creation. pass `tmpdir` as a function argument and pytest will provide that fixture for you. Use it like so:

You can create a file with tmpdir by calling `tmpdir.join('file name')` which will produce an object with some things hanging off of it. The most useful ones in this context are `.dirname` and `.filename`
```python
t = tmpdir.join('file_name')
dirname = t.dirname
filename = t.strpath
```

or in full:

```python
from __future__ import absolute_import, division, print_function
import os
import numpy as np
from skbeam.io.fit2d import fit2d_save, read_fit2d_msk
from numpy.testing import assert_array_equal
import sys


def test_save_output_fit2d(tmpdir):
    t = tmpdir.join('function_values')
    dir_path = t.dirname
    filename = t.strpath
    msk = np.random.random_integers(
        0, 1, (np.random.random_integers(0, 200),
               np.random.random_integers(0, 200))).astype(bool)

    fit2d_save(msk, filename, dir_path=dir_path)
    msk2 = read_fit2d_msk('%s.msk' % filename)
    assert_array_equal(msk2, msk)
```

Then you don't need the backports or py2/py3 compatibility

CJ-Wright: Does tmpdir do cleanup? I was lead astray by [Dan](https://github.com/scikit-beam/scikit-beam/pull/486#discussion_r103450265) :smiling_imp:

ericdill: I doubt that @danielballan led you astray 🤷‍♂️ . He always provides excellent advice AFAICT. 

You can manually clean up if you want by throwing in a `tmpdir.remove(t)` at the end of the function, but I'd **highly** recommend leaving the test data in there. It is generally a very useful feature to be able to inspect whatever is being written by these test functions in case of error. 

Also pytest keeps like 5-7 iterations but automatically deletes stuff older than that. I doubt that this test suite is writing enough data where 5-7 iterations is going to cause problems.  

xref: https://github.com/pytest-dev/pytest/issues/543

ericdill: I was wrong. pytest keeps the last 3 runs. [From the docs](https://docs.pytest.org/en/latest/tmpdir.html):

> Temporary directories are by default created as sub-directories of the system temporary directory. The base name will be pytest-NUM where NUM will be incremented with each test run. Moreover, entries older than 3 temporary directories will be removed.

ericdill:
Since the pyfai numpy 1.10 python 3.6 issue is going to affect all CI builds until it's resolved, I'd like to suggest updating .travis.yml to exclude it from the build matrix. You can do so with the `matrix/exclude` key as follows:

```yaml
matrix:
  include:
    - python: 3.6
      env: BUILD_DOCS=true RUN_TESTS=false NUMPY=1.13

  exclude:
    - env: NUMPY=1.10
      python: 3.6
```

(I'm like 85% confident that this is the right syntax)

.travis.yml
```diff
@@ -27,6 +27,10 @@ env:
   - NUMPY=1.12
   - NUMPY=1.13
 
+exclude:
+  - env: NUMPY=1.10
+    python: 3.6
```
ericdill: this needs to go under the matrix key. so move this to line 18

ericdill: Err these three lines need to go under the matrix key

ericdill:
Travis matrix looks good. I suspect this will be all green shortly

licode:
thanks. good work!

CJ-Wright:
Thank you

PR #487, opened 2017-01-17
Spatial Correlations

ordirules:
Here is an initial pass and suggested API of the mask aware spatial correlations. I tried to remake the API into something more user friendly/scalable. The name is `CrossCorrelator` since it is not necessarily restricted to the spatial domain.

Idea is (`img` and `mask` can be a 1D curve):
```python
    ccorr = CrossCorrelator(mask.shape, mask=mask, method='symavg')
    cimg = ccorr(img)
```
See [pdf](https://github.com/ordirules/notebooks/blob/master/CrossCorrelator.pdf), or alternatively, the [notebook](https://github.com/ordirules/notebooks/blob/master/CrossCorrelator.ipynb) (equations don't show up well for the latter).

Some things to consider:
1. Is `correlation.py` the appropriate place for this code? It's not an accumulator as well, but perhaps should it become one?
2. The routines `_cross_corr`, `_crop_from_mask` and `_expand_image` may likely be replaced by scikit-image routines (or similar). I would like to eventually search but I thought I would start submitting this for open discussion.
3. There will likely be different averaging options. The `method` option accepts either one option, or a list of options. (ex: method =['symavg', 'maxnorm'] etc, where `maxnorm` is currently made up)

@yugangzhang, @danielballan, @tacaswell , @chrisvam, @CJ-Wright please add anyone you think may be able to weigh in on this, we should be extremely critical for the first pass.

Adding in tests is a matter of translating the notebook into a test.

Thanks!

### DIFF
--- a/skbeam/core/correlation.py
+++ b/skbeam/core/correlation.py
@@ -942,6 +942,9 @@ class CrossCorrelator:
         self.tmpimgs = list()
         self.tmpimgs2 = list()
         self.centers = list()
+        self.shapes = list()  # the shapes of each correlation
+        # the positions of each axes of each correlation
+        self.positions = list()
 
         self.ids = np.sort(np.unique(mask))
         # remove the zero since we ignore, but only if it is there (sometimes
@@ -980,7 +983,21 @@ class CrossCorrelator:
             self.maskcorrs.append(maskcorr)
             self.pxlst_maskcorrs.append(maskcorr > 0)
             # centers are shape//2 as performed by fftshift
+            center = np.array(maskcorr.shape)//2
             self.centers.append(np.array(maskcorr.shape)//2)
+            self.shapes.append(np.array(maskcorr.shape))
+            if mask.ndim == 1:
+                self.positions.append(np.arange(maskcorr.shape[0]) - center[0])
+            elif mask.ndim == 2:
+                self.positions.append([np.arange(maskcorr.shape[0]) -
+                                       center[0],
+                                       np.arange(maskcorr.shape[1]) -
+                                       center[1]])
+
+        if len(self.ids) == 1:
+            self.positions = self.positions[0]
+            self.centers = self.centers[0]
+            self.shapes = self.shapes[0]
 
     def __call__(self, img1, img2=None, normalization=None):
         ''' Run the cross correlation on an image/curve or against two
@@ -1056,8 +1073,8 @@ class CrossCorrelator:
                 # only run on overlapping regions for correlation
                 w = self.pxlst_maskcorrs[i]
                 ccorr[w] /= self.maskcorrs[i][w] * \
-                            np.average(self.tmpimgs[i].
-                                       ravel()[self.subpxlsts[i]])**2
+                    np.average(self.tmpimgs[i].
+                               ravel()[self.subpxlsts[i]])**2
 
             ccorrs.append(ccorr)
 
--- a/skbeam/core/correlation.py
+++ b/skbeam/core/correlation.py
@@ -1070,7 +1070,7 @@ def _cross_corr(img1, img2=None):
 
         Parameters
         ----------
-        img1 : 1d or 2d np.ndarray
+        img1 : np.ndarray
             the image or curve to cross correlate
 
         img2 : 1d or 2d np.ndarray, optional
@@ -1090,7 +1090,10 @@ def _cross_corr(img1, img2=None):
             .format(*img1.shape, *img2.shape)
         raise ValueError(errorstr)
 
-    reverse_index = (*((slice(None, None, -1),)*ndim),)
+    # need to reverse indices for second image
+    # fftconvolve(A,B) = FFT^(-1)(FFT(A)*FFT(B))
+    # but need FFT^(-1)(FFT(A(x))*conj(FFT(B(x)))) = FFT^(-1)(A(x)*B(-x))
+    reverse_index = [slice(None, None, -1) for i in range(ndim)]
     imgc = fftconvolve(img1, img2[reverse_index], mode='same')
 
     return imgc
--- a/skbeam/core/correlation.py
+++ b/skbeam/core/correlation.py
@@ -46,7 +46,7 @@ from .utils import multi_tau_lags
 from .roi import extract_label_indices
 from collections import namedtuple
 import numpy as np
-from numpy.fft import fft2, ifft2, fftshift
+from scipy.signal import fftconvolve
 # for a convenient status bar
 try:
     from tqdm import tqdm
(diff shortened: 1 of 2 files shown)


skbeam/core/correlation.py
```diff
@@ -842,3 +850,237 @@ def one_time_from_two_time(two_time_corr):
...
+        '''
+        if axes is None:
+            if len(shape) == 2:
+                axes = (-2, -1)
+            elif len(shape) == 1:
+                axes = (-1,)
+
+        if normalization is None:
+            normalization = ['symavg', 'regular']
+        elif type(normalization) is not list:
```
licode: isinstance is better than type.

skbeam/core/correlation.py
```diff
@@ -842,3 +850,237 @@ def one_time_from_two_time(two_time_corr):
...
+            # quick fix for finite numbers should be integer so
+            # choose some small value to threshold
+            maskcorr *= maskcorr > .5
+            self.maskcorrs.append(maskcorr)
+            self.pxlst_maskcorrs.append(maskcorr > 0)
+
+    def __call__(self, img1, img2=None, normalization=None, PF=False,
+                 return_indices=False):
+        ''' Run the analysis on an image, or two
+
```
licode: what is PF? can you please add explanations on those parameters?

It is also helpful if you can add the reference of the paper where the algorithm is derived.

skbeam/core/correlation.py
ordirules: thanks!

skbeam/core/correlation.py
ordirules: that was a residual parameter. I went through and cleaned up the docstrings, and removed these residual params.
I added a reference to the paper as well in the docstring.

ordirules: Also, see the [notebook](https://github.com/ordirules/notebooks/blob/master/CrossCorrelator.ipynb) referenced above, it contains sample usage cases with figures.

ordirules:
added cross correlation between two images. Example of this can be found [here](https://github.com/ordirules/notebooks/blob/master/CrossCorrelator.ipynb) (last figure).

skbeam/core/correlation.py
```diff
@@ -842,3 +850,322 @@ def one_time_from_two_time(two_time_corr):
...
+        or a list of non-negative integer id's to compute cross-correlations
+        separately on.
+
+        The symmetric averaging scheme introduced here is inspired by a paper
+        from Schätzel, although the implementation is novel in that it allows
+        for the usage of arbitrary masks. Reference to Schatzel's original
+        paper is here:
+            Schätzel, Klaus, Martin Drewel, and Sven Stimac. “Photon
+            correlation measurements at large lag times: improving statistical
+            accuracy.” Journal of Modern Optics 35.4 (1988): 711-718.
```
licode: thanks for adding this.  You may modify it a little bit like https://github.com/scikit-beam/scikit-beam/blob/master/skbeam/core/fitting/background.py#L114  Or check numpy docstrings.

skbeam/core/correlation.py
```diff
@@ -842,3 +850,322 @@ def one_time_from_two_time(two_time_corr):
...
+        if img1.ndim == 2:
+            axes = -2, -1
+        else:
+            axes = -1,
+
+    if img2 is None:
+        img2 = img1
+
+    imgc = fftshift(ifft2(fft2(img1, axes=axes) *
+                    np.conj(fft2(img2, axes=axes)), axes=axes).real)
```
licode: Is this similar to fftconvolve in scipy  https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.fftconvolve.html#scipy.signal.fftconvolve

skbeam/core/correlation.py
ordirules: thanks, done

skbeam/core/correlation.py
ordirules: thanks, I had not seen that. It's not well documented but after playing around, I found it calculates the convolution of the first image with the reflected image, basically my line without the complex conjugate taken (which is not correct for this application):
```python 
imgc = fftshift(ifft2(fft2(img1, axes=axes) *
                    fft2(img2, axes=axes), axes=axes).real)
```
One fix is I can index a mirrored version of the second image. I did just that, but had to figure out some indexing tricks to account for both cases of 1D and 2D. Not sure if this is the best way. It works equivalently after a test with the same notebook linked in this PR. Nice thing of this is that it removes the dependency on dimension in the correlation component.
Also, had to set mode to 'same'. Again, the documentation doesn't explain it. I'll try to go through their source code later to see what each mode really means (there are subtleties such as wrap around etc).

My personal feeling is I'd prefer the fftconvol-free version, since I don't trust it and have never really used it, but I'm biased. Do you have any input on this (maybe @danielballan or @tacaswell )? I will try to read through their source code later. thanks!

ordirules: I vote using fftconvol. For the record, computation as of now is:

`ret = (np.fft.irfftn(sp1 * sp2, fshape)[fslice].copy())`

which is FFT^-1(FFT(A)*FFT(B))

but where the data set has been reshaped to a size more efficient for FFT (speed of calculation is highly dependent upon array size being a multiple of some magic numbers). Makes more sense to use. (there's an if statement to handle complex arrays as well)

licode: Thanks for pointing out why fftconvol is much more efficient on FFT calculation. good to know that.

skbeam/core/correlation.py
```diff
@@ -1086,21 +1075,20 @@ def _cross_corr(img1, img2=None, axes=None):
...
 
-    imgc = fftshift(ifft2(fft2(img1, axes=axes) *
-                    np.conj(fft2(img2, axes=axes)), axes=axes).real)
+    if img1.shape != img2.shape:
+        errorstr = "Image shapes don't match. "
+        errorstr += "(img1 : {},{}; img2 : {},{})"\
+            .format(*img1.shape, *img2.shape)
+        raise ValueError()
+
+    reverse_index = (*((slice(None, None, -1),)*ndim),)
```
licode: I guess this will fail in python2 which we currently are still supporting. Maybe try this
```
reverse_index = [slice(None, None, -1) for i in range(ndim)]
```

licode:
This looks good to me, except that some simple tests are needed. Then this should be ready to go. Good work!

ordirules:
I added some code to keep track of the shift positions. It should be a minor change.
I have not added a test but could a little later if needed?

Also, added example [here](https://github.com/scikit-beam/scikit-beam-examples/pull/66)
Let me know if this is good!

ordirules:
I did not reference @yugangzhang, thanks @yugangzhang for contributions!

yugangzhang:
@ordirules thanks! Nice examples!

ordirules:
Thanks @licode ! :-)

PR #489, opened 2017-02-04
ENH : added construct_rphi_image

ordirules:
Suggesting to add some way of reprojecting an image from polar coordinates to rectangular. It's mask aware.

Uses the quick RectBivariateSpline from scipy. Note, uses a trick where it runs the same code on mask. If it finds a pixel is lower than a certain threshold, it masks and does not return result (i.e. find edge pixels that are wrongly interpolated and ignores them).

Not sure if this is best way, but I'd find this useful 
If you know something better let me know thanks! :-)

See this [suggested example](https://github.com/ordirules/scikit-beam-examples/blob/12e805a3392604480840a25e9e268beafd3234d0/demos/construct_rphi_image/construct_rphi_image.ipynb)


ordirules:
note : tests not written, if this is acceptable, i will add them. thanks!
@licode what do you think? thanks!

skbeam/core/image.py
```diff
@@ -170,3 +171,64 @@ def construct_circ_avg_image(radii, intensities, dims=None, center=None,
...
+    radial_val = utils.radial_grid(center, dims, pixel_size)
+    # keep in domain 0 to 2*pi
+    angle_val = utils.angle_grid(center, dims, pixel_size) % (2*np.pi)
+    # want to quickly interpolate, but remain mask aware.
+    # as a trick, interpolate both the image and the mask
+    # whatever in the mask gets interpolated below 1, we know
+
+    # the data is affected by mask, mask it out.
+    new_img = interp_obj(radial_val, angle_val, grid=False)
+    new_mask = interp_obj_mask(radial_val, angle_val, grid=False)
```
tacaswell: Do this with `np.nan` it will be cleaner and ensure that any pixel in the output that had _any_ input from a masked pixel of the input will be masked in the output.

skbeam/core/image.py
```diff
@@ -170,3 +171,64 @@ def construct_circ_avg_image(radii, intensities, dims=None, center=None,
...
+    w = np.where(new_mask < .99)
+    new_mask[w] = 0
```
tacaswell: Don't need to do this.

skbeam/core/image.py
```diff
@@ -170,3 +171,64 @@ def construct_circ_avg_image(radii, intensities, dims=None, center=None,
...
+                             "make sense and may lead to unexpected results.")
+        # round up, also take into account pixel size change
+        maxr_y, maxr_x = (int(np.max(radii/pixel_size[0])+.5),
+                          int(np.max(radii/pixel_size[1])+.5))
+        dims = 2*maxr_y+1, 2*maxr_x+1
+
+    if center is None:
+        center = (dims[0]-1)/2., (dims[1] - 1)/2.
+
+    if mask is None:
```
tacaswell: if no mask, skip interpolating an array of 1s

skbeam/core/image.py
```diff
@@ -170,3 +171,64 @@ def construct_circ_avg_image(radii, intensities, dims=None, center=None,
...
+        a polar coordinate image. Assumes a 2D array of data. If data is
+        missing, use mask.
+
+        radii : 1d array of coordinates, ascending order. The radii
+            must be in units of pixels.
+
+        angles : 1d array of coordinates, ascending order
+            must be in units of radians
+            domain *must* be 0 to 2*pi monotonically increasing (or a subset)
+            if it is not, code will *try* to repair this
```
tacaswell: I don't see how it tries to do this cleanup

ordirules: this was an incomplete thought... thanks, corrected, please see further comment

skbeam/core/image.py
ordirules: good idea. For some reason, doing this with RectBivariateSpline leads to all np.nans (even by setting degree of interpolation to kx=ky=1, not 3). I would have assumed that linear interpolation would just averaged neighboring pixels (or maybe a few negihboring pixels but not more).
I have changed this to use a simpler version of linear interpolation. I couldn't find it in any library, so I modified some code from stackoverflow. I retrofitted it to work with dimensions that wrap around. I am not sure if this is the best approach yet. Will write tests later.

If you had comments I would like to hear them, in case I am taking a longwinded approach... thanks for the review

ordirules:
Do not merge. (Needs to be modified to accept uniform bins)

ordirules:
Sorry for delay, needed an undistracted chunk of time...
I went through the scipy docs in more detail. `RectBivariateSpline` was the wrong interpolation technique to use here. I went with `RegularGridInterpolator` instead. Consequently, I removed my bilinear interpolation code that I had added. (I should have gone through scipy docs *first* before StackOverflow ;-) )

Note, more complex interpolation techniques cannot be used with this algorithm. The reason is that these more complex techniques involve arithmetic with a lot more neighboring pixels (some even does use all, probably via some convolution filter...). If *any* of these pixels are `np.nan`, then the pixel is ignored. For those more complex interpolation techniques, I'd suggest using what I had proposed before, which I will re-write here:
```python
    interp_obj = RectBivariateSpline(radii, angles, image)
    interp_obj_mask = RectBivariateSpline(radii, angles, mask)

    new_img = interp_obj(radial_val, angle_val, grid=False)
    new_mask = interp_obj_mask(radial_val, angle_val, grid=False)
    w = np.where(new_mask < .99) # unfortunately needs to be slightly less than 1....
    new_mask[w] = 0
    new_img[w] = 0
```
However, this is subject to contention. For ex : how many neighboring pixels must have been included to accept the interpolation as okay? etc...


Anyway, I'd like to squash all these commits because they contain code we never used and never will. @tacaswell is this okay or do you think we should keep them anyway?

As far as I can tell, it should be good. I plotted results before added their values to the tests and they look great. Again, see example [here] (https://github.com/ordirules/scikit-beam-examples/blob/8f3404da413845becf1df1a9edb94aad67ae385d/demos/construct_rphi_image/construct_rphi_image.ipynb)

thanks for the advice @tacaswell you really helped! :-)

ordirules:
I found a small bug when modifying this [example](https://github.com/scikit-beam/scikit-beam-examples/pull/68).

Anyway, it's ready to go now, @tacaswell ?

Also, @danielballan , here is the code the example would run.

Thanks!

ordirules:
@tacaswell any update on this? you requested changes, just curious to see what your status is.

thanks!

skbeam/core/image.py
```diff
@@ -170,3 +171,121 @@ def construct_circ_avg_image(radii, intensities, dims=None, center=None,
...
+    #   iv. [.12, ..., 2*pi, ...] -> [0, ...]
+
+    # 1.a : subtract minimum
+    anglemin = angles[0]
+    angles -= anglemin
+    # 1.b : modulo 2*pi
+    angles = angles % (2*np.pi)
+    # 1.c : find any extra cross-overs and ignore them
+    adiff = np.where(np.diff(angles) < 0)[0]
+    if len(adiff) > 0:
```
tacaswell: I would just raise here, silently dropping data is no good.

skbeam/core/image.py
```diff
@@ -170,3 +171,121 @@ def construct_circ_avg_image(radii, intensities, dims=None, center=None,
     CIMG = np.interp(radial_val, radii, intensities, right=0)
 
     return CIMG
+
+
+def construct_rphi_avg_image(radii, angles, image, mask=None,
+                             center=None, shape=None, pixel_size=(1, 1)):
+    '''  Construct a 2D image in rectangular coordinates from
```
tacaswell: First line should be complete thought.  Add rest as paragraph below.

skbeam/core/image.py
ordirules: I raise now. Just in case, I add an error string highlighting a common problem.
(`np.linspace(0, 2*np.pi, N)` won't work since the last point make the domain larger than 2*pi, but `np.linspace(0, 2*np.pi N, endpoint=False)` works)

ordirules:
thanks @tacaswell changes added

ordirules:
thanks @tacaswell!

PR #490, opened 2017-02-27
WIP: New fun for one-time corrleation using varible time stamps

sameera2004:
Not yet ready to review, I will add a test and example notebook.


ericdill:
Hi @sameera2004  -- If this PR is still of interest, can you test this new functionality? It would also be great to include an example notebook if possible

PR #491, opened 2017-02-28
Time bin

sameera2004:
Bin the time differences into all possible bins

### DIFF
--- a/skbeam/core/utils.py
+++ b/skbeam/core/utils.py
@@ -1271,3 +1271,35 @@ def bin_grid(image, r_array, pixel_sizes, statistic='mean', mask=None,
     bin_centers = bin_edges_to_centers(bin_edge)
 
     return bin_centers, int_stat
+
+
+def time_binning(time_diff):
+    """
+    Bin the time differences into all possible bins
+
+    Parameters
+    ----------
+    time_diff : list
+        time difference
+        ex: time difference from the previous image
+            for each image as a list
+
+    Returns
+    -------
+    time_bin : list
+        time bin for each delay time according to time bins
+    all_lags : array
+        all possible time lags from the time difference
+
+    """
+    time_bins = []
+    all_lags = []
+    for c, t in enumerate(time_diff):
+        time_b = []
+        stamp = 0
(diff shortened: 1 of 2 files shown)


skbeam/core/utils.py
```diff
@@ -1271,3 +1271,36 @@ def bin_grid(image, r_array, pixel_sizes, statistic='mean', mask=None,
...
+
+    """
+    time_bins = []
+    all_lags = []
+    ff_p = '.' + str(ff) + 'f'
+    for c, t in enumerate(time_diff):
+        stamp = 0
+        time_b = []
+        for item in (time_diff[:c+1][::-1]):
+            time_b.append(float(format(stamp, ff_p)))
```
danielballan: Nice docstrings! But can you explain why the floats need to be converted to strings?

danielballan: Oh, I see, you're using string conversion as a binning mechanism. There are standard ways to do this that are more performant and flexible than string conversion and truncation. See `numpy.digitize`.

sameera2004: @danielballan Thanks for the comments. I don't think `np.diigtize` will work here because I don't have a control of the input array to bin and the array of bins. Here I am converting result to string to get rid of floating point issues. This will be very helpful for one-time correlation functions with variable time stamps. Any better ways than this?

danielballan: In general, appending a numpy array to a Python list one and a time and converting floats to strings in order to perform a numerical operation is a red flag that there must be a better way. This will become very slow for long time series. There is, for example, `np.round` which would accomplish this in a vectorized way.

sameera2004: @danielballan yes sure I agree, np.round was my first choice. But it didn't work. So far this was the better way I found. np.round documentation also they are talking about this issue "floating point issues".  I will do more research. Thanks for the suggestions.

licode:
I just fixed some bugs in tests, which can now properly cover all the functions. Please rebase this PR and send it again. @sameera2004

ericdill:
@licode @sameera2004 -- What's the status on this one?

PR #494, opened 2017-03-16
New and improved Cross correlator

ordirules:
I met with Mark Sutton and we discussed the CrossCorrelator code. He had a lot of input and useful advice. He wrote some modifications that made the code 40% faster. I went through them and added some comments to help make them easier to understand for a future contrbutor.

*Changes*
    1. Used more efficient indexing tricks to index in images
    2. Removed the necessity of expanding the image (`fftconvol` does this for free), got rid of `_crop_from_mask`
    3. Removed the `wrap` parameter (since `fftconvol` expands the image to speed up computations,
        wrapping is not possible). This parameter was never used so it is simply ignored.
    4. Modified tests for 2D and 1D `CrossCorrelator`. New code results are identical to old,
        with the exception that new code more efficiently wraps the image.
    5. Removed the memory intensive allocation of arrays for the computations during
        the object initialization.
       Instead, the arrays are allocated and de-allocated per computation (per bin id).
       This is likely close in expense as zero-ing the array, which happened regardless.
    6. Removed `self.ids`. Not very useful.

*The testing*

I went through them and made sure they pass tests so I changed them. They didn't at first, but the reason is minor (basically, my old version's array was 1 or 2 pixels larger, because I was less efficient and including zero padded regions). I also made sure to plot the test results and compare the differences. Again they're minor. I also tested in my own code, and the (notebook example)[https://github.com/scikit-beam/scikit-beam-examples/blob/master/demos/spatial_correlation/Spatial_Correlations.ipynb].

*The future*

I plan on also adding a feature that allows one to force the output of the correlations to be regular. For example, right now 2D correlations return lists of 2D arrays. They're varying shapes, but normally all we're interested in is in a, say 20x20 subregion of the array. However, given the amount of code here, I figured I would stop here for this PR. If you think we should wait for this, I'm also fine.

Thanks for all the support!

### DIFF
--- a/skbeam/core/correlation.py
+++ b/skbeam/core/correlation.py
@@ -1101,13 +1101,11 @@ class CrossCorrelator:
             if 'regular' in normalization:
                 if self_correlation:
                     ccorr /= self.maskcorrs[i] * \
-                        np.average(tmpimg[ppiis,
-                        ppjjs])**2
+                        np.average(tmpimg[ppiis, ppjjs])**2
                 else:
-                    ccorr /= self.maskcorrs[i] * \
-                        np.average(tmpimg[ppiis,
-                        ppjjs])*np.average(tmpimg2[ppiis,
-                        ppjjs])
+                    ccorr /= self.maskcorrs[i] *\
+                             np.average(tmpimg[ppiis, ppjjs]) *\
+                             np.average(tmpimg2[ppiis, ppjjs])
             ccorrs.append(ccorr)
 
         if len(ccorrs) == 1:
--- a/skbeam/core/correlation.py
+++ b/skbeam/core/correlation.py
@@ -920,41 +920,59 @@ class CrossCorrelator:
         if mask is None:
             mask = np.ones(shape)
 
-        # pii, pjj : the list of xindex, yindex into original image
-        # pi, pj : the list of xindex, yindex into smaller image for FFT
+        # row_indices, col_indices : the list of xindex, yindex into original image
+        # row_subindices, col_subindices : the list of xindex, yindex into smaller image for FFT
         # initialize subregions information for the correlation
         # first find indices of subregions and sort them by subregion id
-        pii, pjj = np.where(mask != 0)
-        bind=mask[pii, pjj]
-        ord=np.argsort(bind)
-        bind=bind[ord];pii=pii[ord];pjj=pjj[ord] #sort them all
+        
+        row_indices, col_indices = np.where(mask != 0)
+        bind = mask[row_indices, col_indices]
+        order = np.argsort(bind)
+        bind = bind[order]
+        row_indices = row_indices[order]
+        col_indices = col_indices[order]  # sort them all
 
-        #make array of pointers into position arrays
+        # make array of pointers into position arrays
         # bind = [1,1,1,2,2,2,2,3,3,3...]
-        pos=np.append(0,1+np.where(np.not_equal(bind[1:], bind[:-1]))[0])
-        pos=np.append(pos,len(bind))
-        self.pos=pos
-        self.ids = bind[pos[:-1]]
-        self.nids = len(self.ids)
+        index_starts = 0
+        inds_tmp = 1 + np.where(np.not_equal(bind[1:], bind[:-1]))[0]
+        index_starts = np.append(index_starts, inds_tmp)
+        index_starts = np.append(index_starts, len(bind))
+        self.index_starts = index_starts
+        # not necessary, but keep maybe for futur debugging
+        # self.ids = bind[index_starts[:-1]]
+        self.nids = len(index_starts[:-1])
         # finds the shape of the subregions
-        sizes=np.array([[pii[pos[i]:pos[i+1]].min(),pii[pos[i]:pos[i+1]].max(), \
-             pjj[pos[i]:pos[i+1]].min(),pjj[pos[i]:pos[i+1]].max()] \
-                 for i in range(self.nids)])
-        #make indices for subregions arrays and their sizes
-        pi=pii.copy()
-        pj=pjj.copy()
+        shapes = list()
+        for i in range(self.nids):
+            index_start, index_end = index_starts[i], index_starts[i+1]
+            row_min = row_indices[index_start:index_end].min()
+            row_max = row_indices[index_start:index_end].max()
+            col_min = col_indices[index_start:index_end].min()
+            col_max = col_indices[index_start:index_end].max()
+            shapes.append([row_min, row_max, col_min, col_max])
+        shapes = np.array(shapes)
+
+        #make indices for subregions arrays and their shapes
+        row_subindices = row_indices.copy()
+        col_subindices = col_indices.copy()
         # subtract the minimum index, this trick 
         # will then index into some other array properly
         for i in range(self.nids):
-            pi[pos[i]:pos[i+1]] -= sizes[i,0]
-            pj[pos[i]:pos[i+1]] -= sizes[i,2]
+            index_start, index_stop = index_starts[i], index_starts[i+1]
+            row_subindices[index_start:index_stop] -= shapes[i,0]
+            col_subindices[index_start:index_stop] -= shapes[i,2]
         # now save to object
-        self.pi=pi; self.pj=pj; self.pii=pii; self.pjj=pjj
+        self.row_subindices = row_subindices
+        self.col_subindices = col_subindices
+        self.row_indices = row_indices
+        self.col_indices = col_indices
 
-        # redefine sizes to be the shapes of the subregions
-        sizes = 1+(np.diff(sizes, axis=-1)[:,[0,2]])  #make sizes be for regions
-        self.sizes=sizes.copy() # the shapes of each correlation
-        #WE now have two sets of positions of the subregions (pi,pj) in subregion
+        # redefine shapes to be the shapes of the subregions
+        # make shapes be for regions
+        shapes = np.array(1 + (np.diff(shapes, axis=-1)[:,[0,2]]))
+        self.shapes = shapes
+        # We now have two sets of positions of the subregions (pi,pj) in subregion
         # and (pii,pjj) in images. pos is a pointer such that (pos[i]:pos[i+1])
         # is the indices in the position arrays of subregion i.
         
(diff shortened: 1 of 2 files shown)


skbeam/core/correlation.py
```diff
@@ -1027,55 +1062,74 @@ def __call__(self, img1, img2=None, normalization=None):
...
             if self_correlation:
-                ccorr = _cross_corr(self.tmpimgs[i])*(self.maskcorrs[i] > 0)
+                ccorr = _cross_corr(tmpimg)
             else:
-                ccorr = _cross_corr(self.tmpimgs[i], self.tmpimgs2[i]) * \
-                        (self.maskcorrs[i] > 0)
+                ccorr = _cross_corr(tmpimg, tmpimg2)
 
-            # now handle the normalizations
+            # Note, in this code, non-overlapping regions will now get np.nan
```
ordirules: Note : I made a mistake here, non-overlapping regions will get `np.inf` not `np.nan`. Need to think about this later, just leaving a quick note. (Suggestions welcome)

licode:
Thanks for all the explanation. I will start reviewing this very soon.

licode:
Any paper or reference from Mark on how to make cross correlation faster? Or this is purely code optimization? Just want to get some big picture first before diving into it.

licode:
Another question related to forward analysis. Once you get this correlation curve, do you need to run statistical test, like to test stationary or not? Or mainly use physics function to fit?

ordirules:
@licode there is no paper. The speedup is mainly from some better indexing (explained in line 930). Also, there is less memory usage in that large arrays are not allocated in the initialization (rather arrays are allocated, and de-allocated on the fly).

The use case for these correlation curves is just to measure some shift.
It is similar to this:
http://scikit-image.org/docs/dev/auto_examples/transform/plot_register_translation.html#sphx-glr-auto-examples-transform-plot-register-translation-py

except, we want to see the cross-correlation, and we want it to handle large numbers of bins.

On top of that, I just noticed the linked paper on that page:
https://www.osapublishing.org/ol/abstract.cfm?uri=ol-33-2-156

They introduce three algorithms here. I don't have time now but I may take a look as well.
We could potentially also test these in the future. However, we've used this current version and have obtained some nice results. Writing up a paper that obtained some physically meaningful results to submit hopefully soon...

So I would say, this code is meant to speed up original code, and in the future, I'll review that paper linked as well, see how it performs.

licode:
I am quite familiar with that paper. I used that code from skimage for image registration. We can talk more on that later.

danielballan:
Just chiming in to say this is exciting!

skbeam/core/correlation.py
```diff
@@ -911,93 +909,130 @@ def __init__(self, shape, mask=None, normalization=None,
...
+
+        # initialize subregions information for the correlation
+        # first find indices of subregions and sort them by subregion id
+        pi, pj = np.where(mask != 0)
+        # typing not necessary, but it may be better to be explicit
+        # (especially if we cythonize in the future)
+        bind = mask[pi, pj].astype(np.uint32)
+
+        # sort them so that we can easily index into subregions with one number
+        order = np.argsort(bind)
```
licode: any reason to order the data here?

ordirules: yes, it's actual the most important step to allow quicker indexing.
When the data is ordered, you can make lots of assumptions, for example, the nonzero points of
`np.diff(bind[order])` are basically where the identifier changes. note `np.diff(bind)` (without ordering) doesn't give any information.

licode: got it, smart idea.

ordirules: you can credit Mark Sutton for that (he doesn't have a github). that's the integral piece he's basically added here, yeah pretty clever :-)

skbeam/core/correlation.py
```diff
@@ -911,93 +909,130 @@ def __init__(self, shape, mask=None, normalization=None,
...
             self.submasks.append(submask)
-            self.subpxlsts.append(np.where(submask.ravel() == 1)[0])
-            self.tmpimgs.append(tmpimg)
-            # make sure it's a copy and not a ref
-            self.tmpimgs2.append(tmpimg.copy())
+
             maskcorr = _cross_corr(submask)
-            # quick fix for finite numbers should be integer so
             # choose some small value to threshold
             maskcorr *= maskcorr > .5
```
licode: Does this threshold need to be changed? or set as fixed.

ordirules: I would leave it. The FFT introduces finite numbers even at zero values (as opposed to brute force convolution). But they're very small. Since we know results should be at least 1 or greater (since the submask is essentially 1's and 0's).

However, perhaps a better way to do this to round to the nearest integer. You can't just typecast to int though, I think. Reason is beacuse `maskcorr.astype(int)` for values like 2.99999 would become 2 when they should be 3 etc.)
Maybe it's something to try for a later pull request?

skbeam/core/correlation.py
```diff
@@ -1027,55 +1062,74 @@ def __call__(self, img1, img2=None, normalization=None):
...
+            # np.nan
             if 'symavg' in normalization:
                 # do symmetric averaging
-                Icorr = _cross_corr(self.tmpimgs[i] *
-                                    self.submasks[i], self.submasks[i])
+                Icorr = _cross_corr(tmpimg * self.submasks[i],
+                                    self.submasks[i])
                 if self_correlation:
-                    Icorr2 = _cross_corr(self.submasks[i], self.tmpimgs[i] *
+                    Icorr2 = _cross_corr(self.submasks[i], tmpimg *
```
licode: It looks to me Icorr and Icorr2 are the same? when you do cross correlation between A and B, it doesn't matter which one goes first?

skbeam/core/correlation.py
```diff
@@ -1113,77 +1167,6 @@ def _cross_corr(img1, img2=None):
     # fftconvolve(A,B) = FFT^(-1)(FFT(A)*FFT(B))
     # but need FFT^(-1)(FFT(A(x))*conj(FFT(B(x)))) = FFT^(-1)(A(x)*B(-x))
     reverse_index = [slice(None, None, -1) for i in range(ndim)]
-    imgc = fftconvolve(img1, img2[reverse_index], mode='same')
+    imgc = fftconvolve(img1, img2[reverse_index], mode='full')
```
licode: the size will not be the same if using full.

skbeam/core/correlation.py
ordirules: not exactly the same; One is reverse of the other, this is very important for symmetric averaging.
However, instead of running convolution again, we could do:
```python
Icorr2 = np.copy(Icorr[::-1, ::-1])
```
(I'm not sure if copy is necessary, I'll have to check)
I have to think about it to be certain and test just in case... (But I'm sure this is equivalent)
maybea cleaner? thanks ill check later :)

skbeam/core/correlation.py
licode: I see. I thought the convolution results might be complicated, but actually either 0, or larger than 1. Then 0.5 fits ok.

skbeam/core/correlation.py
licode: right, forgot the inverse relation. I think no need to do copy here.

skbeam/core/correlation.py
ordirules: It's true, but it works out here. I should explain this. So previously, the way I ran the computation using (`ifft(fft(a)*conj(fft(b)) )`), I had to double the image size, then run the convolution mentioned. The reason for doubling is to ensure that boundary points are not correlated together. 

When I switched to `fftconvol` as you suggested previously, I expanded the image and used `mode='same'` to ensure the output was same size as this doubled array.

However, what i did not appreciate (which Mark Sutton pointed out) is that `fftconvol` does this magic for you. It doubles your image (plus a few extra pixels to make it a magic number) to prevent this wraparound. Thus, expanding the image and using `method='same'` was more or less doing the same thing twice. All is needed to do is not expand the image, let fftconvol do it for you and set `method='full'`.

So in light of that, after dropping image doubling, I had to select `method=full` to ensure the image is roughly the same size. However, in this case, `fftconvol` is smarter than me and the array is actually consistently slightly smaller (i.e. it wraps the boundaries as tight as possible).

I had done testing and saw the array sizes were off by one or two pixels (but never more). I plotted both the 1D and 2D cases and compared to previous data and they're almost the same.

I hope I explained correctly? It is easier to explain in person.

skbeam/core/correlation.py
ordirules: thanks for the feedback, I'll change it to a reverse index a little later\
(need to be careful with 1D and 2D case)

licode: I vote let's leave it as it is for now. We may need to deal it differently in 1d and 2d.

skbeam/core/correlation.py
licode: good to know. thanks for the explanation.

skbeam/core/correlation.py
```diff
@@ -1027,55 +1062,74 @@ def __call__(self, img1, img2=None, normalization=None):
...
 
             if 'regular' in normalization:
-                # only run on overlapping regions for correlation
-                w = self.pxlst_maskcorrs[i]
-                ccorr[w] /= self.maskcorrs[i][w] * \
-                    np.average(self.tmpimgs[i].
-                               ravel()[self.subpxlsts[i]])**2
-
+                if self_correlation:
+                    ccorr /= self.maskcorrs[i] * \
```
ordirules: non-overlapping regions will be `np.inf` instead of `0` now. Maybe we should change this?

skbeam/core/correlation.py
```diff
@@ -911,93 +909,131 @@ def __init__(self, shape, mask=None, normalization=None,
...
-            self.subpxlsts.append(np.where(submask.ravel() == 1)[0])
-            self.tmpimgs.append(tmpimg)
-            # make sure it's a copy and not a ref
-            self.tmpimgs2.append(tmpimg.copy())
+
             maskcorr = _cross_corr(submask)
-            # quick fix for finite numbers should be integer so
             # choose some small value to threshold
             maskcorr *= maskcorr > .5
+            maskcorr[np.where(maskcorr == 0)] = np.nan
```
ordirules: I added this line so that 0 regions (basically where there is no correlation information) lead to `np.nan`. Previous convention was setting them to zero. But I think `np.nan` makes more sense.
This is a slight convention change but I don't think many have used it yet. I could revert to zero if better.

ordirules:
great thanks! :-)

skbeam/core/correlation.py
ordirules: the code has been changed to that non-overlapping regions are `np.nan`, for the record

licode:
sure. good work!

PR #495, opened 2017-03-30
fixed shape checking, added test

ordirules:
the shape check should work with lists or tuples (i.e. `(1,2) == [1,2]` will give false for shape checking, we need to change that)

also, I just noticed, there are no tests for RadialBinnedStatistic! I can add some later if no one objects. Could maybe be to this PR too if you feel it's fitting (in a few days). I use this library alot!

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -620,7 +620,7 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
             The values of the selected statistic in each bin.
         """
         # check for what I believe could be a common error
-        if values.shape != self.expected_shape:
+        if tuple(values.shape) != tuple(self.expected_shape):
             raise ValueError('"values" has incorrect shape.'
                              ' Expected: ' + str(self.expected_shape) +
                              ' Received: ' + str(values.shape))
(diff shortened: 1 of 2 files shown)


skbeam/core/accumulators/binned_statistic.py
```diff
@@ -620,7 +620,7 @@ def __call__(self, values, statistic=None):
             The values of the selected statistic in each bin.
         """
         # check for what I believe could be a common error
-        if values.shape != self.expected_shape:
+        if tuple(values.shape) != tuple(self.expected_shape):
```
ericdill: I don't understand the change here. `self.expected_shape` is a tuple according to the docstring in the `__init__` and `values.shape` is coming off of the numpy array, so that API is pretty well fixed. Are you finding that you hit an issue `if values.shape != self.expected_shape`?

ordirules: yeah basically I'm running into a case where I saved shape as a tuple in a database and when I retrieve it it's returned as a list (I used sqlite, although perhaps I made an error somewhere, this is from memory)
However, maybe it's good to still raise since it helped to hint an sqlite problem, so maybe better to leave as is?

ericdill: IIUC, the issue came from passing in the shape as a list into the `expected_shape` input parameter to the `__init__`?

ordirules: yep

ericdill: Ah, so the right place for this update would probably be in the `__init__`. That seems like a reasonable patch

ordirules: thanks! you're right, that makes much more sense. ill add now

ordirules:
thanks, maybe @chrisvam (but I know he's busy) or @CJ-Wright 
@licode also maybe you have time to look at this quick fix? thanks @ericdill !

licode:
LGTM.

licode:
Let's have another PR for tests of RadialBinnedStatistic.

PR #496, opened 2017-04-07
Made links work

habi:
I've removed the spaced which 'destroyed' the links.


ericdill:
Ah thanks for catching that @habi !

PR #497, opened 2017-05-16
Add k lines

licode:
Completed k lines are included. Although cross section values for some of the k lines are zero most of the time, it is better to add those lines to achieve better accuracy. Per request by hxn.


licode:
more comments on this?

PR #500, opened 2017-05-31
ENH: speed up binned_statistic

CJ-Wright:


CJ-Wright:
Ready for review!

CJ-Wright:
This might actually be slower in some cases, so don't merge yet.

licode:
Any progress on this? Please rebase this PR as there are some updates on test from master branch.

CJ-Wright:
It is still slower, although I might try to do some stuff with numba now that they have parallel for loops, so maybe that is faster?

licode:
numba is a good option. What did you compare with in terms of speed? What is the reference?

ordirules:
any progress on this? curious to see if jitting speeds things up

ericdill:
@CJ-Wright -- What's the status on this PR?

CJ-Wright:
This is currently dead in the water as the parallelization was not faster that numpy. @chiahaoliu and I were discussing some potential improvements to this, however we don't have anything concrete yet.

CJ-Wright:
It seems that `numba` supports `np.digitize` This means that we may be able to accelerate the init part? I'm not certain what the slowest part is yet.

CJ-Wright:
I've uploaded a new version which seems to be a bit faster. It uses the thread pool since we can do all the operations in memory (without needing to pickle things).
@ericdill Can you please remove the inactive label?

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -40,6 +40,7 @@
 
 import numpy as np
 from ..utils import radial_grid, angle_grid, bin_edges_to_centers
+from multiprocessing.dummy import Pool
```
CJ-Wright: Note that this is the threading pool.

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -316,9 +318,12 @@ def __call__(self, values, statistic=None):
                     null = np.nan
                 np.seterr(**old)
             self.result.fill(null)
-            for i in np.unique(self.xy):
+            p = Pool()
+
+            def op(i):
                 self.result[i] = statistic(values[self.xy == i])
 
+            [_ for _ in p.imap_unordered(op, np.unique(self.xy), 10)]
```
CJ-Wright: We get away with this because: 1) numpy releases the GIL (most of the time), 2) we know that each process will be touching a separate element in `self.result` so it is implicitly thread safe 3) all the important players are in global memory. We could move this to multiprocessing to avoid the GIL entirely but 1) processes take more time to spin up/manage 2) we would need to declare the shared nature of the memory. If someone has strong opinions/experience with this please let me know if one way is better than the other.

CJ-Wright:
Failing on un-related test (I don't have permissions to rebuild the failing one). @ericdill @tacaswell @ordirules Ready for review.

ericdill:
I have little experience with multiprocessing in Python and so will defer to others' expertise

skbeam/core/accumulators/binned_statistic.py
tacaswell: I do not understand the second (2) about 'declaring the shared nature of the memory'

tacaswell:
I am :-1: on this, I think it is better to handle this either at a higher level out side of the call (by doing a map of the data split into chunks by frame) or at a lower level and writing a numba/cython/cextension function that will take your array + mask and do the statistics in one loop over the data.

Have you profiled this code?  My naive guess is that the biggest expense is the `self.xy == i` comparison and then the fancy slicing to get the data for each bin.  If the statistic computation really is the bottle neck, passing in a callable that contains palatalization might be worth it.

The biggest win from this code is that we do not do all of the work in `__init__` everytime we call it (which generalizes pretty well!), I do not think it is worth going down the route of embedding thread pools in this class.  If we do go down this path, then it needs to be accompanied be part of a broader and articulated strategy for handling parallelism, not an ad-hoc one-off.

skbeam/core/accumulators/binned_statistic.py
CJ-Wright: We would need to have some sort of shared array object or give each process a copy of the array, which is not a problem in threading. https://docs.python.org/2/library/multiprocessing.html#sharing-state-between-processes

CJ-Wright:
I don't understand how this could be handled at "a higher level". My understanding is that to do the binning we need both the full data set and the full pixel to radial map. Lower level is attractive, and I tried working with [numba](https://github.com/numba/numba/issues/2538) but doesn't give much in speedup and causes a host of other issues. Numba doesn't take arbitrary functions in it's args, so we would need to perform a closure, which would mean that we can't cache the compiled results which is a rather nasty time sink, additionally we would have to either cache smartly, or re-compile the inner function at statistic definition time.

I have profiled the code. You are correct that the `self.xy == i` and the slicing are the bottle necks, hence the application of threading to spread out the work loads for those computations.

My understanding of the biggest win from this code was that the `__init__` did a fair amount of the work that generalizes to all the images which use the same pixel to something else map. This way the initial work of digitizing the map could be done once for a stack of images.

I am more than happy to get suggestions for other approaches (and even more happy to close this in deference to a better PR), however this is a major problem since this is critical code for many data processing workflows and is slow compared to other's approaches (eg. pyFAI).

CJ-Wright:
We could potentially stuff the `self.xy == i` results into a dict/list in the init so we don't have to compute that.
```python
d = {}
for i in np.unique(xy):
    d[i] = np.where(self.xy == i)
```
This may speed up computation (since we will only compute that map once) and be in the spirit of the class base approach (don't do things multiple times that a feature of the map and not the data).

skbeam/core/accumulators/binned_statistic.py
CJ-Wright: We might consider moving the `self.xy == i` info into the `__init__` method to create a list of pixels in a bin, since this can be calculated from the initial information and would speed up computation as we don't need to recompute it.

CJ-Wright: I'll see if I can get one in by COB Friday.

CJ-Wright:
ready for review! (On my computer this is about an order of magnitude faster for arbitrary statistics)

tacaswell:
:+1: This is a very clever implementation.

Could it be further sped up by pre-computing the `argsort` and `bincount`?

CJ-Wright:
Yes but that could show down users who only want to compute the mean and such. However we can cache it somewhere in the object of we compute it the first go round.

PR #502, opened 2017-06-04
FIX: it seems that fabio doesn't flip the masks on the way out so we …

CJ-Wright:
…shall

### DIFF
--- a/skbeam/io/fit2d.py
+++ b/skbeam/io/fit2d.py
@@ -1,5 +1,5 @@
 from __future__ import print_function, division
-
+import numpy as np
 from skbeam.io.save_powder_output import _create_file_path
 from fabio.fit2dmaskimage import Fit2dMaskImage
 import fabio
@@ -17,7 +17,7 @@ def fit2d_save(mask, filename, dir_path=None):
     dir_path: str, optional
         Path to the destination file
     """
-    saver = Fit2dMaskImage(data=~mask)
+    saver = Fit2dMaskImage(data=~np.flipud(mask))
     saver.write(_create_file_path(dir_path, filename, '.msk'))
 
 
@@ -35,5 +35,5 @@ def read_fit2d_msk(filename):
     ndarray:
         The mask as boolian array
     """
-    a = fabio.open(filename)
+    a = np.flipud(fabio.open(filename))
     return ~a.data.astype(bool)


skbeam/io/fit2d.py
```diff
@@ -17,7 +17,7 @@ def fit2d_save(mask, filename, dir_path=None):
     dir_path: str, optional
         Path to the destination file
     """
-    saver = Fit2dMaskImage(data=~mask)
+    saver = Fit2dMaskImage(data=~np.flipud(mask))
```
licode: Please put a note in function doc to mention this. It looks like a fix is needed in fabio?

licode:
I just fixed some bugs in tests, which can now properly cover all the functions. Please rebase this PR and send it again.

licode:
I also find a issue relate to fit2d_save, https://github.com/scikit-beam/scikit-beam/blob/master/skbeam/io/tests/test_fit2d_save.py#L49

I put it as xfail. Do you have any input on this? thanks.

licode:
any update here?

CJ-Wright:
Sorry for the delay I just need to check this against fit2d.

skbeam/io/fit2d.py
ordirules: yes explanation would help? why is this needed?

CJ-Wright:
See https://github.com/silx-kit/fabio/issues/161

skbeam/io/fit2d.py
```diff
@@ -35,5 +35,5 @@ def read_fit2d_msk(filename):
     ndarray:
         The mask as boolian array
     """
-    a = fabio.open(filename)
+    a = np.flipud(fabio.open(filename))
```
ordirules: this should not be flipped?

ordirules:
If I understand correctly, based on that thread, the saving needs to be flipped, but in principle the reading should not need to be. Can you confirm this @CJ-Wright ? (left a comment on the line to precise the position)

ordirules:
(also we should add a test to validate this)

ericdill:
@CJ-Wright -- no longer of interest?

CJ-Wright:
I think that the fabio issue has the correct answer. If things are to be fixed they should be fixed there.

PR #504, opened 2017-06-27
CI: python/numpy version change in ci

licode:


danielballan:
Minor note: You can use 'CI:' instead of 'DOC:' to indicate that this is a change to Continuous Integration settings, not documentation.

skbeam/core/tests/test_correlation.py
```diff
@@ -279,13 +279,13 @@ def test_CrossCorrelator1d():
                                        np.nan, 2.275846e-14]))
 
     assert_array_almost_equal(ycorr_1D_symavg[::20],
-                              np.array([3.347542,  1.54268227,  0.86220476,
+                              np.array([-5.3002753,  1.54268227,  0.86220476,
```
danielballan: Any idea why this was wrong in the first place?

licode: I didn't fix the issue related to divider in python2 and 3. But the value was changed back to original value later.

skbeam/io/tests/test_fit2d_save.py
```diff
@@ -44,8 +44,9 @@
 import numpy as np
 from skbeam.io.fit2d import fit2d_save, read_fit2d_msk
 from numpy.testing import assert_array_equal
+import pytest
 
-
+@pytest.mark.xfail(raises=IOError)
```
danielballan: Is there an open issue for this. cc @CJ-Wright

skbeam/core/tests/test_correlation.py
danielballan: I'm not following. I assume this test passed at some point in the past, so I'm wondering why one element would be off by 8.4.

danielballan: Oh, nevermind. By "the value was changed to back original value later" you mean you reverted the commit. Got it.

danielballan:
OK, the tests are actually being executed by Travis now. Thanks @licode for doing this round of drudgery. Merging!

PR #508, opened 2017-07-13
Add rpix

ordirules:
I've been back and forth about suggesting this but I think it saves time. Sometimes we want to supply the 2D image of bins rather than let RadialBinnedStatistic compute them.
One could then argue that why not used `BinnedStatistic1D`? There is still a small benefit, where we don't have to keep `ravel()`'ing everything.

What this saves is instead of having to do this:
```python
    rbinstat = BinnedStatistic1D(q_map.reshape(-1), bins=bins, mask=mask.ravel())
    sqy = rbinstat(image.ravel())
```

we can do this:
```python
    rbinstat = RadialBinnedStatistic(shape, mask=mask.ravel(), rpix=q_map)
    sqy = rbinstat(image)
```

(for the scientists here, qmap needs to be computed with Ewald curvature taken into account, so it's not a simple 2D geometric radius)

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -170,7 +170,7 @@ class BinnedStatisticDD(object):
 
         for i in np.arange(self.D):
             # Rounding precision
-            decimal = int(-np.log10(dedges[i].min())) + 6
+            decimal = int(-np.log10(np.maximum(1e-6, dedges[i].min()))) + 6
             # Find which points are on the rightmost edge.
             on_edge = np.where(np.around(sample[:, i], decimal) ==
                                np.around(self.edges[i][-1], decimal))[0]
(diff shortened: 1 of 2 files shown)


skbeam/core/accumulators/binned_statistic.py
```diff
@@ -661,6 +661,9 @@ def __init__(self, shape, bins=10, range=None,
         mask : 2-dimensional np.ndarray of ints, optional
             array of zero/non-zero values, with shape `shape`.
             zero values will be ignored.
+        r_map : the map of pixel radii for each pixel. This is useful when the
```
licode: rpix instead of r_map?

ordirules: oops thanks for catching that. actually i think `r_map` is a more suitable name. wha tod you think?

licode: r_map is better.

ordirules: cool. to be consistent, renamed them all (`rpix` was part of original notation). also renamed `phipix` to `phi_map`.

licode:
LGTM.

ordirules:
thanks! :)

PR #510, opened 2017-07-13
CI/TST: cover tests

CJ-Wright:


CJ-Wright:
Ready for review!

CJ-Wright:
codecov doesn't seem to trigger
https://travis-ci.org/scikit-beam/scikit-beam/jobs/253358742#L1307
@danielballan @ericdill thoughts?

ericdill:
What's the advantage of having the codecov.yml file?

.travis.yml
```diff
@@ -38,14 +38,12 @@ before_install:
...
   - export GIT_FULL_HASH=`git rev-parse HEAD`
-  - conda create -n testenv pip nose pytest python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY scipy scikit-image six coverage cython xraylib lmfit=0.8.3 netcdf4 flake8
+  - conda create -n testenv pip nose pytest python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY scipy scikit-image six coverage cython xraylib lmfit=0.8.3 netcdf4 flake8 pyFAI codecov -c conda-forge
   - source activate testenv
-  - pip install pyFAI
   # # need to build_ext -i for the tests so that the .so is local to the source
   # # code.  We could also setup.py develop, but I'm not sure if that is any
   # # better
   - python setup.py install build_ext -i
-  - pip install codecov
```
ericdill: I'd prefer to keep pulling this from pip. Keeping the conda-forge recipe up to date will always lag behind the codecov folks uploading their own package to pypi and I'm not sure that we really gain anything from pulling codecov from conda-forge

.coveragerc
```diff
@@ -4,7 +4,6 @@ source=skbeam
 omit =
     */python?.?/*
     */site-packages/nose/*
-    *test*
```
ericdill: What's the advantage of removing this test line from coveragerc?

.travis.yml
stuartcampbell: I agree

stuartcampbell:
I think the reason to have a `.codecov.yml` file is to just override the defaults.

.travis.yml
```diff
@@ -38,14 +38,12 @@ before_install:
 
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
-  - conda create -n testenv pip nose pytest python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY scipy scikit-image six coverage cython xraylib lmfit=0.8.3 netcdf4 flake8
+  - conda create -n testenv pip nose pytest python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY scipy scikit-image six coverage cython xraylib lmfit=0.8.3 netcdf4 flake8 pyFAI codecov -c conda-forge
   - source activate testenv
-  - pip install pyFAI
```
licode: conda installation of pyfai fails at numpy1.10 and py36.

.coveragerc
danielballan: Yeah, I had the same thought, and there was some discussion about this (which now I can't find...). The gist of is, including the tests in coverage is an extra way to help you notice if they aren't being _run_. Philosophically, I don't feel like tests themselves should count as "coverage". But given that we just ran a couple months without Travis even running the tests, CJ has a practical point here.

.travis.yml
CJ-Wright: So does the pip installation
https://travis-ci.org/scikit-beam/scikit-beam/jobs/253358153#L418

CJ-Wright: conda-forge does not still support numpy1.10

.coveragerc
CJ-Wright: The tests won't count against the coverage (assuming that this is combined with the .codecov.yaml), they will be assigned their own coverage value, note that this is ripped from [matplotlib](https://github.com/matplotlib/matplotlib/blob/master/ci/codecov.yml)
We also made this [change](https://github.com/NSLS-II/databroker/pull/145#issuecomment-285515858) in [databroker](https://github.com/NSLS-II/databroker/blame/master/.codecov.yml).

CJ-Wright:
The point of this as @danielballan pointed out was to have a separate coverage section for tests. This way we could see how many/much of the test suite actually runs, given our previous testing problems.

For an example of what this looks like, please see the end of this PR: https://github.com/xpdAcq/redsky/pull/13#

ericdill:
So to roll up what's left here:

1. install codecov from pypi not conda-forge
2. I'm convinced that having coverage on the test suite is reasonable for all the reasons presented in the comments of this PR. I think it would be better, from a maintenance perspective, if we didn't try to customize codecov with a yaml included in the repo. Better off just letting the coverage stats reflect whatever we configure with the .coveragerc. (tl;dr, delete codecov.yml)
3. rebase on to master since another PR that was merged underneath this one introduced merged conflicts

CJ-Wright:
1. is done https://github.com/scikit-beam/scikit-beam/pull/510/commits/06584a3fe78655316bfc79d33c10bb854106c6c9#diff-354f30a63fb0907d4ad57269548329e3R43
2. I don't know if I agree, but its fine I'll delete it.
3. Sure

ericdill:
@CJ-Wright Can you push up another commit on this branch based on what we discussed above? I think this is just about ready to go in

CJ-Wright:
Is this still pending some action from me? If so what?

ericdill:
Seems fine to me.

PR #513, opened 2017-08-06
ENH: added binmap to binned_statistic

ordirules:
Here's something that has been bothering me and @yugangzhang . When binning using `BinnedStatistic`, there is no straightforward reverse transformation from the flattened bin ids back to the bins. Here's a suggested improvement. Add a `binmap` member which computes the reverse transformation.
For example, it allows you to recover the bins from the top two plots in this figure, coming from the `self.xy` member of the bottom plot:

```python
# import numpy and matplotlib etc
from pylab import *
ion()

from skbeam.core.accumulators.binned_statistic import RPhiBinnedStatistic,\
        BinnedStatisticDD
from skbeam.core.utils import radial_grid, angle_grid

shape = np.array([100, 100])

R = radial_grid(shape/2, shape)
Phi = angle_grid(shape/2., shape)

img = R*np.cos(5*Phi)
rs = RPhiBinnedStatistic(img.shape)

res = rs(img)

binmap = rs.binmap.reshape((-1, *img.shape))

figure(2);clf()
subplot(2,2,1)
title("Rs")
imshow(binmap[0])
subplot(2,2,2)
title("Phis")
imshow(binmap[1])
subplot(2,2,3)
title("RPhibins")
imshow(rs.xy.reshape(shape))
```

This leads to the following image:
Image of bins
(From [imgur](http://imgur.com/a/McErx))
The top two plots are the r and phi bins, reverse transformed, and the last plot is the flattened bins `self.xy` that `binned_statistic` uses to perform the bin count.

NOTE : This is incomplete. Still need to add tests. I also tested 3D and it also works. I will also take the opportunity to add some general DD tests.

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -187,6 +187,30 @@ class BinnedStatisticDD(object):
         self.statistic = statistic
 
     @property
+    def binmap(self):
+        ''' Return the map of the bins per dimension.
+                i.e. reverse transformation of flattened to unflattened bins
+
+            Returns
+            -------
+            D np.ndarrays of length N where D is the number of dimensions
+                and N is the number of data points.
+        '''
+        N, = self.xy.shape
+        binmap = np.zeros((self.D, N))
+        denominator = 1
+
+        for i in range(self.D):
+            ind = self.D - i - 1
+            subbinmap = (self.xy // denominator)
+            if i < self.D - 1:
+                subbinmap = subbinmap % self.nbin[self.ni[ind - 1]]
+            binmap[ind] = subbinmap
+            denominator *= self.nbin[self.ni[ind]]
+
+        return binmap
+
+    @property
     def flatcount(self):
         # Compute flatcount the first time it is accessed. Some statistics
         # never access it.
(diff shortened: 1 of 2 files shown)


CJ-Wright:
Would it be also possible to capture this information during the statistic computation? This way it doesn't need to be recomputed.

ordirules:
we could. My reasoning for calculating on the fly is that it's not needed for the binning (and that's kind of the point of the flattened version `self.xy`, to save space). It will also likely be rarely used (mostly for physical review of bins, or debugging). So I would vote not to actually store it anywhere.

Maybe others have opinions? @licode or @danielballan, maybe @chrisvam as well?

stuartcampbell:
I agree, for certain cases it could be quite large (and hence why you've rebinned) so I agree with @ordirules and keep it light weight and calculate of needed. 

Just my 2 cents...

ordirules:
I think we can aim for 100% coverage on the tests for `binned_statistic`. I'd like to add them as I mentioned, but maybe should be on a separate PR? This one is good to go.

ordirules:
thanks @danielballan !

danielballan:
Thank _you_. This is a nice feature.

PR #515, opened 2017-09-01
WIP: twotheta in degrees

CJ-Wright:


stuartcampbell:
I would've expected it to be in radians - wouldn't it make more sense to just add the documentation that this is the case as radians ?

CJ-Wright:
I would have thought that most people are more familiar with twotheta in degrees (at least in the x-ray scattering/diffraction world?). Almost all the papers report it in degrees. https://www.google.com/search?q=two+theta+degrees+crystallography&source=lnms&tbm=isch&sa=X&ved=0ahUKEwjMlo715YTWAhVHvxQKHVF8CMoQ_AUICigB&biw=1599&bih=819

stuartcampbell:
I agree people are, but most computation libraries (e.g. numpy, scipy, etc)  use radians.

CJ-Wright:
I guess that is an interesting question, is this a computation library or a more scientist oriented library.

ericdill:
I think it would behoove us to keep this consistent with the rest of the scientific stack (i.e., use radians). Adding any barrier to moving between skbeam and numpy/scipy/etc without very careful consideration I think is a mistake

CJ-Wright:
Fair enough, all the more reason to make certain it is in the docs, but I expect the beamline scientists to expect two theta to be in radians.

ericdill:
@CJ-Wright if you feel like we should be doing two theta in degrees instead of radians then I urge you to submit an issue and ping the scikit-beam contributors. That way we can discuss it, see where we want to settle and be able to point to an issue where it was discussed for documentation purposes. After all, everyone likes making decisions by committee, right? 😎

skbeam/core/utils.py
```diff
@@ -1115,7 +1115,8 @@ def twotheta_to_q(two_theta, wavelength):
     Returns
```
ericdill: I can't reach the line I want here. Can you note that two_theta here is assumed to be in radians?

skbeam/core/utils.py
```diff
@@ -1115,7 +1115,8 @@ def twotheta_to_q(two_theta, wavelength):
...
     -------
     q : array
-        An array of :math:`q` values
+        An array of :math:`q` values in the inverse of the units
+        of ``wavelength``
```
ericdill: IIRC `q` here is in inverse angstroms so we can be more specific about the units of `wavelength`

skbeam/core/utils.py
```diff
@@ -1078,7 +1078,7 @@ def q_to_twotheta(q, wavelength):
     Returns
```
ericdill: Likewise I can't quite reach line 1073 to comment on it, but IIRC `q` here needs to be in inverse angstroms, right?

skbeam/core/utils.py
CJ-Wright: That is not automatically true. Some experiments are done in inverse nanometers

skbeam/core/utils.py
CJ-Wright: As above nm^-1 is also viable.

skbeam/core/utils.py
ericdill: Oh yeah my bad. That's definitely correct 🦆 . Ignore my comment above too about the inverse of this function

ordirules: if you want to be pedantic, it's technically radians per length as opposed to cycles per wavelength (this has always confused me)

ordirules: here, quick [wiki search](https://en.wikipedia.org/wiki/Wavenumber) for instance:
> Wavenumber can be used to specify quantities other than spatial frequency. In optical spectroscopy, it is often used as a unit of temporal frequency assuming a certain speed of light. In this context, it is the number of cycles—not radians—per unit length.

ericdill:
@jrmlhermitte @CJ-Wright what's the status on this? Anything left to do?

CJ-Wright:
I don't think so, although @sbillinge may have some thoughts on the radians/twotheta debate. But we can leave that for another PR.

CJ-Wright:
Can I please get a merge?

ericdill:
@stuartcampbell @jrmlhermitte what's the consensus here?

jrmlhermitte:
It is just a documentation and it is more precise so I am good with merging. If no one disagrees perhaps we can merge in a day?

ericdill:
oh i thought there were code changes. My mistake :)

CJ-Wright:
Just docs (we can discuss the units on everything later)

jrmlhermitte:
looks good, merging

PR #518, opened 2017-09-08
ENH: Added r_map option to RPhiBinnedStatistic

jrmlhermitte:
Same idea as #508 .
This is not ready yet. Will be a good opportunity to finally add tests to `RPhiBinnedStatistic`...

Only `r_map` can be overridden but not `phi_map`. We could add it, but I anticipate this to be only useful for the cases of Ewald curvature on 2D data sets.

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -546,7 +546,7 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
     """
 
     def __init__(self, shape, bins=10, range=None,
-                 origin=None, mask=None, statistic='mean'):
+                 origin=None, mask=None, r_map=None, statistic='mean'):
         """
         Parameters:
         -----------
@@ -572,6 +572,9 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
         mask : 2-dimensional np.ndarray of ints, optional
             array of zero/non-zero values, with shape `shape`.
             zero values will be ignored.
+        r_map : the map of pixel radii for each pixel. This is useful when the
+            detector has some curvature or is a more complex 2D shape embedded
+            in a 3D space (for example, Ewald curvature).
         statistic : string or callable, optional
             The statistic to compute (default is 'mean').
             The following statistics are available:
@@ -593,7 +596,9 @@ class RPhiBinnedStatistic(BinnedStatistic2D):
         if origin is None:
             origin = (shape[0] - 1) / 2., (shape[1] - 1) / 2.
 
-        r_map = radial_grid(origin, shape)
+        if r_map is None:
+            r_map = radial_grid(origin, shape)
+
         phi_map = angle_grid(origin, shape)
 
         self.expected_shape = tuple(shape)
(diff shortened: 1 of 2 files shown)


skbeam/core/accumulators/tests/test_binned_statistic.py
```diff
@@ -6,7 +6,7 @@
 from numpy.testing import assert_array_equal, assert_array_almost_equal
 import numpy as np
 import scipy.stats
-from ...utils import bin_edges_to_centers
+from skbeam.core.utils import bin_edges_to_centers
```
ordirules: I personally don't think test files should do relative imports. Testing should occur from outside.

ordirules:
I kind of would like to have this in (if accepted) before adding the much needed tests for `RPhiBinnedStatistic` which have not been added. Can someone review? @licode?

thanks!

skbeam/core/accumulators/tests/test_binned_statistic.py
ordirules: would like an opinion from you @danielballan . @CJ-Wright  also suggested this should be a relative import, i agree more with him now as a test should be guaranteed to test the correct instance

licode:
Sure, I will take a look.

ordirules:
Thanks. I also posted adding more tests as an issue (#521). It's a minute detail but I think worth the effort.

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -572,6 +572,9 @@ def __init__(self, shape, bins=10, range=None,
         mask : 2-dimensional np.ndarray of ints, optional
             array of zero/non-zero values, with shape `shape`.
             zero values will be ignored.
+        r_map : the map of pixel radii for each pixel. This is useful when the
```
licode: add line to explain data type of r_map, like 2-dimensional np.ndarray, optional?

licode: r_map can be used to define the radius of each pixel relative to the origin in reciprocal space, like Ewald sphere?

skbeam/core/accumulators/tests/test_binned_statistic.py
licode: I think either one is ok.

skbeam/core/accumulators/binned_statistic.py
jrmlhermitte: thanks. and yes, I like your wording better. I have added it in if that's okay. thanks!

skbeam/core/accumulators/tests/test_binned_statistic.py
jrmlhermitte: I decided to leave the global import finally. The tests are run in a controlled environment anyway so it shouldn't affect it. The bonus of global imports is you can run this as a script in ipython. Sometimes a little quicker for debugging, at least for me.
(I know there's the pytest --pdb option but sometimes I find it convenient to modify the script and have it return variables and play with them further, plot things etc.)

licode:
LGTM.

jrmlhermitte:
thanks @licode !

PR #520, opened 2017-10-06
Docstrings latex

### LINKED ISSUE #516: Math is not rendering properly in the docs
Content in the docstrings like

```python
def q_to_twotheta(q, wavelength):
    r"""
    Helper function to convert q to two-theta.
    By definition the relationship is:
    ..math ::
        \sin\left(\frac{2\theta}{2}\right) = \frac{\lambda q}{4 \pi}
    thus
    ..math ::
        2\theta_n = 2 \arcsin\left(\frac{\lambda q}{4 \pi}\right
    Parameters
    ----------
    q : array
        An array of :math:`q` values
    wavelength : float
        Wavelength of the incoming x-rays
    Returns
    -------
    two_theta : array
        An array of :math:`2\theta` values
    """
    q = np.asarray(q)
    wavelength = float(wavelength)
    pre_factor = wavelength / (4 * np.pi)
    return 2 * np.arcsin(q * pre_factor)
```
in ``skbeam/core/utils.py`` is not rendering properly in the documentation at scikit-beam.github.io/scikit-beam [here](https://scikit-beam.github.io/scikit-beam/resource/api/core/utils.html#skbeam.core.utils.q_to_twotheta)

screenshot: 
screen shot 2017-09-01 at 8 37 15 pm

I have no idea what magic is needed to make this work. Taking a stab in the dark and hoping that @danielballan will have some documentation juju to fix this

ordirules:
Tries to resolve #516 
syntax should be `.. math::` and not `..math ::`
@ericdill @tacaswell @danielballan 

(Note : i haven't checked docstrings outside of utils.pyyet)

### DIFF
--- a/skbeam/core/utils.py
+++ b/skbeam/core/utils.py
@@ -1006,7 +1006,7 @@ def q_to_d(q):
 
     By definition the relationship is:
 
-    ..math ::
+    .. math::
 
         q = \\frac{2 \pi}{d}
 
@@ -1032,7 +1032,7 @@ def d_to_q(d):
 
     By definition the relationship is:
 
-    ..math ::
+    .. math::
 
         d = \\frac{2 \pi}{q}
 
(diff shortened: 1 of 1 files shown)


ordirules:
Did a quick `grep -ri math | less` in the main `scikit-beam` directory. I did not find any other syntax issues.
This is good to go.

ordirules:
np :-)

PR #524, opened 2017-11-29
ENH: speed up by caching bincount and argsort

CJ-Wright:
from suggestion by @tacaswell

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -100,7 +100,6 @@ class BinnedStatisticDD(object):
         """
 
         # This code is based on np.histogramdd
-        self.argsort_index = None
         try:
             # Sample is an ND-array.
             N, self.D = sample.shape
@@ -185,6 +184,7 @@ class BinnedStatisticDD(object):
             self.xy += Ncount[self.ni[i]] * self.nbin[self.ni[i + 1:]].prod()
         self.xy += Ncount[self.ni[-1]]
         self._flatcount = None  # will be computed if needed
+        self._argsort_index = None
         self.statistic = statistic
 
     @property
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -221,7 +221,7 @@ class BinnedStatisticDD(object):
 
     @property
     def argsort_index(self):
-        # Compute flatcount the first time it is accessed. Some statistics
+        # Compute argsort the first time it is accessed. Some statistics
         # never access it.
         if self._argsort_index is None:
             self._argsort_index = self.xy.argsort()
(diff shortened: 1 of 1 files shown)


skbeam/core/accumulators/binned_statistic.py
```diff
@@ -317,11 +319,14 @@ def __call__(self, values, statistic=None):
                 np.seterr(**old)
             self.result.fill(null)
 
-            # Sort by bin number
-            idx = self.xy.argsort()
-            vfs = values[idx]
+            if not self.argsort_index:
```
kalvdans: You can't convert a numpy array directly to a truth value:

`ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

Use `self.argsort_index is None` instead. Same for self.bincount below.

CJ-Wright: Ah yes I had forgot about numpy truthiness

CJ-Wright:
ready for more review!

kalvdans:
I have one more comment; bincount seems already be handled by self._flatcount
Can you use that variable instead?

CJ-Wright:
ready for more review

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -218,6 +219,14 @@ def flatcount(self):
             self._flatcount = np.bincount(self.xy, None)
         return self._flatcount
 
+    @property
+    def argsort_index(self):
+        # Compute flatcount the first time it is accessed. Some statistics
```
kalvdans: Substitute flatcount with argsort_index in the comment as well.

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -218,6 +219,14 @@ def flatcount(self):
             self._flatcount = np.bincount(self.xy, None)
         return self._flatcount
 
+    @property
+    def argsort_index(self):
+        # Compute argsort the first time it is accessed. Some statistics
+        # never access it.
+        if self._argsort_index is None:
+            self._argsort_index = self.xy.argsort()
```
ericdill: Is there a usage pattern where the user changes ``xy``? If so we'll have a cache invalidation problem here. From my brief read of this class it seems that  ``xy`` is only set in the ``__init__`, so if the user mucks with ``xy`` that could cause problems, but I'm also not sure that we can do anything to prevent that other than document it. :shrug:

CJ-Wright: We'd need cache invalidation on the `flatcount` too. If they are well versed enough to play with `xy` then I'd hope that they understand what that means in terms of the internals. Maybe leave this for another PR?

ericdill: Yeah I'm not proposing any new work here. I was curious about the thinking on cache invalidation. Apologies for not being clearer

ericdill:
Restarting the failed travis job. If that passes, then 👍

ericdill:
Did we ever figure out why the histogram tests seem to randomly fail?

CJ-Wright:
I think the problem was some floating point logic with numba which I removed.

PR #527, opened 2017-12-22
PKG: install required dependencies before setup()

mrakitin:
Here is a proposed update to install `numpy` and `cython` dependencies required to run `setup.py` in clean environments. This may possibly be failing on conda.


jrmlhermitte:
Looks good to me. Tested run on a test environment.

PR #529, opened 2017-12-30
TST: BinnedStats with mask doesn't match scipy.stats.binned_stats ...

CJ-Wright:
…with mask

### DIFF
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -124,8 +124,12 @@ class BinnedStatisticDD(object):
         # Select range for each dimension
         # Used only if number of bins is given.
         if range is None:
-            smin = np.atleast_1d(np.array(sample.min(0), float))
-            smax = np.atleast_1d(np.array(sample.max(0), float))
+            if mask is not None:
+                smin = np.atleast_1d(np.array(sample[mask].min(axis=0), float))
+                smax = np.atleast_1d(np.array(sample[mask].max(axis=0), float))
+            else:
+                smin = np.atleast_1d(np.array(sample.min(axis=0), float))
+                smax = np.atleast_1d(np.array(sample.max(axis=0), float))
         else:
             smin = np.zeros(self.D)
             smax = np.zeros(self.D)
@@ -158,9 +162,7 @@ class BinnedStatisticDD(object):
             # Would be better to do this using bincount "weights", perhaps.
             thissample = sample[:, i]
             if mask is not None:
-                thissample[mask == 0] = (self.edges[i][0] -
-                                         0.01 * (
-                                             1 + np.fabs(self.edges[i][0])))
+                thissample[~mask] = np.floor(self.edges[i][0] - 0.01 * (1 + np.fabs(self.edges[i][0])))
             Ncount[i] = np.digitize(thissample, self.edges[i])
 
         # Using digitize, values that fall on an edge are put in the
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -162,7 +162,8 @@ class BinnedStatisticDD(object):
             # Would be better to do this using bincount "weights", perhaps.
             thissample = sample[:, i]
             if mask is not None:
-                thissample[~mask] = np.floor(self.edges[i][0] - 0.01 * (1 + np.fabs(self.edges[i][0])))
+                thissample[~mask] = np.floor(
+                    self.edges[i][0] - 0.01 * (1 + np.fabs(self.edges[i][0])))
             Ncount[i] = np.digitize(thissample, self.edges[i])
 
         # Using digitize, values that fall on an edge are put in the
--- a/skbeam/core/accumulators/binned_statistic.py
+++ b/skbeam/core/accumulators/binned_statistic.py
@@ -125,8 +125,8 @@ class BinnedStatisticDD(object):
         # Used only if number of bins is given.
         if range is None:
             if mask is not None:
-                smin = np.atleast_1d(np.array(sample[mask].min(axis=0), float))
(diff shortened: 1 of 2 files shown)


CJ-Wright:
It seems the RadialPhi binned statistics are causing some issues in the tests, any thoughts?

jrmlhermitte:
There is something funny here. I don't see the same diff as running `git diff master` on my synchronized repo.

Seems like you've also removed `def argsort_index(self):`
is this just me?

ericdill:
@jrmlhermitte i think you're seeing an oddity with git here not @CJ-Wright's branch. I see the same thing you're seeing but if I rebase CJ's branch onto master then the apparent `argsort_index` deletion goes away 🤷‍♂️

skbeam/core/accumulators/tests/test_binned_statistic.py
```diff
@@ -269,6 +269,33 @@ def test_binmap3d():
...
     assert_array_almost_equal(rbinmap1[0][::1000], rbinmap2[0][::1000])
-    assert_array_almost_equal(rbinmap1[0][::1000], np.array([1, 10,  9,  8, 7,
-                                                             6,  5,  4,  3, 2,
+    assert_array_almost_equal(rbinmap1[0][::1000], np.array([1, 10, 9, 8, 7,
+                                                             6, 5, 4, 3, 2,
                                                              1]))
+
+
+@pytest.mark.parametrize(['stat', 'stat_f'], stats_list)
+def test_BinnedStatistics1D_mask(stat, stat_f):
```
ericdill: nit: maybe call this ``stat_func`` instead of ``stat_f``. Took me a few minutes to figure out what ``stat_f`` meant 😛

ericdill: Can you add a short description about what the test is testing (like a one-line test function docstring)

ericdill: Seems like another simple test you could do to make sure the mask is working would be to:

1. make an array of ones
2. make a mask that masks half of the array
3. run through the binning code
4. make sure that the resultant binned values are equal to half the size of the array.

That might be easier to reason about instead of somewhat duplicating the ``test_BinnedStatistics1D`` function

CJ-Wright:
I guess this PR was meant to raise a more philosophical question too: should we define the bins using the full `sample` or using not masked out portions of the `sample`? In scipy we don't have the option since the `sample` needs to be masked ahead of time. This PR goes with the latter option. However, I'm open to the former.

jrmlhermitte:
thanks @ericdill ! @CJ-Wright and I figured it out also a few min after i posted this. Github is doing some magic. :-D

And good catch @CJ-Wright !

I'd  be tempted to vote for the latter as well. Since we're supplying the mask with the bins from the beginning I think it would be more sensible to expect to bin over the valid pixels. But I could see a user that wants more reproducibility to want to supply different masks but expect the same binning. I would argue though that if they want that reproducibility, they should supply the bins directly. Either with
```python
mybinstat1 = BinnedStatistic(sample, bins=100)
# supply the bin edges from the statistic you want to the masked BinnedStatistic
mybinstat2 = BinnedStatistic(sample, bins=mybinstat1.bin_edges, mask=mymask)
```

what do you think? Is there anyone who uses this who may be open to discussion?

skbeam/core/accumulators/tests/test_binned_statistic.py
jrmlhermitte: I'm :+1: on reducing thinking time, sounds like a good change!

jrmlhermitte: I also agree with docstrings on tests :-), i think they're lacking!

jrmlhermitte: I the current test is fine, but perhaps documentation should be added to make it clear that the point of this test is to compare scipy to `binned_statistic`?

skbeam/core/accumulators/binned_statistic.py
```diff
@@ -158,9 +162,8 @@ def __init__(self, sample, statistic='mean',
             # Would be better to do this using bincount "weights", perhaps.
             thissample = sample[:, i]
             if mask is not None:
-                thissample[mask == 0] = (self.edges[i][0] -
-                                         0.01 * (
-                                             1 + np.fabs(self.edges[i][0])))
+                thissample[mask == 0] = np.floor(
```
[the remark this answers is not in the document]
jrmlhermitte: PS: a comment about this. This was a hack to make sure the masked values go out of range. We could consider a better alternative (but it would be nice to go through benchmarking before committing, and stash some of this somewhere).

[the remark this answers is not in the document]
CJ-Wright: The floor makes certain it is below the threshold. There were some nasty corner cases where it didn't work otherwise. I can remove it otherwise.

CJ-Wright:
@jrmlhermitte I like the idea of trying out different masks with the same binner. Although that seems to be mutually exclusive with this PR. Additionally that would require us to hold a bit more state (namely the original `sample`).

skbeam/core/accumulators/tests/test_binned_statistic.py
ericdill: @jrmlhermitte check out [this other test in the test suite](https://github.com/scikit-beam/scikit-beam/blob/master/skbeam/core/accumulators/tests/test_binned_statistic.py#L200-L222). It is quite similar to this one that seems like it is doing the heavy lifting of comparing scipy to binned_statistic. Given that test already exists it seems like we just want to make sure the masking code is working properly which could be a much simpler test with no duplicated code.  Another possible avenue for testing the masking code would be to update the linked test to take a mask parameter as part of the pytest parametrize fixture and then there'd be no duplicated code in two different tests

PR #533, opened 2018-01-30
WIP: Flatten Correlation

jrmlhermitte:
EDIT to elaborate more on this. I changed how the indexing is done in CrossCorrelator to be dask friendly. There should be no API change here.

Dask doesn't support fancy indexing yet:
```py
import numpy as np
import dask.array as da
arr= da.from_array(np.ones((10,10)), chunks=10)

In [29]: arr[[1,2], [1,2]]
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
NotImplementedError: Don't yet support nd fancy indexing
```
but flattening will work:
```py
In [30]: arr.ravel()[[1*10+1, 2*10+2]]
Out[30]: dask.array<getitem, shape=(2,), dtype=float64, chunksize=(2,)>
```

I think it might be good to work in terms of flattened arrays for now. So far works for me but I need to test further. This PR serves as a place holder. Do not merge.

### DIFF
--- a/optional-requirements.txt
+++ b/optional-requirements.txt
@@ -1,5 +1,4 @@
 lmfit==0.8.1
 scipy
 scikit-image
-xraylib
 netCDF4
--- a/skbeam/core/correlation.py
+++ b/skbeam/core/correlation.py
@@ -921,11 +921,14 @@ class CrossCorrelator:
             mask = np.ones(shape)
 
         self.shape = shape
-        # for 1d quick indexing
-        width = shape[0]
         self.ndim = len(shape)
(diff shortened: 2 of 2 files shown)


jrmlhermitte:
This passes tests and is just a rewrite on how the indexing is done so it's dask friendly.
It's ready to go.

jrmlhermitte:
This is outdated, and an attempt to change code to work with an external library (which has likely changed in the meantime), and the author chooses not to support this anymore. I will close, but feel free to take over if there is interest.

PR #536, opened 2018-08-02
Fix format string

ksunden:
Just changing `]` to `}`to properly use both arguments to the format function


ericdill:
thanks @ksunden !

PR #537, opened 2018-10-01
Trun element_list to a list of strings

### LINKED ISSUE #262: Turn string constant into a list of strings
It would make some sense to change this: `'non_fitting_values': {'element_list': 'Ar, Fe, Ce_L, Pt_M',` to be a list of strings instead of a single string in `skxray/fitting/base/parameter_data.py`

souravsingh:
Fixes #262


danielballan:
Thanks, @souravsingh. Apologies for the slow turnaround on this.

PR #538, opened 2018-12-02
Added .editorconfig

KOLANICH:


tacaswell:
I am :-1: on there here, see https://github.com/matplotlib/matplotlib/pull/12922 for longer comments.

danielballan:
I agree with [@WeatherGod's comment](https://github.com/matplotlib/matplotlib/pull/12922#issuecomment-443838399) on that same matplotlib issue, but I don't feel strongly, so I'm willing to close this.

PR #542, opened 2019-06-17
New-style pip API compat during setup

ronpandolfi:


setup.py
```diff
@@ -33,8 +33,12 @@ def _import(module, function=None, alias=None):
     try:
         _import(**dep)
     except ImportError:
-        import pip
-        pip.main(['install', lib])
+        try:
+            from pip import main # old API
+        except ImportError:
+            from pip._internal import main # new API
```
danielballan: I’m surprised that the new recommended approach would involve importing from a non-public internal module. Any idea if there is a way to do this using public APIs?

ronpandolfi: There is no officially supported public API for pip. 

https://groups.google.com/forum/#!topic/pypa-dev/JVTfS6ZdAuM

The alternative is `subprocess.call`.

https://stackoverflow.com/a/50255019

tacaswell:
Shouldn't we be using `setup_requires` instead?

ericdill:
Turns out we can just remove that whole code block (lines 10-38) and replace it with
```
import numpy as np
from Cython.Build import cythonize
```

That's what I did in #544 to get the docs building again. I dont remember the original reason we made the setup.py quite so complex. Unless anyone else remembers and has a compelling reason to switch back from these two import lines, I'd vote for dumping all of that code (lines 10-38). Then we dont need this PR either.

ericdill:
Going to close this for now. Re-open if we need to keep this block of importlib. Thanks for the PR @ronpandolfi !

Issue #543, opened 2019-07-02
Documentation page is down


ambarb:
On clicking on the documentation link
http://scikit-beam.github.io/scikit-beam/ 

There is a `Page not found error`.

ambarb:
It seems that the examples are also broken.

ambarb:
now it magically works -- the examples
Unsure if this is a repo issue or a GitHub infrastructure issue.

ericdill:
Looks like the last build on Jenkins just pushed an empty repo back to the gh-pages branch (this is what serves the scikit-beam docs. @danielballan I think may have been the last person to work on that auto docs build infrastructure?

https://github.com/scikit-beam/scikit-beam/tree/gh-pages?files=1

ericdill:
Docs are failing. travis has the following stack trace:

```
# make -C example notebooks
sphinx-build -b html -v -d _build/doctrees   . _build/html
Running Sphinx v2.1.0
making output directory... done
[autosummary] generating autosummary for: api_changes.rst, api_changes/2014-08-29-gridder.rst, api_changes/2014-09-03-gridder.rst, api_changes/2014-09-04-constants.rst, api_changes/2014-09-08_fit_quad_to_peak.rst, api_changes/2014-09-09-physics_peak.rst, api_changes/2014-09-12-physics_peak.rst, api_changes/2014-09-30-rcparams.rst, api_changes/2014-10-03-core.rst, api_changes/2014-11-12-fitting.rst, ..., resource/dev_guide/adding_files.rst, resource/dev_guide/doc_doc.rst, resource/dev_guide/git_guide.rst, resource/dev_guide/index.rst, resource/dev_guide/logging.rst, resource/dev_guide/workflow.rst, warnings.rst, whats_new.rst, whatsnew/0.5.rst, whatsnew/index.rst
[...]

PR #544, opened 2019-07-04
Various updates to get docs building again

ericdill:
Some things to sort through:

* Had to bump the pinning on lmfit to 0.8.3 since 0.8.1 is no longer available on the conda-forge channel on anaconda.org. This can, of course, be fixed by rebuilding 0.8.1 on the conda-forge feedstock, but I'd hope that it isn't actually necessary to do that and lmfit 0.8.3 might work. I remember @licode was working through updating his XRF code to work with newer versions of lmfit, but I'm not sure how that progressed. Oh, actually, lmfit 0.8.3 [_is_ the version that is being packaged on conda-forge](https://github.com/conda-forge/scikit-beam-feedstock/blob/master/recipe/meta.yaml#L36), so this change should be fine.

* Delete complex import logic instead just preferring to import the required libraries. It's not at all clear why this wasn't done in the first place. These changes to setup.py should render #542 unnecessary.

* Get rid of the "See Also" heading in one of the docstrings. Apparently that is what was causing the stack trace in #543

* The docs successfully build locally. There's no guarantee that they'll build on travis. Speaking of which, the travis config is a whole other ball of yarn that needs to be untangled and cleaned up.


tacaswell:
Looks green!  Lets see how this goes....Thanks @ericdill !

ericdill:
Hooray docs are fixed http://scikit-beam.github.io/scikit-beam/

danielballan:
Thanks, @ericdill!

PR #548, opened 2019-08-19
Add angular_cut() function

madelinegullen:
Hi, I'm a summer intern at BNL with @ambarb as my mentor. I was hoping to add what could be a useful function for image processing that takes angled row cuts of image data and returns the data from that row or set of rows. The information returned includes the angle of the cut taken, an array of the image data values at the points sampled, and an array of the radii relative to the center of the image for each point sampled.

Here is a demo notebook: https://github.com/madelinegullen/angular_cut/blob/master/_angular_cut_example.ipynb

### DIFF
--- a/skbeam/core/utils.py
+++ b/skbeam/core/utils.py
@@ -1300,9 +1300,9 @@ def bin_grid(image, r_array, pixel_sizes, statistic='mean', mask=None,
 
         Returns
         -------
-        data : tuple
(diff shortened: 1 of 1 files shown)


ambarb:
@danielballan I will try on my own and report back.  Please recommend someone else to also review this.  Does skbeam require 2 reviews or just one.

PR #550, opened 2019-08-20
Find center of rings using Hough Transform 

madelinegullen:
Here's an alternate find center rings function that uses the Hough Transform. It seemed to work better on soft edged images than using auto_find_center_rings(). It employs the circular Hough Transform to find candidate rings and then finds the center coordinate as the average of the circles' centers. 

My function is based on a sci_kit image Hough Transform notebook:  https://scikit-image.org/docs/dev/auto_examples/edges/plot_circular_elliptical_hough_transform.html?highlight=hough%20transform

Example notebook for my function:
https://nbviewer.jupyter.org/github/madelinegullen/find_center_rings/blob/master/example_find_center_rings.ipynb

### DIFF
--- a/skbeam/core/roi.py
+++ b/skbeam/core/roi.py
@@ -43,11 +43,13 @@ from __future__ import absolute_import, division, print_function
 import collections
 import scipy.ndimage.measurements as ndim
 from skimage.draw import line
-from skimage import img_as_float, feature, color, draw
+from skimage import img_as_float, feature, color, draw, transfrom
 from skimage.measure import ransac, CircleModel
 import numpy as np
 from . import utils
 import logging
+import matplotlib.pyplot as plot
+
 
 logger = logging.getLogger(__name__)
 
@@ -757,3 +759,65 @@ def auto_find_center_rings(avg_img, sigma=1, no_rings=4, min_samples=3,
         edge_pts_xy = edge_pts_xy[~inliers]
 
     return center, image, radii
+
+def find_center_rings_hough(image, num_rings, rad_range, intensity_thresh, sigma, v_range = None):
+    """
+    Take angled cut or multiple incremented cuts of a image data.
+
+    Parameters
+    ----------
+    image : ndarray
+        Image data
+    num_rings : int
+        Number of candidate circles to be searched for
+    rad_range : array
+        The min, max, and increment for range of candidate radii to be searched
+    intensity_thresh : array
+        Min and max of intensity to be used in canny filter
+    sigma : float
+        Standard deviation applied in canny filter
+    v_range : array
+        Min and max for matplotlib colormap
+
+    Returns
+    -------
+    center_pt : array
+        Average center of the candidate circles
+    radii: array
+        radii of the candidate circles
+
+    """
+
+    low_thresh = intensity_thresh[0]
+    hi_thresh = intensity_thresh[1]
+    sigma = sigma
+    edges = feature.canny(image, sigma, low_threshold= low_thresh, high_threshold= hi_thresh)
+
+    low = rad_range[0]
+    hi = rad_range[1]
+    step = rad_range[2]
+
+    # Range of possible radii
+    hough_radii = np.arange(low, hi, step)
+    hough_res = transform.hough_circle(edges, hough_radii)
+
+    # Selects the most prominent circles, up to num_rings
+    accums, cx, cy, radii = transform.hough_circle_peaks(hough_res, hough_radii,
+                                               total_num_peaks=num_rings)
+
+    # Draw them
+    fig, ax = plt.subplots(ncols=1, nrows=1, figsize=(10, 4))
+    Z = color.gray2rgb(image)
+    for center_y, center_x, radius in zip(cy, cx, radii):
+        circy, circx = draw.circle_perimeter(center_y, center_x, radius,
+                                        shape=Z.shape)
+        ax.scatter(x = circx,y =circy, s = 1, c = 'red')
+    center_pt = [np.average(cx), np.average(cy)]
+
+    ax.scatter(center_pt[0], center_pt[1], s =20, c = 'red')
+    if v_range != None:
+        ax.imshow(image, cmap="viridis", vmin = v_range[0], vmax = v_range[1])
+    else:
+        ax.imshow(image, cmap="viridis")
+    plt.show()
+    return center_pt,radii
--- a/skbeam/core/utils.py
+++ b/skbeam/core/utils.py
@@ -1300,9 +1300,9 @@ def bin_grid(image, r_array, pixel_sizes, statistic='mean', mask=None,
 
         Returns
         -------
-        data : tuple
-            For each line cut made, returns the angle of the cut, an array of the intensity values sampled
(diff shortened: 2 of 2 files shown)


skbeam/core/roi.py
```diff
@@ -757,3 +759,65 @@ def auto_find_center_rings(avg_img, sigma=1, no_rings=4, min_samples=3,
...
+    Returns
+    -------
+    center_pt : array
+        Average center of the candidate circles
+    radii: array
+        radii of the candidate circles
+
+    """
+
+    low_thresh = intensity_thresh[0]
```
danielballan: A nice pattern for this is:

```py
low_thresh, hi_thresh = intensity_thresh
```

It's more succinct, and more importantly it provides some validation. If the user mistakenly provides, `intensity_thresh=[1, 2, 3]`, the suggestion about would raise an error about an unexpected number of elements in the list (it wants exactly two) whereas the current implementation would silently ignore the extra value.

skbeam/core/roi.py
```diff
@@ -757,3 +759,65 @@ def auto_find_center_rings(avg_img, sigma=1, no_rings=4, min_samples=3,
...
+    hi_thresh = intensity_thresh[1]
+    sigma = sigma
+    edges = feature.canny(image, sigma, low_threshold= low_thresh, high_threshold= hi_thresh)
+
+    low = rad_range[0]
```
danielballan: The same pattern could be useful here. Or, to be even more succinct, you can do:

```y
np.arange(*rad_range)
```

if you like. The `*` "unpacks" the values into the function signature.

skbeam/core/roi.py
```diff
@@ -757,3 +759,65 @@ def auto_find_center_rings(avg_img, sigma=1, no_rings=4, min_samples=3,
...
+    hough_res = transform.hough_circle(edges, hough_radii)
+
+    # Selects the most prominent circles, up to num_rings
+    accums, cx, cy, radii = transform.hough_circle_peaks(hough_res, hough_radii,
+                                               total_num_peaks=num_rings)
+
+    # Draw them
+    fig, ax = plt.subplots(ncols=1, nrows=1, figsize=(10, 4))
+    Z = color.gray2rgb(image)
+    for center_y, center_x, radius in zip(cy, cx, radii):
```
danielballan: The function _returns_ once center position but internally uses separate centers here to draw the figure. Have you considered using `center_pt` here instead? The figure might look a little worse but it will be a more accurate representation of what information is flowing downstream to the next step in the analysis, which won't have access to `cx` and `cy`, only `center_pt` and `radii`.

madelinegullen: Right, I thought about this too and decided to at least use the averaged center for the plotting to give the user a better visual gauge of whether the center is where they think it should be. I was thinking of including the center coords of all the rings as some return value. I could do that, but for the plotting do you think it would be helpful to plot all the centers or leave it as the averaged one?

madelinegullen: I just made the changes you suggested and added the centers to be part of what's returned!

ambarb: nice idea to return average center and the center for each ring

ambarb:
Is it possible to mute the plotting?  It is great for debugging or a fast visual confirmation, but once someone knows it works, I am not sure if everyone would want to see the plot.

ambarb:
@wenhu @cmazzoli Would this function be useful to find the center of the holography image or similar imaging data (transmission CDI)?  If so, would you find the name informative enough to try the function.

CJ-Wright:
Do you need complete rings for this?

danielballan:
> Do you need complete rings for this?

The principle of the Hough transform should work just the same on a section (half, quarter, etc.) of a ring as on a whole ring.

skbeam/core/roi.py
danielballan: Seems good. It might be worth considering shortening the return value to `centers, radii` because it is easy enough for the caller to then compute the average center---and, say, the standard deviation or any other stats---if they want them.

The case for leaving it `center_pt` in, as it is now, would be as a "convenience" if we strongly expect users to be interested in the average center specifically as the primary result of interest.

skbeam/core/roi.py
```diff
@@ -757,3 +759,65 @@ def auto_find_center_rings(avg_img, sigma=1, no_rings=4, min_samples=3,
...
+                                               total_num_peaks=num_rings)
+
+    # Draw them
+    fig, ax = plt.subplots(ncols=1, nrows=1, figsize=(10, 4))
+    Z = color.gray2rgb(image)
+
+    centers =[]
+    for center_y, center_x, radius in zip(cy, cx, radii):
+        centers.append(center_x, center_y)
+        circy, circx = draw.circle_perimeter(center_y, center_x, radius,
```
tacaswell: Matplotlib can directly draw circles https://matplotlib.org/api/_as_gen/matplotlib.patches.Circle.html#matplotlib.patches.Circle

```python
import matplotlib.patches as mpatches
circ = mpatches.Circle((center_x, center_y), radius, facecolor='none', edgecolor='r')
ax.add_artist(circ)
```

skbeam/core/roi.py
```diff
@@ -757,3 +759,65 @@ def auto_find_center_rings(avg_img, sigma=1, no_rings=4, min_samples=3,
         edge_pts_xy = edge_pts_xy[~inliers]
 
     return center, image, radii
+
+def find_center_rings_hough(image, num_rings, rad_range, intensity_thresh, sigma, v_range = None):
```
tacaswell: This should probably go into the 'feature' module (https://github.com/scikit-beam/scikit-beam/blob/master/skbeam/core/feature.py) as this is about finding and extracting a feature from an image rather than generating a label array for use.  That suggest the function above this should be moved as well and should probably adopt the same return signature as we are proposing here.

madelinegullen: thanks for the feedback! I made the changes you requested and moved it to feature with a separate plotting function

madelinegullen:
@CJ-Wright 
Heres a demonstration notebook that includes an example with an incomplete ring https://github.com/madelinegullen/find_center_rings/blob/master/example_find_center_rings.ipynb 
If you had a specific image you wanted to test it on I could give it a go!

skbeam/core/roi.py
```diff
@@ -757,3 +759,65 @@ def auto_find_center_rings(avg_img, sigma=1, no_rings=4, min_samples=3,
...
+    """
+    Take angled cut or multiple incremented cuts of a image data.
```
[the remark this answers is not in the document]
madelinegullen: Do you mean the content or the extra "a"? I just got rid of the a

ambarb:
@madelinegullen It seems there was another change request.  Is this something you can handle?

skbeam/core/roi.py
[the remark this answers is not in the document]
ambarb: @tacaswell is this resolved.  @madelinegullen thought she changed it so that the "a" was removed so it should now read `....cuts of image data`. I think she made all changes in the last item of this string of requests.  Is it on @madelinegullen to click resolve (or myself now) or you?  Unsure of best git practice.

PR #554, opened 2019-08-30
Specify numpy version during the installation; support OSX builds

mrakitin:
#552 removed the testing of multiple numpy versions. Fixing it here for pip installation.

Also, added OSX to the test matrix.


mrakitin:
Thanks!

PR #561, opened 2020-02-19
Compatibility with xraylib 4.0

dmgav:
Starting from xraylib v4.0, the functions are raising exception `ValueError` when called with non-existent emission line/element combination. Previous versions printed optional error message and returned 0. In this PR, the respective functions are changed to catch `ValueError` and return 0.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ parts/
 sdist/
 var/
 venv/
+pip-wheel-metadata/
 *.egg-info/
 .installed.cfg
 *.egg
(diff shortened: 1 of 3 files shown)


mrakitin:
I think before it gets merged, it's worth to have a test with 2 versions of xraylib. We can build v4.0.0 for that test in nsls2forge.

mrakitin:
Great progress, thanks for addressing my suggestion, @dmgav!

I noticed that TravisCI complains about a few outdated config items in the `.travis.yml`:
image

Do you think we can also fix it here?

Issue #431, opened 2016-03-28
Handling NaNs


vivekthampy:
The multi_tau_auto_corr functions returns nothing if there are any NaN values in the array defined by the labelled array. It will be great if we don't have to worry about selecting regions which do not have any NaN values. @sameera2004 @tacaswell @stuwilkins

tacaswell:
This is a feature to allow marking bad frames with nan.

If you have fixed bad pixels just make sure they are removed from your roi
mask.

ericdill:
For an example of the use case that @tacaswell is talking about: https://github.com/scikit-beam/scikit-beam-examples/blob/master/demos/time_correlation/Removing_bad_images_one_time_correlation.ipynb

stuwilkins:
@ericdill that wasn't Viveks comment, he has pixels in the ROI which are always bad and marked as NaNs.

@vivekthampy you could fix this by multiplying your ROI by the flat field checking for NaN. For example

```python
np.ones((960,960), dtype=np.int64) * (np.isnan(np.rot90(new_flat, -1)) == False)
```

Would produce a 960 by 960 mask of 1 and 0 where the pixels are bad. Multiplying this by your ROI should solve the problem. Let me know if it works. Perhaps we should make a useful function for that?

stuwilkins:
P.S. Where ```new_flat``` is the flat field correction.

ericdill:
> @ericdill that wasn't Viveks comment, he has pixels in the ROI which are always bad and marked as NaNs.

@stuwilkins I was merely trying to provide a little more context to the response of @tacaswell

sameera2004:
Yes multi tau auto correlation code assume that you don't have bad pixels, you mask out bad pixels. @tacaswell and @ericdill are talking about when you have bad frames you can convert them to NAN array and it will be taken care as in the example @ericdill was mentioning.

 If you think it is useful after creating the rois you can check the rois.
We can add that check inside the correlation code if you think it is useful??
 Right now CHX correlation pipeline check
1) the rois after creating it
2 ) check  if any of rois the number of pixels are zero before using multi tau auto correlation.

mpmdean:
Hey @vivekthampy,

Just to add to the explanation (in a hopefully helpful way) NaNs are used in the algorithm to tag and not use bad _frames _ (not bad CCD regions) in a proper way.

Start by either indexing the data array to throw away bad regions or choosing the roi to avoid NaNs.

[...]

PR #564, opened 2020-04-07
Fix too much nan skipping

tacaswell:
There is some extra commits in here to be more forgiving when optional dependencies are not installed, I can drop those commits if needed.

Only the last commit is actually relevant to the bug at had.


dmgav:
What is the idea of the changes? I remember that CI was passing without issues.

tacaswell:
- The changes to the NaN handling are to address #431
- xraylib and pyfai are both "optional" so I made the tests also optional (I was having trouble getting them to install locally)
 - moving from globals -> fixtures is to make sure the tests all run independently

skbeam/core/correlation.py
```diff
@@ -128,7 +130,7 @@ def _one_time_process(buf, G, past_intensity_norm, future_intensity_norm,
 
         # take out the past_ing and future_img created using bad images
```
dmgav, line 131:
```suggestion
        # take out the past_img and future_img created using bad images
```

skbeam/core/tests/test_correlation.py
```diff
@@ -96,8 +98,8 @@ def test_lazy_vs_original():
     assert np.all(two_time[1] == final_gen_result_two.lag_steps)
 
 
-def test_lazy_two_time():
-    setup()
+def test_lazy_two_time(corr_setup):
+    num_levels, num_bufs, xdim, ydim, stack_size, img_stack, rois = corr_setup
```
mrakitin, line 102: Why don't we call `corr_setup()` here?

dmgav: I guess the tuple of the input values is computed by 'corr_setup' fixture and passed as a parameter. It may happen that the original test was written before fixtures were available.

skbeam/core/tests/test_correlation.py
```diff
@@ -202,6 +204,19 @@ def test_bad_images():
...
 
 
+def test_one_bad_pixel(corr_setup):
+    num_levels, num_bufs, xdim, ydim, stack_size, img_stack, rois = corr_setup
+    img_stack = img_stack.astype(float)
+    g2_ref, lag_steps_ref = multi_tau_auto_corr(4, num_bufs, rois, img_stack)
+    img_stack[:, 250, 100] = np.nan
+    g2_test, lag_steps_test = multi_tau_auto_corr(4, num_bufs, rois, img_stack)
+
+    assert (g2_ref == g2_test).all()
```
dmgav, line 214: It would be safer to compare the arrays using `numpy.testing.assert_arrays_almost_equal`, but in this case the result is computed by the same function, so it is supposed to be identical. So I guess it may be left as is.

skbeam/core/tests/test_correlation.py
```diff
@@ -202,6 +204,19 @@ def test_bad_images():
...
+    assert (lag_steps_ref == lag_steps_test).all()
```
dmgav, line 215: See comment above

Issue #565, opened 2020-04-24
1-time from 2-time does not average appropriately


ambarb:
https://github.com/scikit-beam/scikit-beam/blob/4fc0cfd4c20292e481a30e233cc70d27a3e5b5e0/skbeam/core/correlation.py#L829-L851

The 1-time results is the average of the 2-time array along the diagonal.  This function calculates the sum and divides by the total number of frames.  But the total number of "equivalent" pixels are not equal to the total number of frames (say the t1=0, t2= final_frame is one pixel).

Should be a simple fix to https://github.com/scikit-beam/scikit-beam/blob/4fc0cfd4c20292e481a30e233cc70d27a3e5b5e0/skbeam/core/correlation.py#L850

None elegantly:
`one_time_corr[:, j] = np.trace(g, offset=j)/ (two_time_corr.shape[2] - j )`
should fix it.

Confirmed plots in notebook, but I haven't cloned this repo to make the changes. But pictures of the "current sk-beam code" versus the change above (aka "fair averaging") are below.

![Toy_example](https://user-images.githubusercontent.com/13630233/80168745-a6101980-85b1-11ea-8743-3d23db095a34.png)
![Real_data](https://user-images.githubusercontent.com/13630233/80168750-a90b0a00-85b1-11ea-903e-90a0cc95d14a.png)

ambarb:
If the proposed change looks ok,   I will have time next week.  It is a small change and I could use the practice.

dmgav:
It looks like the change is fixing the problem.

ambarb:
ok.  I found something else and will do a PR on both sometime next week.

ambarb:
does anyone have suggestions on improving test for this function? 
https://github.com/scikit-beam/scikit-beam/blob/4fc0cfd4c20292e481a30e233cc70d27a3e5b5e0/skbeam/core/tests/test_correlation.py#L205-L224 

I have to change it as the test failed with definition changes.  It also doesn't test if two arrays are handled correctly so I had planned on adding this too.

ambarb:
https://github.com/scikit-beam/scikit-beam/pull/566

PR #566, opened 2020-04-29
fair weights to average in one-time from two-time

ambarb:
* changed `one_time_from_two_time()` to fairly weight pixel mean in diagonal
  * used `np.nanmean()` in case of dropped frames
  * handles multiple two-times 
* updated test to include checking on multiple two-time arrays
  *  cover both major changes to function
  * added one_time calculation to check contrast ratio is unaffected by future change. assumes contrast calculations are already correct for all functions.

In the future, should I branch skbeam for my updates and then fork ?

The tests passed locally so let me know if I need to do anything else before someone is willing to review.

resulting_2time

compare_one_times

### DIFF
--- a/skbeam/core/cdi.py
+++ b/skbeam/core/cdi.py
@@ -43,6 +43,7 @@ import time
 from scipy.ndimage.filters import gaussian_filter
 
 import logging
+
 logger = logging.getLogger(__name__)
 
 
@@ -90,13 +91,11 @@ def gauss(dims, sigma):
         ND gaussian
     """
     x = _dist(dims)
-    y = np.exp(-(x / sigma)**2 / 2)
+    y = np.exp(-((x / sigma) ** 2) / 2)
     return y / np.sum(y)
 
 
-def pi_modulus(recon_pattern,
-               diffracted_pattern,
-               offset_v=1e-12):
+def pi_modulus(recon_pattern, diffracted_pattern, offset_v=1e-12):
     """
     Transfer sample from real space to q space.
     Use constraint based on diffraction pattern from experiments.
@@ -117,13 +116,15 @@ def pi_modulus(recon_pattern,
     """
     diff_tmp = np.fft.fftn(recon_pattern) / np.sqrt(np.size(recon_pattern))
     index = diffracted_pattern > 0
-    diff_tmp[index] = (diffracted_pattern[index] *
-                       diff_tmp[index] / (np.abs(diff_tmp[index]) + offset_v))
+    diff_tmp[index] = (
+        diffracted_pattern[index]
+        * diff_tmp[index]
+        / (np.abs(diff_tmp[index]) + offset_v)
+    )
     return np.fft.ifftn(diff_tmp) * np.sqrt(np.size(diffracted_pattern))
 
 
-def find_support(sample_obj,
-                 sw_sigma, sw_threshold):
+def find_support(sample_obj, sw_sigma, sw_threshold):
     """
     Update sample area based on thresholds.
 
@@ -144,7 +145,7 @@ def find_support(sample_obj,
     sample_obj = np.abs(sample_obj)
     conv_fun = gaussian_filter(sample_obj, sw_sigma)
     conv_max = np.max(conv_fun)
-    return conv_fun >= (sw_threshold*conv_max)
+    return conv_fun >= (sw_threshold * conv_max)
 
 
 def cal_diff_error(sample_obj, diffracted_pattern):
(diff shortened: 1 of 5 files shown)


ambarb:
Looks like most of the travis changes to test_correlation.py were other parts of the code that haven't been tested by Travis CI in a while.   

Note PR motivated by https://github.com/scikit-beam/scikit-beam/issues/565

skbeam/core/tests/test_correlation.py
```diff
@@ -181,69 +190,95 @@ def test_auto_corr_scat_factor():
...
 
 def test_one_time_from_two_time():
+    np.random.seed(333)
     num_lev = 1
     num_buf = 10  # must be even
     x_dim = 10
     y_dim = 10
     stack = 10
     imgs = np.random.randint(1, 3, (stack, x_dim, y_dim))
+    for i in range(imgs.shape[0]):
```
dmgav, line 221: I think this could be accomplished as `imgs[:, 0, 0] += 5`

ambarb: ok.  will work on that.

ambarb: @dmgav your suggestion is better. thanks for the kind review.
 Just a general git etiquette question:  once the tests pass, do I click resolve conversation or do one of you?

ambarb:
@tacaswell @dmgav we noticed almost 2 weeks ago flake8 is configured in two files `setup.cfg` and `.flake8`.  Should we combine?  I can do it here if you say which way you want this done.

skbeam/core/cdi.py
```diff
@@ -307,21 +322,24 @@ def cdi_recon(diffracted_pattern, sample_obj, sup,
...
-    if pi_modulus_flag == 'real':
+    if pi_modulus_flag == "real":
         real_operation = True
-    elif pi_modulus_flag == 'complex':
+    elif pi_modulus_flag == "complex":
         real_operation = False
     else:
-        raise ValueError('py_modulus_flag must be one of {"complex",'
-                         'real"} not' '{!r}'.format(pi_modulus_flag))
+        raise ValueError(
```
dmgav, line 335: `        raise ValueError('py_modulus_flag must be one of ("complex",'
                         '"real") not' '{!r}'.format(pi_modulus_flag))
`
I think that this is the issue with flake8. The latest version of flake8 seems to make more comprehensive code analysis, so it finds issues in all projects. This is an example when flake8 found a real bug: the code wouldn't work as expected if this branch was hit.

dmgav: (The difference is that {} is replaced by ())

dmgav:
```        
            raise ValueError('py_modulus_flag must be one of ("complex", "real")' 
                 'not {!r}'.format(pi_modulus_flag))
```
Better formatted version

ambarb: Ahh.  That makes sense, I was just focused on why it thought things were not in the required format.  I was going to go back in time and start fresh so your message is timely.  

I don't understand why the files I haven't touched were getting flagged.  Do the travis checks check everything?

dmgav:
> @tacaswell @dmgav we noticed almost 2 weeks ago flake8 is configured in two files `setup.cfg` and `.flake8`. Should we combine? I can do it here if you say which way you want this done.

I think that the best way is to put configuration for flake8 only in `.flake8`, since this configuration file is used when you run flake8 both locally and on CI. I am not sure if configuration from `setup.cfg` is used at all, but some settings there seem to be out of date. For example, I don't think that the file `xrf_model.py` should be skipped. Currently the settings in `.flake8` are sufficient for the test to pass, so I think that flake8 entry can be safely removed from `setup.cfg`.

skbeam/core/cdi.py
dmgav: flake8 check the whole project every time you run it, unless you specify the files you want to check. You may have older version of flake8 installed locally, so you may need to update it to see the same result as on Travis, which is always using the latest version.

dmgav:
I think that 'flake8' could be simply removed from `setup.cfg` without adding anything to `.flake8`, because the test is passed successfully on the whole project with configuration already contained in `.flake8`. We want to keep the number of excluded files as low as possible.

ambarb:
ok.  removed entire line and all tests passed. will push that version of .flake8

ambarb:
> ok. removed entire line and all tests passed. will push that version of .flake8

ambarb:
I am not sure how I managed to close.

PR #567, opened 2020-05-01
FIX: disable deprecation message when using xraylib v4.0.0 or higher

dmgav:
This PR contains small change that disables very annoying deprecation message displayed when the function `SetErrorMessages(0)` is called to disable error messages with `xraylib` v4.0.0 or higher. Starting with v4.0.0, xraylib is raising exceptions instead of printing error messages and the function `SetErrorMessages` does absolutely nothing and shouldn't be called. On the other hand, if an older version of xraylib is installed, the function should be used to disable unnecessary error messages.
The changes include checking for the version of xraylib and bypass calling of `SetErrorMessages` if the version is higher than v4.0.0.

### DIFF
--- a/skbeam/core/constants/xrf.py
+++ b/skbeam/core/constants/xrf.py
@@ -40,6 +40,7 @@ from __future__ import absolute_import, division, print_function
 from collections.abc import Mapping
 import logging
 
+from distutils.version import LooseVersion
 import numpy as np
 import six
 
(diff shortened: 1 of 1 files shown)


dmgav:
I propose to merge the PRs https://github.com/scikit-beam/scikit-beam/pull/566 and https://github.com/scikit-beam/scikit-beam/pull/564 before this one. Then we can probably release the new version

PR #571, opened 2020-08-08
Adding error bars to one_time_from_two_times 

tankonst:
Added errors (standard deviation of the mean) for calculating one-time correlation function from two-time correlation function. Made this as optional output, so all the existing code would still work.

### DIFF
--- a/skbeam/core/correlation.py
+++ b/skbeam/core/correlation.py
@@ -934,7 +934,7 @@ def one_time_from_two_time(two_time_corr, calc_errors = False):
     one_time_corr : array
         matrix of one time correlation
         shape (number of labels(ROI's), number of frames)
-    err_one_time_corr: arry
+    err_one_time_corr: array
         matrix of errors for one time correlation
         shape (number of labels(ROI's), number of frames)
     """
--- a/skbeam/core/correlation.py
+++ b/skbeam/core/correlation.py
@@ -915,7 +915,7 @@ def _validate_and_transform_inputs(num_bufs, num_levels, labels):
     )
 
 
-def one_time_from_two_time(two_time_corr, calc_errors = False):
+def one_time_from_two_time(two_time_corr, calc_errors=False):
     """
     This will provide the one-time correlation data from two-time
     correlation data. An estimator for errors can be calculated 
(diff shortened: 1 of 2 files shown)


skbeam/core/correlation.py
```diff
@@ -915,29 +915,43 @@ def _validate_and_transform_inputs(num_bufs, num_levels, labels):
...
+        True to calculate error bars, False to not calculate error bars
 
     Returns
     -------
     one_time_corr : array
         matrix of one time correlation
         shape (number of labels(ROI's), number of frames)
+    err_one_time_corr: array
+        matrix of errors for one time correlation
+        shape (number of labels(ROI's), number of frames)
```
mrakitin, line 939: This part is conditional, it may or may not be returned. Should it be mentioned?

skbeam/core/correlation.py
```diff
@@ -915,29 +915,43 @@ def _validate_and_transform_inputs(num_bufs, num_levels, labels):
...
 
     one_time_corr = np.zeros((two_time_corr.shape[0], two_time_corr.shape[2]))
+    err_one_time_corr = np.zeros((two_time_corr.shape[0], two_time_corr.shape[2]))
     for i, g in enumerate(two_time_corr):
         for j in range(g.shape[1]):
             one_time_corr[i, j] = np.nanmean(np.diag(g, k=j))
-    return one_time_corr
+            if calc_errors:
+                err_one_time_corr[i, j] = np.nanstd(np.diag(g, k=j)) / np.sqrt(
+                    len(np.diag(g, k=j))
```
mrakitin, line 949: `np.diag(g, k=j)` is used twice in this equation. Should it be stashed to a variable and used once? Not sure about the performance overhead, but it looks like it can slightly improve it.

skbeam/core/correlation.py
```diff
@@ -915,29 +915,43 @@ def _validate_and_transform_inputs(num_bufs, num_levels, labels):
...
+                )
+    if calc_errors:
+        return one_time_corr, err_one_time_corr
+    else:
+        return one_time_corr
```
mrakitin, line 954: I guess we have to return it for backwards-compatibility reasons. Right?

tankonst: yes, that is right

skbeam/core/correlation.py
tankonst: added this

tankonst: Also added the error bar test for the previous version test.

skbeam/core/correlation.py
```diff
@@ -915,29 +915,44 @@ def _validate_and_transform_inputs(num_bufs, num_levels, labels):
...
     one_time_corr : array
         matrix of one time correlation
         shape (number of labels(ROI's), number of frames)
+    (optional, if calc_errors=True) err_one_time_corr: array
+        matrix of errors for one time correlation
+        shape (number of labels(ROI's), number of frames)
     """
 
     one_time_corr = np.zeros((two_time_corr.shape[0], two_time_corr.shape[2]))
+    err_one_time_corr = np.zeros((two_time_corr.shape[0], two_time_corr.shape[2]))
```
mrakitin, line 943: Maybe we should not allocate this array if `calc_errors=False`.

tankonst: Thanks, @mrakitin ! I have changed to conditional memory allocation.

PR #585, opened 2024-11-10
Mnt/np and cython updates

tacaswell:
### Fixed

Building with newer versions of numpy+cython

### Removed

support for 32 bit integers in some histogram contexts, however it may have always been silently broken.


jennmald (approved):
Looks good!

tacaswell:
I self-merged on @jennmald 's review.

PR #586, opened 2025-08-27
Dev/change default branch

tacaswell:
## Description

Deal with fallout from changing the name of the default branch


tacaswell:
self-merging as this is only CI config and docs.

